PR #2, opened 2016-09-21
Services.KeyVault key rotation support classes

scottbommarito:
Added two classes to `NuGet.Services.KeyVault` that are necessary across multiple repos for key rotation. Also moved `CachingSecretReader`, which was in `NuGetGallery`, to this project.

The new classes are...

- `KeyVaultConfigurationService` which implements `IKeyVaultConfigurationService`
- `KeyVaultConfigurationDictionary` which implements `IDictionary<string, string>`

Both allow access to a dictionary of configuration values that are refreshed with an `ISecretInjector` on every access. `KeyVaultConfigurationService` is intended for asynchronous use and `KeyVaultConfigurationDictionary` is intended for synchronous use. They are best used with a `SecretInjector` that uses a `CachingSecretReader`, so that excess `KeyVault` calls are minimized.

I may want to add some more testing to `KeyVaultConfigurationService`, given that only the core path is currently tested and `Set`, `Remove`, and other auxiliary methods are untested.


src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
scottbommarito: mostly copied from NuGetGallery

https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Configuration/SecretReader/CachingSecretReader.cs

tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
scottbommarito: mostly copied from NuGetGallery

https://github.com/NuGet/NuGetGallery/blob/master/tests/NuGetGallery.Facts/App_Start/CachingSecretReaderFacts.cs

src/NuGet.Services.KeyVault/IKeyVaultConfigurationService.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// Gets an argument from the dictionary synchronously.
+        /// </summary>
+        /// <typeparam name="T">Converts the argument from a string into this type.</typeparam>
+        /// <param name="key">The key mapping to the desired argument.</param>
+        /// <returns>The argument mapped to by the key converted to type T.</returns>
+        /// <exception cref="KeyNotFoundException">Thrown when the key is not mapped to an argument.</exception>
+        /// <exception cref="ArgumentNullException">Thrown when the argument mapped to by the key is null or empty.</exception>
+        /// <exception cref="NotSupportedException">Thrown when the argument mapped to by the key cannot be converted into an object of type T.</exception>
+        [Obsolete("Use GetOrThrow unless a synchronous context is completely necessary.")]
+        T GetOrThrowSync<T>(string key);
```
xavierdecoster: Typically, if there's a sync and async version of a method, the code convention is to add the `Async` suffix to the async method. In this case, [this](https://github.com/NuGet/ServerCommon/pull/2/files#diff-014f37185050392133c2569f0f672287R24) would become `GetOrThrowAsync` and `GetOrThrowSync` would just become `GetOrThrow`. 

Perhaps rename for consistency?

src/NuGet.Services.KeyVault/IKeyVaultConfigurationService.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Gets an argument from the dictionary synchronously.
+        /// </summary>
+        /// <typeparam name="T">Converts the argument from a string into this type.</typeparam>
+        /// <param name="key">The key mapping to the desired argument.</param>
+        /// <param name="defaultValue">The value returned if there is an issue getting the argument from the cache.</param>
+        /// <returns>The argument mapped to by the key converted to type T or defaultValue if the argument could not be acquired and converted.</returns>
+        [Obsolete("Use GetOrDefault unless a synchronous context is completely necessary.")]
+        T GetOrDefaultSync<T>(string key, T defaultValue = default(T));
```
xavierdecoster: Same here. Use `Async` suffix on [this](https://github.com/NuGet/ServerCommon/pull/2/files#diff-014f37185050392133c2569f0f672287R33) method and rename `GetOrDefaultSync` to `GetOrDefault`?

src/NuGet.Services.Configuration/KeyVaultConfigurationReader.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace NuGet.Services.Configuration
 {
-    public class SecretConfigurationReader : IConfigurationRoot
+    public class KeyVaultConfigurationReader : IConfigurationRoot
```
skofman1: >KeyVault [](start = 17, length = 8)

this class was  named SecretConfigurationReader and not KeyVaultConfigurationReader because it (intentionally) has no knowledge that KeyVault is used. 
KeyVault is an implementation detail. Tomorrow we can read our secrets from a different source.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public class CachingSecretReader : ISecretReader
+    {
+        public const int DefaultRefreshIntervalSec = 60 * 60 * 24; // 1 day
+        private readonly int _refreshIntervalSec;
+
+        private ISecretReader _internalReader;
+        private Dictionary<string, Tuple<string, DateTime>> _cache;
+
+        public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec)
```
skofman1: >refreshIntervalSec [](start = 67, length = 18)

verify this value is legal (non negative)

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            _internalReader = secretReader;
+            _cache = new Dictionary<string, Tuple<string, DateTime>>();
+
+            _refreshIntervalSec = refreshIntervalSec;
+        }
+
+        public async Task<string> GetSecretAsync(string secretName)
+        {
+            if (!_cache.ContainsKey(secretName) || DateTime.UtcNow.Subtract(_cache[secretName].Item2).TotalSeconds >= _refreshIntervalSec)
```
skofman1: >DateTime.UtcNow.Subtract(_cache[secretName].Item2).TotalSeconds >= _refreshIntervalSec [](start = 51, length = 86)

Move to a separate method with a friendly name like IsCachedSecretExpired

tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
```diff
@@ -0,0 +1,67 @@
...
+            Thread.Sleep(delayBeforeRefreshingMs);
+            mockSecretReader.Setup(x => x.GetSecretAsync(It.IsAny<string>())).Returns(Task.FromResult(secondSecret));
+
+            // Act 2
+            string value2 = await cachingSecretReader.GetSecretAsync("secretname");
+            value2 = await cachingSecretReader.GetSecretAsync("secretname");
+
+            // Assert 2
+            mockSecretReader.Verify(x => x.GetSecretAsync(It.IsAny<string>()), Times.Exactly(2));
+            Assert.Equal(secondSecret, value2);
```
skofman1: one test per method

tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
```diff
@@ -0,0 +1,67 @@
...
+            // Act
+            string value1 = await cachingSecretReader.GetSecretAsync("secretname");
+            value1 = await cachingSecretReader.GetSecretAsync("secretname");
+
+            // Assert
+            mockSecretReader.Verify(x => x.GetSecretAsync(It.IsAny<string>()), Times.Once);
+            Assert.Equal(firstSecret, value1);
+
+            // Arrange 2
+            Thread.Sleep(delayBeforeRefreshingMs);
```
skofman1: >Thread.Sleep(delayBeforeRefreshingMs); [](start = 12, length = 38)

a UT should be very fast, so please dont add a Thread.Sleep inside. There are a bunch of ways to mock DateTime in tests. Read :)

src/NuGet.Services.Configuration/KeyVaultConfigurationReader.cs
scottbommarito: Hmm I think I got confused and didn't mean to rename this, woops.

src/NuGet.Services.KeyVault/KeyVaultConfigurationDictionary.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _secretInjector = secretInjector;
+
+            foreach (var key in unprocessedArguments.Keys)
+            {
+                base[key] = unprocessedArguments[key];
+            }
+        }
+
+        public new string this[string key]
```
skofman1: >this [](start = 26, length = 4)

won't work

src/NuGet.Services.KeyVault/IKeyVaultConfigurationService.cs
scottbommarito: Hmm...I understand the convention, but `GetOrThrowSync` is meant to almost never be called and is a method for a very small portion of the time when, for whatever reason (usually due to an interface that we don't control), we need to check configuration values in a synchronous context. Given this, I thought it was ok to leave it as `GetOrThrow` instead of `GetOrThrowAsync` because it's the intended method like 95% of the time, but it might be better to change it anyway. Thoughts?

xavierdecoster: So are you saying that, besides these two methods, all of the public API in this library is async? In that case, ignore my comment :)

scottbommarito:
Closing for now because `SecretService` will be moved and added to another PR when the time comes.

PR #3, opened 2016-09-28
SecretDictionary and CachedSecretReader (job secret rotation)

scottbommarito:
Added `SecretDictionary` for secret rotation of jobs (and anything else that's not MVC).

Implements `IDictionary<string, string>` and serves mostly to add an additional layer between the user and another `IDictionary<string, string>` by injecting secrets upon every access.

Also added `CachedSecretReader` because it will be shared across many things.


scottbommarito:
Closed because this was merged as part of another PR.

PR #4, opened 2016-10-17
SecretDictionary

scottbommarito:
I added the `CachingSecretReader` and `SecretDictionary` classes. These will be used across all our services to regularly refresh our secrets.

For ease and resusability for the sake of the Secret Rotation job, I also moved the `FindCertificateByThumbprint` code into `KeyVaultConfiguration` and exposed both the configuration and the `KeyVaultClient` publicly.


tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
```diff
@@ -0,0 +1,67 @@
...
+            // Act
+            string value1 = await cachingSecretReader.GetSecretAsync("secretname");
+            value1 = await cachingSecretReader.GetSecretAsync("secretname");
+
+            // Assert
+            mockSecretReader.Verify(x => x.GetSecretAsync(It.IsAny<string>()), Times.Once);
+            Assert.Equal(firstSecret, value1);
+
+            // Arrange 2
+            Thread.Sleep(delayBeforeRefreshingMs);
```
skofman1: > Thread.Sleep(delayBeforeRefreshingMs) [](start = 11, length = 38)

tests should never use sleep, since they should be fast. The way to mock DateTime is to use IDateTimeProvider. In this particular case, since this is framework code, I suggest running this test once, and then adding a [Skip] attribute.

src/NuGet.Services.Configuration/ISecretReaderFactory.cs
```diff
@@ -7,6 +7,8 @@ namespace NuGet.Services.Configuration
 {
     public interface ISecretReaderFactory
```
skofman1: >ISecretReaderFactory [](start = 21, length = 20)

can we move the interface to NuGet.Services.KeyVault

src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -39,5 +40,30 @@ public KeyVaultConfiguration(string vaultName, string clientId, string certifica
             StoreName = storeName;
             StoreLocation = storeLocation;
         }
+        
+        public X509Certificate2 GetCertificate()
```
skofman1: >GetCertificate [](start = 32, length = 14)

please move back

src/NuGet.Services.Configuration/ISecretReaderFactory.cs
```diff
@@ -7,6 +7,8 @@ namespace NuGet.Services.Configuration
...
     {
+        KeyVaultConfiguration CreateKeyVaultConfiguration();
```
skofman1: This is SecretReader, it doesn't know anything about KeyVault

src/NuGet.Services.KeyVault/SecretDictionary.cs
```diff
@@ -0,0 +1,117 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGet.Services.KeyVault
+{
+    public class SecretDictionary : IDictionary<string, string>
```
skofman1: Lets move this one to NuGet.Services.Configuration

skofman1:
:shipit:

PR #5, opened 2016-10-19
Fix NuGet.Services.Configuration project.json

scottbommarito:
A relatively minor fix to the project.json files to avoid issues with type-forwarding assemblies in the build.


skofman1:
🚢 
🐑

PR #6, opened 2016-10-19
Added extensions to IDictionary<string, string> for ease of use in jobs

scottbommarito:
Currently, our jobs in `NuGet.Jobs` and our jobs in `NuGet.Services.Metadata` both use their own "TryGet" infrastructure. This manifests as [`JobConfigurationManager`](https://github.com/NuGet/NuGet.Jobs/blob/master/src/NuGet.Jobs.Common/Configuration/JobConfigurationManager.cs) in `NuGet.Jobs` and [`CommandHelpers`](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/CommandHelpers.cs) in `NuGet.Services.Metadata`.

I have added some extensions to `IDictionary<string, string>` to `NuGet.Services.Configuration` to consolidate the logic in a single place.

(`ConvertFromString` will also be using in the `SecretService` class for secret rotation and so it is in its own utility class.)

### DIFF
--- a/src/NuGet.Services.Configuration/ConfigurationUtility.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationUtility.cs
@@ -8,15 +8,21 @@ namespace NuGet.Services.Configuration
 {
     public static class ConfigurationUtility
     {
+        /// <summary>
+        /// Converts a string into T.
+        /// </summary>
+        /// <typeparam name="T">Type that value will be converted into.</typeparam>
+        /// <param name="value">String to convert.</param>
+        /// <returns>Value converted into T.</returns>
+        /// <exception cref="NotSupportedException">Thrown when a conversion from string to T is impossible.</exception>
         public static T ConvertFromString<T>(string value)
         {
             var converter = TypeDescriptor.GetConverter(typeof(T));
             if (converter != null)
             {
-                // This will throw a NotSupportedException if no conversion is possible.
                 return (T)converter.ConvertFromString(value);
             }
-            // If there is no converter, no conversion is possible, so throw a NotSupportedException.
+
             throw new NotSupportedException("No converter exists from string to " + typeof(T).Name + "!");
         }
     }
--- a/src/NuGet.Services.Configuration/ConfigurationUtility.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationUtility.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel;
+
+namespace NuGet.Services.Configuration
+{
+    public static class ConfigurationUtility
+    {
+        public static T ConvertFromString<T>(string value)
+        {
+            var converter = TypeDescriptor.GetConverter(typeof(T));
+            if (converter != null)
+            {
+                // This will throw a NotSupportedException if no conversion is possible.
+                return (T)converter.ConvertFromString(value);
+            }
+            // If there is no converter, no conversion is possible, so throw a NotSupportedException.
+            throw new NotSupportedException("No converter exists from string to " + typeof(T).Name + "!");
+        }
+    }
+}
--- a/src/NuGet.Services.Configuration/DictionaryExtensions.cs
+++ b/src/NuGet.Services.Configuration/DictionaryExtensions.cs
@@ -7,6 +7,13 @@ namespace NuGet.Services.Configuration
 {
     public static class DictionaryExtensions
     {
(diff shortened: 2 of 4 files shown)


tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(ValueData))]
+        public void GetConvertsValueAndThrows<T>(T value) where T : struct
+        {
+            // Arrange
+            const string key = "key";
+            const string notKey = "notAKey";
+            IDictionary<string, string> dictionary = new Dictionary<string, string>
```
skofman1: >IDictionary<string, string> [](start = 12, length = 27)

var

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+        [Theory]
+        [MemberData(nameof(ValueData))]
+        public void TryGetValueConvertsAndDoesNotThrow<T>(T value) where T : struct
+        {
+            // Arrange
+            const string key = "key";
+            const string notKey = "notAKey";
+            IDictionary<string, string> dictionary = new Dictionary<string, string>
```
skofman1: >IDictionary<string, string> [](start = 12, length = 27)

var

skofman1:
:shipit:

src/NuGet.Services.Configuration/DictionaryExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Configuration
+{
+    public static class DictionaryExtensions
+    {
+        public static string TryGetValue(this IDictionary<string, string> dictionary, string key)
```
scottbommarito: @skofman1 I just added this and an additional test case (after your review) because I noticed that `JobConfigurationManager` and `CommandHelpers` also have a similar method. Any thoughts?

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
scottbommarito: this is intentional because the extension is on `IDictionary<string, string>` so I want to verify that it works with `IDictionary<string, string>` and not `Dictionary<string, string>` specifically.

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
scottbommarito: see above

src/NuGet.Services.Configuration/ConfigurationUtility.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace NuGet.Services.Configuration
+{
+    public static class ConfigurationUtility
+    {
+        public static T ConvertFromString<T>(string value)
+        {
+            var converter = TypeDescriptor.GetConverter(typeof(T));
+            if (converter != null)
+            {
+                // This will throw a NotSupportedException if no conversion is possible.
```
shishirx34: You mean to say `ConvertFromString` could possibly throw NotSupportedException?

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                {key, value.ToString()}
+            };
+
+            // Act
+            var valueFromDictionary = dictionary.GetOrThrow<T>(key);
+
+            // Assert
+            Assert.Equal(value, valueFromDictionary);
+            Assert.Throws<KeyNotFoundException>(() => dictionary.GetOrThrow<T>(notKey));
```
shishirx34: how about validation of `NotSupportedException`?

src/NuGet.Services.Configuration/ConfigurationUtility.cs
scottbommarito: Yea, I will add the appropriate XML docs

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
scottbommarito: added, and also added a couple tests for `ConvertFromString` specifically

PR #7, opened 2016-10-20
ISecretService and SecretService

scottbommarito:
This adds the `ISecretService` interface and the `SecretService` class that implements it to `NuGet.Services.Configuration`. `SecretService` is intended to be used by services that cannot refresh their secrets synchronously (likely due to the possible deadlock caused by MVC apps using KeyVault) such as NuGetGallery.

### DIFF
--- a/tests/NuGet.Services.Configuration.Tests/SecretServiceFacts.cs
+++ b/tests/NuGet.Services.Configuration.Tests/SecretServiceFacts.cs
@@ -29,7 +29,7 @@ namespace NuGet.Services.Configuration.Tests
                 {secretName, secretKey}
             };
 
-            var configService = new SecretService(mockSecretInjector.Object, arguments);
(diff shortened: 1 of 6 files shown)


scottbommarito:
@skofman1 I made the requested changes.

tests/NuGet.Services.Configuration.Tests/SecretServiceFacts.cs
```diff
@@ -0,0 +1,245 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading.Tasks;
+using NuGet.Services.KeyVault;
+using Xunit;
+
+namespace NuGet.Services.Configuration.Tests
+{
+    public class SecretServiceFacts
```
skofman1: >SecretServiceFacts [](start = 17, length = 18)

rename

PR #8, opened 2016-10-27
Dictionary<string, string> Extensions 2.0

scottbommarito:
Rename `GetOrNull` to `GetOrDefault`.
Remove constraint on `GetOrThrow`.
Modified test cases appropriately.


src/NuGet.Services.Configuration/DictionaryExtensions.cs
```diff
@@ -41,7 +29,7 @@ public static string GetOrNull(this IDictionary<string, string> dictionary, stri
             }
             catch
```
skofman1: don't catch the exception. Should propagate to user.

scottbommarito:
@shishirx34 @ryuyu @skofman1 Quick PR!

PR #9, opened 2016-11-01
IConfigurationProvider and ICachingConfigurationProvider

scottbommarito:
I split `IConfigurationProvider` into two interfaces: `ICachingConfigurationProvider` and `IConfigurationProvider`.

`IConfigurationProvider` contains `GetOrThrow` and `GetOrDefault`.
`ICachingConfigurationProvider` contains `GetOrThrowSync` and `GetOrDefaultSync`.

`IConfigurationProvider` is for providing configuration for which all configuration values cannot be cached, such as environment settings. For such configuration, it is impossible to implement a fail-safe `GetOrThrowSync` and `GetOrThrowDefault` because we cannot rely on having all of the possible configuration values cached.

This is the case for the configuration used by our Search Service, which uses environment variables. By introducing `IConfigurationProvider` I can reuse some of the logic from `SecretConfigurationProvider` to implement a class that the Search Service will use for configuration.


src/NuGet.Services.Configuration/SecretConfigurationProvider.cs
```diff
@@ -11,7 +11,7 @@ namespace NuGet.Services.Configuration
     /// <summary>
     /// Asynchronously provides configuration and command line arguments and injects them with secrets using an ISecretInjector on every access to refresh them.
     /// </summary>
-    public class SecretConfigurationProvider : IConfigurationProvider
+    public class SecretConfigurationProvider : SettingsProvider, IConfigurationProvider
```
ryuyu: Extending SettingProvider instead of implementing ISettingsProvider was intended here?

scottbommarito: Yes, it uses the shared logic in the `SettingsProvider` abstract base class. I have the interface split from the abstract implementation intentionally.

dtivel:
Update the PR's title and comment; `ISettingsProvider` no longer exists as of your last commit.

PR #11, opened 2016-11-07
improved the exception message if the certificate is not found

scottbommarito:
Spent far too long today debugging a stupid issue where the validation was incorrect but I thought it was a problem with the thumbprint. I think adding it to the exception message will improve debugging in the future.


xavierdecoster:
:shipit:

shishirx34:
Looks good.

PR #12, opened 2016-11-10
ConfigurationFactory

scottbommarito:
Created `IConfigurationFactory` and its implementation, `ConfigurationFactory`, as the major source for configuration in our projects. It's purpose is to inflate a class with configuration based on a `IConfigurationProvider` that is provided to it.

Using `IConfigurationFactory`, configuration can now be constructed at a single point and thereafter passed around statically. The user of the class has control over when configuration is generated, thus allowing any flow (for example, a single request) to maintain a single configuration throughout. Additionally, all configuration can be accessed statically, so there is no longer any issues with contexts that cannot be asynchronous.

Also, removed the synchronous methods from `IConfigurationProvider` since they were only used for returning cached values. All accesses to actual secret values must be asynchronous. `SecretConfigurationProvider` was removed as well because its uses can be entirely replaced by `ConfigurationFactory`.


src/NuGet.Services.Configuration/Configuration.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace NuGet.Services.Configuration
+{
+    /// <summary>
+    /// Base class to extend when using an <see cref="IConfigurationFactory"/>.
+    /// </summary>
+    public class Configuration
```
skofman1: >Configuration [](start = 17, length = 13)

should be an abstract class

src/NuGet.Services.Configuration/IConfigurationProvider.cs
```diff
@@ -8,51 +8,29 @@
 namespace NuGet.Services.Configuration
 {
     /// <summary>
-    /// Asynchronously provides configuration or command line arguments.
+    /// Asynchronously provides configuration.
+    /// Used when all possible configuration cannot be cached, for example RoleEnvironment.
     /// </summary>
     public interface IConfigurationProvider
```
skofman1: >IConfigurationProvider [](start = 21, length = 22)

I would recommend doing a much simpler interface: one method:

bool TryGet(string key, out string value)

skofman1: >interface [](start = 11, length = 9)

I would recommend doing a much simpler interface: one method:

bool TryGet(string key, out string value)

Since IConfigurationProvider is a wrapper around a RoleEnvironment/json file/environment variable, this is all you need.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace NuGet.Services.Configuration
+{
+    /// <summary>
+    /// <see cref="IConfigurationFactory"/> implementation that uses an <see cref="IConfigurationProvider"/> to inject configuration.
+    /// </summary>
+    public class ConfigurationFactory : IConfigurationFactory
+    {
+        private readonly IConfigurationProvider _configProvider;
+
+        public ConfigurationFactory(IConfigurationProvider configProvider)
```
skofman1: >configProvider [](start = 59, length = 14)

check for null

skofman1:
> using System;

file headers everywhere

---
Refers to: src/NuGet.Services.Configuration/ConfigurationFactory.cs:1 in 9be2b0f. [](commit_id = 9be2b0f1403a719a971380ef58b351f4d773e8cc, deletion_comment = False)

src/NuGet.Services.Configuration/IConfigurationProvider.cs
scottbommarito: `IConfigurationProvider` is a wrapper around a RoleEnvironment/json file/environment variable that prepares the configuration for consumption, which includes asynchronously injecting secrets. Therefore the accessors must be async. `TryGet` does not work with async, since `out` parameters are not possible.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+            else
+            {
+                var defaultValueAttribute = property.Attributes.OfType<DefaultValueAttribute>().FirstOrDefault();
+                if (defaultValueAttribute != null)
+                {
+                    try
+                    {
+                        // Use the default value specified by the DefaultValueAttribute if it can be converted into the type of the property.
+                        var defaultValue = (TP) (defaultValueAttribute.Value.GetType() == property.PropertyType ? defaultValueAttribute.Value : property.Converter.ConvertFrom(defaultValueAttribute.Value));
```
skofman1: >ty.Pr [](start = 96, length = 5)

break line

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+            if (property.Attributes.OfType<RequiredAttribute>().Any())
+            {
+                // If the property is required, use GetOrThrowAsync to access configuration.
+                // It will throw if the configuration is not found or invalid.
+                value = await _configProvider.GetOrThrowAsync<TP>(settingName);
+            }
+            else
+            {
+                var defaultValueAttribute = property.Attributes.OfType<DefaultValueAttribute>().FirstOrDefault();
+                if (defaultValueAttribute != null)
```
scottbommarito: It grabs the default value and then passes it into the `IConfigurationProvider`. The `IConfigurationProvider` returns the default if it cannot find the value in configuration.

`value = await _configProvider.GetOrDefaultAsync(settingName, defaultValue);`

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                configKey = configNameProperty.Key;
+            }
+
+            // Add the prefix specified by the ConfigurationKeyAttribute to the configuration key if it exists.
+            var configNamePrefixProperty =
+                property.Attributes.OfType<ConfigurationKeyPrefixAttribute>().FirstOrDefault();
+            if (configNamePrefixProperty != null)
+            {
+                configKey = configNamePrefixProperty.Prefix + configKey;
```
dtivel: Shouldn't you check for null/empty here like [above](https://github.com/NuGet/ServerCommon/pull/12/files#diff-1f2ec043823e5381e70758cea0bfdbe0R55)?

src/NuGet.Services.Configuration/ConfigurationKeyAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Used to specify a configuration key in a <see cref="Configuration"/> subclass while using a <see cref="ConfigurationFactory"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
+    public class ConfigurationKeyAttribute : Attribute
+    {
+        public ConfigurationKeyAttribute(string key)
+        {
+            Key = key;
```
dtivel: Throw if `key` is null or empty.

src/NuGet.Services.Configuration/ConfigurationKeyPrefixAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace NuGet.Services.Configuration
+{
+    /// <summary>
+    /// Used to specify a prefix to add to the configuration key in a <see cref="Configuration"/> subclass while using a <see cref="ConfigurationFactory"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
+    public class ConfigurationKeyPrefixAttribute : Attribute
+    {
+        public ConfigurationKeyPrefixAttribute(string prefix)
```
dtivel: Throw if `prefix` is null or empty.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (string.IsNullOrEmpty(configNameProperty.Key))
+                {
+                    throw new ArgumentNullException(nameof(configNameProperty.Key),
+                        $"Configuration key specified by {nameof(ConfigurationKeyAttribute)} is null or empty!");
+                }
+
+                configKey = configNameProperty.Key;
+            }
+
+            // Add the prefix specified by the ConfigurationKeyAttribute to the configuration key if it exists.
```
dtivel: `ConfigurationKeyPrefixAttribute`

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
```diff
@@ -0,0 +1,184 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
scottbommarito: This code is identical to what used to be in `SecretConfigurationProviderFacts` but using a `TestConfigurationProvider` instead of the now defunct `SecretConfigurationProvider`.

src/NuGet.Services.Configuration/Configuration.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public Configuration()
+        {
+            CreatedTime = DateTime.UtcNow;
+        }
+
+        public DateTime CreatedTime { get; }
```
dtivel: Use `DateTimeOffset`.

src/NuGet.Services.Configuration/Configuration.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace NuGet.Services.Configuration
+{
+    /// <summary>
+    /// Base class to extend when using an <see cref="IConfigurationFactory"/>.
+    /// </summary>
+    public class Configuration
+    {
+        public Configuration()
+        {
+            CreatedTime = DateTime.UtcNow;
```
dtivel: Use `DateTimeOffset.UtcNow`.

src/NuGet.Services.Configuration/Configuration.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
```
dtivel: Add copyright header to all files.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        /// <summary>
+        /// Injects a property of <param name="instance">instance</param> specified by <param name="property">a <see cref="PropertyDescriptor"/></param> with configuration.
+        /// </summary>
+        /// <typeparam name="T">Type of <param name="instance">the instance</param>.</typeparam>
+        /// <typeparam name="TP">Type of <param name="property">the property</param>.</typeparam>
+        /// <param name="instance">Instance to inject configuration into a property of.</param>
+        /// <param name="property"><see cref="PropertyDescriptor"/> that describes the property to inject the configuration into.</param>
+        /// <returns>A task that completes when the property has been injected into.</returns>
+        public async Task InjectPropertyWithConfiguration<T, TP>(T instance, PropertyDescriptor property)
```
dtivel: Throw for `null` arguments.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+                        var defaultValue =
+                            (TP)
+                            (defaultValueAttribute.Value.GetType() == property.PropertyType
+                                ? defaultValueAttribute.Value
+                                : property.Converter.ConvertFrom(defaultValueAttribute.Value));
+                        value = await _configProvider.GetOrDefaultAsync(configKey, defaultValue);
+                    }
+                    catch (Exception)
+                    {
+                        throw new ArgumentException($"Default value for {configKey} specified by DefaultValueAttribute is malformed ({defaultValueAttribute.Value ?? "null"})!");
```
dtivel: `{nameof(DefaultValueAttribute)}` like [here](https://github.com/NuGet/ServerCommon/pull/12/files#diff-1f2ec043823e5381e70758cea0bfdbe0R58);

src/NuGet.Services.Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Gets a value from a given key.
+        /// </summary>
+        /// <param name="key">The key associated with the desired value.</param>
+        /// <returns>The value associated with the given key.</returns>
+        /// <exception cref="KeyNotFoundException">Thrown when the key is not found in the list of keys.</exception>
+        /// <exception cref="ArgumentNullException">Thrown when the value associated with the given key is null or empty.</exception>
+        protected abstract Task<string> Get(string key);
+
+        public async Task<T> GetOrThrowAsync<T>(string key)
+        {
```
dtivel: Throw for `null` or empty `key`.

src/NuGet.Services.Configuration/ConfigurationProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+            return ConfigurationUtility.ConvertFromString<T>(await Get(key));
+        }
+
+        public async Task<T> GetOrDefaultAsync<T>(string key, T defaultValue = default(T))
```
dtivel: Throw for `null` or empty `key`.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// <see cref="IConfigurationFactory"/> implementation that uses an <see cref="IConfigurationProvider"/> to inject configuration.
+    /// </summary>
+    public class ConfigurationFactory : IConfigurationFactory
+    {
+        private readonly IConfigurationProvider _configProvider;
+
+        public ConfigurationFactory(IConfigurationProvider configProvider)
+        {
+            _configProvider = configProvider;
```
ryuyu: Is there any plan to allow multiple configurationProviders to be managed by one factory? Or is the idea that each provider should make its own factory?

scottbommarito: My idea was that if we want to have multiple `IConfigurationProvider`s for one factory, we should create an `AggregateConfigurationProvider` that aggregates a number of `IConfigurationProvider`s and give it to the factory. I haven't found a use for that yet so I didn't write it yet, but it would be fairly easy to do.

ryuyu: So make a new provider that is an aggregation of the providers you need? i guess this makes sense. Sounds fine to me.

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+            }
+        };
+
+        /// <summary>
+        /// Returns true if the object is convertible to type.
+        /// </summary>
+        /// <param name="value">The object to test.</param>
+        /// <param name="type">Type to test.</param>
+        /// <returns>True if the <param name="value">object</param> is convertible to <param name="type">type</param>, false otherwise.</returns>
+        private static bool IsValueWrongType(object value, Type type)
```
dtivel: Check for valid types instead of invalid types, since the former is a much smaller set than the latter.  The logic will need to be inverted, but I think it'll be simpler and avoid an odd conclusion (`IsValueWrongType(...)` returned `false` so it's a "right" type.).

        private static bool CanConvertTo(object value, Type type)
        {
            if (value == null)
            {
                return !type.IsValueType || Nullable.GetUnderlyingType(type) != null;
            }

            return TypeDescriptor.GetConverter(type).CanConvertFrom(value.GetType());
        }

This will avoid exceptions from failed `ConvertFrom(...)` calls.  If `type` were null, `TypeDescriptor.GetConverter(...)` would throw an `ArgumentNullException`; however, you don't have any such test cases.  Explicitly checking for a `null` `type` would be trivial though.

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+            {
+                var configTuple = configPair.Value;
+                
+                // True if the expected value is null or if it is an empty string.
+                var expectedValueIsMissing = 
+                    configTuple.ExpectedValue == null ||
+                    (configTuple.ExpectedValue is string && string.IsNullOrEmpty((string) configTuple.ExpectedValue));
+                
+                var expectedIsInvalid = IsValueWrongType(configTuple.ExpectedValue, configTuple.Type);
+                var defaultIsInvalid = IsValueWrongType(configTuple.DefaultValue, configTuple.Type);
```
dtivel: var isExpectedValueValid = IsValueTypeValid(configTuple.ExpectedValue, configTuple.Type);
    var isDefaultValueValid = IsValueTypeValid(configTuple.DefaultValue, configTuple.Type);

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+        /// Dynamically adds a property to a <see cref="TypeBuilder"/>.
+        /// </summary>
+        /// <param name="typeBuilder">The <see cref="TypeBuilder"/> to add the property to.</param>
+        /// <param name="name">The name of the property.</param>
+        /// <param name="configTuple">Specifies how to construct the property.</param>
+        private static void AddProperty(TypeBuilder typeBuilder, string name, ConfigurationTuple configTuple)
+        {
+            // Create the property attribute.
+            var propertyAttributes = configTuple.DefaultValue != null ? PropertyAttributes.HasDefault : PropertyAttributes.None;
+            var propertyBuilder = typeBuilder.DefineProperty(name, propertyAttributes, configTuple.Type, null);
```
dtivel: `parameterTypes: null`

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+                        typeof(ConfigurationKeyPrefixAttribute).GetConstructor(new[] { typeof(string) }),
+                        new object[] { configTuple.ConfigKeyPrefix }));
+            }
+
+            // Create the field that the property will get and set.
+            var fieldBuilder = typeBuilder.DefineField($"_{name}", configTuple.Type, FieldAttributes.Private);
+
+            // Create the get method for the property that will return the field.
+            var getMethod = typeBuilder.DefineMethod($"get_{name}",
+                MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.SpecialName, configTuple.Type, null);
```
dtivel: `parameterTypes: null`

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+                    configTuple.ExpectedValue == null ||
+                    (configTuple.ExpectedValue is string && string.IsNullOrEmpty((string) configTuple.ExpectedValue));
+                
+                var expectedIsInvalid = IsValueWrongType(configTuple.ExpectedValue, configTuple.Type);
+                var defaultIsInvalid = IsValueWrongType(configTuple.DefaultValue, configTuple.Type);
+
+                if ((configTuple.Required && expectedValueIsMissing) || 
+                    expectedIsInvalid || 
+                    defaultIsInvalid ||
+                    configTuple.ConfigKey == string.Empty)
```
dtivel: Add a test for `null` `ConfigKey`.

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+        /// Used to test that the <see cref="ConfigurationFactory"/> fails correctly when given incorrect types.
+        /// </summary>
+        public class NoConversionFromThisClass
+        {
+        }
+
+        /// <summary>
+        /// Test data for the <see cref="ConfigurationFactory"/> tests.
+        /// </summary>
+        public static IEnumerable<object[]> ConfigurationFactoryTestData => new[]
```
dtivel: Do you only support configuration values of primitive types?  What about nullables?  If yes, add two test cases:

1.  successful nullable conversion
2.  unsuccessful nullable conversion

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
```diff
@@ -0,0 +1,454 @@
...
+        {
+            new object[]
+            {
+                // Succeeds
+                // Tests base case
+                new Dictionary<string, ConfigurationTuple>
+                {
+                    {
+                        "stringProperty",
+                        new ConfigurationTuple(typeof(string), false, "i am a string")
```
dtivel: Named parameters would aid readability here and below.

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
```diff
@@ -0,0 +1,184 @@
...
+            Assert.Equal(secondSecret, value1);
+            Assert.Equal(value1, value2);
+        }
+
+        [Theory]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(int))]
+        [InlineData(typeof(bool))]
+        [InlineData(typeof(DateTime))]
+        public async void HandlesKeyNotFound(Type type)
```
dtivel: Prefer `Throws` instead of `Handles` in this case as it is more specific about the expected behavior.

Also `ThrowsForKeyNotFound` describes the scenario better.

You already do this [here](https://github.com/NuGet/ServerCommon/pull/12/files#diff-0beafe4a015085507da5f16982096ceeR140).

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
```diff
@@ -0,0 +1,184 @@
...
+            // GetOrDefault with default specified
+            Assert.Equal(memberOfType, await (dynamic)getOrDefaultMethod.Invoke(dummy, notFoundKeyDefaultSpecifiedArgs));
+        }
+
+        [Theory]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(int))]
+        [InlineData(typeof(bool))]
+        [InlineData(typeof(DateTime))]
+        public async void HandlesNullOrEmptyArgument(Type type)
```
dtivel: `ThrowsForNullOrEmptyArgument`

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
```diff
@@ -0,0 +1,184 @@
...
+            // GetOrDefault with default specified
+            Assert.Equal(memberOfType, await (dynamic)getOrDefaultMethod.Invoke(configProvider, emptyKeyDefaultSpecifiedArgs));
+        }
+        
+        private class NoConversionFromStringToThisClass
+        {
+        }
+
+        [Fact]
+        public async void ThrowsNotFoundException()
```
dtivel: `ThrowsForNonExistentSecretName`

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
scottbommarito: I need to add more comments to the `ConfigurationTuple` class.

`ConfigKey` refers to the value of the `ConfigurationKeyAttribute` applied to the object, not the name of the property that is used. If `ConfigKey` is null, no `ConfigurationKeyAttribute` is applied and it uses the name of the property to access configuration, so `ConfigKey` being empty is the fail case. I can add a test for null `ConfigKey` in a separate test case.

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
scottbommarito: Any type should work, I haven't done a nullable but I'll add one.

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
scottbommarito: In retrospect, this test case doesn't even check `NotFound`, it checks `NotSupported`.

`ThrowsForNotSupportedConversion`

tests/NuGet.Services.Configuration.Tests/TestConfigurationProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+            string value;
+
+            if (!_configuration.TryGetValue(key, out value))
+            {
+                throw new KeyNotFoundException();
+            }
+
+            if (string.IsNullOrEmpty(value))
+            {
+                throw new ArgumentNullException();
```
dtivel: Yes, read [this](https://msdn.microsoft.com/en-us/library/system.argumentnullexception(v=vs.110).aspx#Remarks).

src/NuGet.Services.Configuration/ConfigurationFactory.cs
scottbommarito: It's intended to use `configKey` so that one can easily find the the attribute that is malformed. In retrospect, I think all of the `ArgumentException`s need to use `configKey`.

src/NuGet.Services.Configuration/ConfigurationProvider.cs
scottbommarito: This method is not meant to throw so I will keep it as is. The call to `GetOrThrowAsync` will produce the `ArgumentNullException` indicating that the key is missing.

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
scottbommarito: It doesn't always throw. `GetOrDefaultAsync` returns a default value instead of throwing. I named it `HandlesKeyNotFound` for this purpose. Is there a better name for that?

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
scottbommarito: I remember I explicitly used `TypeDescriptor.GetConverter(type).ConvertFrom(value)` because I wanted to test when the types are convertible but the actual value is not.

E.G `"10"` can be converted to an int but `"this is a string"` cannot.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
scottbommarito: Added check below.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
scottbommarito: Added checks below.

src/NuGet.Services.Configuration/ConfigurationKeyPrefixAttribute.cs
scottbommarito: Added check below

src/NuGet.Services.Configuration/ConfigurationProvider.cs
scottbommarito: Added check below

tests/NuGet.Services.Configuration.Tests/ConfigurationFactoryFacts.cs
scottbommarito: added nullable tests below

tests/NuGet.Services.Configuration.Tests/TestConfigurationProvider.cs
scottbommarito: I added `ConfigurationNullOrEmptyException`. There didn't seem to be a relevant pre-existing exception.

scottbommarito:
@dtivel Next iteration is ready!

src/NuGet.Services.Configuration/ConfigurationKeyAttribute.cs
```diff
@@ -13,6 +13,11 @@ public class ConfigurationKeyAttribute : Attribute
     {
         public ConfigurationKeyAttribute(string key)
         {
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new ArgumentNullException(nameof(key));
```
dtivel: Strictly speaking, you should throw an `ArgumentException` here with a message "Argument cannot be null or empty."  An `ArgumentException` handles the generic case, while `ArgumentNullException` handles a very narrow case.  Throwing an `ArgumentNullException` when key is a zero-length string is misleading.

src/NuGet.Services.Configuration/ConfigurationKeyPrefixAttribute.cs
```diff
@@ -13,6 +13,11 @@ public class ConfigurationKeyPrefixAttribute : Attribute
     {
         public ConfigurationKeyPrefixAttribute(string prefix)
         {
+            if (string.IsNullOrEmpty(prefix))
+            {
+                throw new ArgumentNullException(nameof(prefix));
```
dtivel: See previous comment.

src/NuGet.Services.Configuration/ConfigurationNullOrEmptyException.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
dtivel: Add copyright header.

src/NuGet.Services.Configuration/ConfigurationNullOrEmptyException.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Configuration
+{
+    public class ConfigurationNullOrEmptyException : Exception
```
dtivel: Add XML doc comments.

tests/NuGet.Services.Configuration.Tests/ConfigurationProviderFacts.cs
```diff
@@ -80,64 +138,77 @@ public class ConfigurationProviderFacts
...
         }
-        
+
         private class NoConversionFromStringToThisClass
         {
         }
 
         [Fact]
-        public async void ThrowsNotFoundException()
+        public async void ThrowsForNotSupportedConversion()
```
dtivel: NotSupported => Unsupported

tests/NuGet.Services.Configuration.Tests/DictionaryConfigurationProvider.cs
```diff
@@ -24,14 +24,19 @@ protected override Task<string> Get(string key)
         {
             string value;
 
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new ArgumentNullException(key);
```
dtivel: See previous comment about throwing the right exception type.

tests/NuGet.Services.Configuration.Tests/ConfigurationAttributeFacts.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
dtivel: Add copyright header.

tests/NuGet.Services.Configuration.Tests/DummyConfigurationProvider.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// Returns "test value" for all configuration keys unless the given key is null or empty.
+    /// </summary>
+    public class DummyConfigurationProvider : ConfigurationProvider
+    {
+        protected override Task<string> Get(string key)
+        {
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new ArgumentNullException(key);
```
dtivel: See previous comment about exception type.

tests/NuGet.Services.Configuration.Tests/DummyConfigurationProvider.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
dtivel: Add copyright header.

src/NuGet.Services.Configuration/ConfigurationNullOrEmptyException.cs
scottbommarito: It is now thrown in `ConfigurationProvider`.

scottbommarito:
@dtivel Second iteration done!

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,135 @@
...
+                    throw new ArgumentException(
+                        $"Configuration key for {configKey} specified by {nameof(ConfigurationKeyAttribute)} is null or empty!",
+                        nameof(configNameProperty.Key));
+                }
+
+                configKey = configNameProperty.Key;
+            }
+
+            // Add the prefix specified by the ConfigurationKeyPrefixAttribute to the configuration key if it exists.
+            var configNamePrefixProperty =
```
skofman1: >configNamePrefixProperty [](start = 16, length = 24)

this can cause a problem in the gallery scenario, since all configuration items have the same prefix, and we don't want to add an identical attribute on all properties. I would suggest either having this attribute on the configuration class, or make it configurable on the factory level/input parameter.

skofman1:
:shipit:

src/NuGet.Services.Configuration/ConfigurationFactory.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// <summary>
+        /// Injects a property of <param name="instance">instance</param> specified by <param name="property">a <see cref="PropertyDescriptor"/></param> with configuration.
+        /// </summary>
+        /// <typeparam name="T">Type of <param name="instance">the instance</param>.</typeparam>
+        /// <typeparam name="TP">Type of <param name="property">the property</param>.</typeparam>
+        /// <param name="instance">Instance to inject configuration into a property of.</param>
+        /// <param name="property"><see cref="PropertyDescriptor"/> that describes the property to inject the configuration into.</param>
+        /// <returns>A task that completes when the property has been injected into.</returns>
+        public async Task InjectPropertyWithConfiguration<T, TP>(T instance, PropertyDescriptor property)
```
scottbommarito: It's public because I use reflection to access it, and reflection can only access public methods.

src/NuGet.Services.Configuration/ConfigurationFactory.cs
scottbommarito: The `ConfigurationKeyPrefixAttribute` can now be applied to the entire `Configuration` class.

PR #13, opened 2016-11-17
ISecretWriter

scottbommarito:
Added `ISecretWriter`, which allows setting and deleting of Secrets. It will be used by the Secret Rotation job. 

`SetSecretAsync` - Sets the value and tags of a secret and returns the `Secret` that was set. Has two overloads, one which takes in the name, value, and tags directly and another that takes in a `Secret`.

`DeleteSecretAsync` - Deletes a secret and returns the `Secret` that was deleted. Has two overloads, one which takes in the name directly and another that takes in a `Secret`.

I also made several changes to the library necessary to allow writing of secrets.

- Created a `Secret` class, which mostly serves as a layer of abstraction between the consumers of the library and `Microsoft.Azure.KeyVault.Secret`. Additionally, if another source of secrets comes up, it will wrap secrets from those sources as well.
- Modified `GetSecretAsync` in `ISecretReader` to return `Secret`s and not the value directly. Otherwise, it would be impossible to get the tags from the secret. Also added an overload to the method that takes in a `Secret` for simplicity of use.
- Modified `SecretInjector` and some of our unit tests to properly handle the new return type.

Technically, these are breaking changes to the API, but since none of our projects actually use `ISecretReader` directly, instead supplying it to a `SecretInjector`, there should be no necessary code changes.

Also, made a minor fix to `ConfigurationFactory` to prevent an exception from `IConfigurationProvider` from being masked by an error message that does not apply to it.

### DIFF
--- a/src/NuGet.Services.KeyVault/KeyVaultWriter.cs
+++ b/src/NuGet.Services.KeyVault/KeyVaultWriter.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
+    public class KeyVaultWriter : KeyVaultReader, ISecretWriter
+    {
+        public KeyVaultWriter(KeyVaultConfiguration configuration) : 
+            base(configuration)
+        {
+        }
+
+        public Task<Secret> SetSecretAsync(Secret secret)
+        {
+            return SetSecretAsync(secret.Name, secret.Value, secret.Tags);
+        }
+        
+        public async Task<Secret> SetSecretAsync(string secretName, string value, Dictionary<string, string> tags = null)
+        {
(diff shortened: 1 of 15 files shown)


src/NuGet.Services.KeyVault/KeyVaultWriter.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            return DeleteSecretAsync(secret.Name);
+        }
+
+        public async Task<Secret> DeleteSecretAsync(string secretName)
+        {
+            return new KeyVaultSecret(await KeyVaultClient.Value.DeleteSecretAsync(Vault, secretName));
+        }
+    }
+}
```
shishirx34: I find it weird that `Set` and `Delete` functions return the secret value, rather than the `boolean` of operation was successful or not.

scottbommarito: This was to be consistent with the `KeyVault` API, which has that behavior. It may be better to return a boolean in this case, since there is much more information in the `KeyVault.Secret` than in our `Secret`.

shishirx34: I agree, not a comment on your changes but just the API.

scottbommarito:
Closed as no longer needed.

PR #16, opened 2017-01-19
Use ServerCommon as the master location for common.ps1

scottbommarito:
Currently, each repo that uses `common.ps1` has a copy of it, which has lead to a the individual `common.ps1` files slowly drifting apart as they have been changed.

This PR introduces the following changes to fix this issue.

- Merge all individual `common.ps1` files into the `common.ps1` in this repo.
- Introduce an `init.ps1` file whose purpose is to download all of our common build utilities.

Now, our build scripts merely have to download and run `init.ps1` to completely initialize our build utilities.
```
wget -Uri "https://raw.githubusercontent.com/NuGet/ServerCommon/master/build/init.ps1" -OutFile "$PSScriptRoot/build/init.ps1"

. "$PSScriptRoot/build/init.ps1"
```
We can remove all copies of `common.ps1` and `7zip` from our repositories.


build/common.ps1
```diff
@@ -188,14 +232,71 @@ Function Install-NuGet {
 		Trace-Log 'Creating folder "$($NuGetFolderPath)"'
 		New-Item $NuGetFolderPath -Type Directory | Out-Null
 	}
-	else {
-		Trace-Log 'Target folder "$($NuGetFolderPath)" already exists.'
+
+	$CredentialProviderBundle = (Join-Path $NuGetClientRoot '.nuget\CredentialProviderBundle.zip')
```
ryuyu: This doesn't look like it belongs here? I'm a little bit confused why this credentials bundle is pulled here.

scottbommarito: My understanding is that the VSTS Credential Provider is required to authenticate with any package feeds on VSTS. It was like this in one of the `common.ps1` files which downloads packages from VSTS. We pull it down here so that if we need to authenticate with any VSTS feeds during our build, we don't have to call a separate method.

scottbommarito: (Note that all of our `common.ps1` files have [this comment](https://github.com/NuGet/ServerCommon/pull/16/files#diff-aceb12561cec7a1da6166dd1739177b2R226), even though they didn't actually download the VSTS Credential Provider.)

build/common.ps1
```diff
@@ -337,13 +448,16 @@ Function New-Package {
 	param(
 		[Alias('target')]
 		[string]$TargetFilePath,
+        [string]$TargetProfile,
```
xavierdecoster: indentation seems off

build/common.ps1
```diff
@@ -179,21 +180,121 @@ Function Build-Solution {
...
+        $opts = 'config', '-f', "$NuGetClientRoot\.gitmodules", "submodule.$SubmodulePath.branch", "$BranchToUse"
+        Trace-Log "git $opts"
+        & cmd /c "git $opts 2>&1"
+    } else {
+        Trace-Log "No submodule path specified! Using pre-existing submodule configuration."
+    }
+    
+    $opts = 'submodule', 'update'
+    $opts += '--init'
+    $opts += '--remote'
```
shishirx34: nit: why not simply do
```
$opts = 'submodule', 'update', '--init', '--remote'
```

build/common.ps1
```diff
@@ -179,21 +180,121 @@ Function Build-Solution {
...
+        Error-Log "Required argument FeedName was not provided."
+    }
+
+    if (-not $Username) {
+        Error-Log "Required argument Username was not provided."
+    }
+
+    $opts = , 'sources'
+    $opts += 'update'
+    $opts += '-NonInteractive'
```
shishirx34: you can remove initial `,` is that needed?
```
$opts = 'sources', 'update', '-NonInteractive'
```

PR #19, opened 2017-01-26
Use AssemblyInfo.g.cs pattern instead of appending to AssemblyInfo.cs

scottbommarito:
This is necessary so that building all our projects will never lead to a changes on Git.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -65,9 +65,9 @@ Invoke-BuildStep 'Clearing artifacts' { Clear-Artifacts } `
 
 Invoke-BuildStep 'Set version metadata in AssemblyInfo.cs' { `
     param($Version, $Branch, $Commit)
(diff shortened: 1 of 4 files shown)


build.ps1
```diff
@@ -100,7 +100,6 @@ Invoke-BuildStep 'Patching versions of artifacts' {`
         Trace-Log "Patching versions of NuGet packages to $SemanticVersion"
         
         & $NupkgWrenchExe release "$Artifacts" --new-version $SemanticVersion
-        & $NupkgWrenchExe release "$Artifacts\NuGet.Services.Configuration.$SemanticVersion.nupkg" --new-version $SemanticVersion --id "NuGet.Services.KeyVault"
```
scottbommarito: No idea why this was here but it was an invalid path and never doing anything. Additionally, we'd never want to change the id of the `NuGet.Services.Configuration` pkg to be `NuGet.Services.KeyVault`

PR #22, opened 2017-02-17
Move from .xproj to .csproj + project.json

joelverhagen:
Using .csproj instead of .xproj allows for a consistent delay signing setup, which I am working on. My plan is to move all server projects that are .xproj to .csproj (either project.json or packages.config depending on sibling projects).

.csproj + project.json is preferred but I will favor consistency when there are a lot of other projects in the repo that are packages.config.


joelverhagen:
Addresses https://github.com/NuGet/Engineering/issues/261.

joelverhagen:
> Before you merge, kick off a CI build on VSTS on your branch and make sure it works.

Already done.

PR #24, opened 2017-02-18
Add a step to get private build tools from another repository

joelverhagen:
The goal here is to pull in a private repository which contains our .snk.

The repo is `PRIVATE_BUILD_TOOLS_REPO` and the commit is `PRIVATE_BUILD_TOOLS_COMMIT` (optional, defaulting to our current correct value).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ bld/
 # Files created and used by our build scripts
 AssemblyInfo.*.cs
 .nuget/CredentialProviderBundle.zip
+.nuget/CredentialProvider.VSS.exe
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -25,6 +25,9 @@ bld/
 # Files created and used by our build scripts
 AssemblyInfo.*.cs
 .nuget/CredentialProviderBundle.zip
+.nuget/CredentialProvider.VSS.exe
```
scottbommarito: We may need to add this to all of our other repos as well...

PR #25, opened 2017-02-22
Remove ReleaseLabel parameter and clean up build.ps1

joelverhagen:
We had pointless `param()`. Removed those too.


joelverhagen:
> Can you do this for the other build scripts as well?

I'll do this as I bump the commit hash for signing.

PR #26, opened 2017-03-15
Improve submodule support

dtivel:
With the addition of a new git submodule in our deployment repo, it became necessary to rewrite our submodule helper function to support greater configuration per submodule.

This change:

* Splits `Update-Submodules` into a couple of helper functions (`Reset-Submodules` and `Update-Submodule`).
* Deduplicates git invocation code through an `Invoke-Git` helper function.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -205,7 +205,7 @@ Function Update-Submodule {
         [string] $Name,
         [string] $Path,
         [string] $Branch,
-        [string] $Url
+        [string] $RemoteUrl
     )
 
     Trace-Log "Configuring submodule $Name ($Path) to use branch $Branch."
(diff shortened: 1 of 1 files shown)


build/common.ps1
```diff
@@ -181,45 +181,57 @@ Function Build-Solution {
...
+    Invoke-Git -Arguments $args
+}
+
+Function Update-Submodule {
+    [CmdletBinding()]
+    Param(
+        [string] $Name,
+        [string] $Path,
+        [string] $Branch,
+        [string] $Url
```
scottbommarito: I would rename this variable to `RemoteUrl` so that it's clearer what it is.

PR #29, opened 2017-04-18
Modify build scripts to support MSBuild 15

scottbommarito:
Fixes: https://github.com/NuGet/NuGetGallery/issues/3789

Solution presented here is this:

1 - Created a wrapper project for `Microsoft.VisualStudio.Setup.Configuration.Interop` that acquires the installation path of the VS17 instances (wasn't able to use `Microsoft.VisualStudio.Setup.Configuration.Interop` in PowerShell directly)
2 - Changed `Get-MSBuild` to download the wrapper project's assembly from our MyGet feed and use it if the MSBuild version requested is 15 or greater.
3 - Changed the other build methods to be consistent with regards to which `MSBuild` version is being used.

Steps after merging this:
1 - Build version 1.0.0 of `NuGet.VisualStudio.Setup.Configuration.Interop` and publish it to MyGet.
2 - Pin the version.

If anyone could figure out a way to use the `Microsoft.VisualStudio.Setup.Configuration.Interop` package directly without a wrapper project that would be awesome, but I wasn't able to.


src/NuGet.VisualStudio.Setup.Configuration.Interop/VisualStudioSetupConfigurationHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+            var e = query.EnumAllInstances();
+            int fetched;
+
+            var instances = new ISetupInstance[1];
+            do
+            {
+                e.Next(1, instances, out fetched);
+                if (fetched > 0)
+                {
+                    instancePaths.Add(instances[0].GetInstallationPath());
```
joelverhagen: Does this return dev14?

scottbommarito: Nope.

PR #33, opened 2017-04-19
Merge dev to master

scottbommarito:


NuGet.Server.Common.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26328.2
+VisualStudioVersion = 15.0.26410.1
```
ryuyu: We should really see if there's any consequence for removing this line or something. Seems to update with every VS version change

PR #41, opened 2017-05-06
Added Nuget.Services.Cursor project

agr:
Copied cursors from the Nuget.Services.Metadata.
Made the base classes generics.
Made the concrete classes to use `DateTimeOffset` types instead of `DateTime` - migration should be pretty easy since the serialization is done with `dateTime.ToString("O")` and serialized string can be deserialized to `DateTimeOffset` without any additional effort.
Added tests for the `DurableCursor` class.

### DIFF
--- a/NuGet.Server.Common.sln
+++ b/NuGet.Server.Common.sln
@@ -30,6 +30,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.Storage", "s
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.Cursor", "src\NuGet.Services.Cursor\NuGet.Services.Cursor.csproj", "{3269C147-FA0D-4865-83E2-50897C557C82}"
 EndProject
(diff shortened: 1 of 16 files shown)


src/NuGet.Services.Cursor/DurableCursor.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class DurableCursor : ReadWriteCursor<DateTimeOffset>
+    {
+        Uri _address;
+        Storage.Storage _storage;
+        DateTimeOffset _defaultValue;
+
+        public DurableCursor(Uri address, Storage.Storage storage, DateTimeOffset defaultValue)
+        {
+            _address = address;
```
skofman1: null checks

agr:
@scottbommarito added your precious `AggregateCursor` with some edits.

NuGet.Server.Common.sln
```diff
@@ -1,7 +1,11 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
+<<<<<<< HEAD
```
scottbommarito: gasp

PR #42, opened 2017-05-15
Add [Serializable] to ConfigurationNullOrEmptyException

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/3818

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <config>
+    <add key="repositorypath" value=".\packages" />
+  </config>
+  <packageRestore>
+    <add key="enabled" value="True" />
+    <add key="automatic" value="True" />
+  </packageRestore>
+  <packageSources>
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="nugetbuild" value="https://dotnet.myget.org/F/nuget-build/api/v3/index.json" />
+    <add key="dotnetcore" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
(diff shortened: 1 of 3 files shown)


tests/NuGet.Services.Logging.Tests/NuGet.Services.Logging.Tests.csproj
```diff
@@ -46,5 +46,8 @@
       <Name>NuGet.Services.Logging</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
scottbommarito: This was automatically added by Visual Studio.

It seems to be in all of the other test `.csproj` files as well so it must be fine.

NuGet.config
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
scottbommarito: Added this because we were missing it. My build was failing repeatably until I added this file, which we should have in all of our repos.

NuGet.config
```diff
@@ -0,0 +1,18 @@
...
+<configuration>
+  <config>
+    <add key="repositorypath" value=".\packages" />
+  </config>
+  <packageRestore>
+    <add key="enabled" value="True" />
+    <add key="automatic" value="True" />
+  </packageRestore>
+  <packageSources>
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
```
joelverhagen: Clean this up as spoken offline about

joelverhagen: Oh, already fixed.

PR #45, opened 2017-05-18
Added ForceSslMiddleware and tests

agr:
An Owin middleware that forces HTTPS for any GET and HEAD requests passing through. All other methods produce 400 SSL Required response.
Shared code for the Gallery and Search service.
Needed for [#3795](https://github.com/NuGet/NuGetGallery/issues/3795) and [#3975](https://github.com/NuGet/NuGetGallery/issues/3975).


src/NuGet.Services.ForceSslMiddleware/ForceSslMiddleware.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace NuGet.Services.ForceSslMiddleware
+{
+    public class ForceSslMiddleware : OwinMiddleware
+    {
+        private readonly int _sslPort;
+
+        public ForceSslMiddleware(OwinMiddleware next, int sslPort) : base(next)
+        {
+            _sslPort = sslPort;
```
skofman1: null check

skofman1:
We currently have ForceSslWhenAuthenticatedMiddleware. Seems like that class and the cookie it places are no longer needed, since we force SSL on everything, and not only authenticated requests, do we still need that middleware?

skofman1:
Additional question: do you have an estimation on how this will impact customers? How many requests we get with GET/HEAD that come from http?

agr:
This middleware completely supersedes the `ForceSslWhenAuthenticatedMiddleware`, so we should be able to drop it. Also the `RequireSslAttribute` would no longer be needed as we'd always have the HTTPS.

agr:
Some statistics:
We are getting 2-2.5 million requests over HTTP per day (not HTTPS). Almost all of them are GET requests and will be properly redirected. Also, there are ~5K HEAD requests that also would be properly redirected.

We are getting about 10K POST requests per day, most of the are $batch OData requests from a single IP address that result in HTTP 301, because they are made to nuget.org (and get redirected to www.nuget.org), seems to be not properly processed by whoever sends them. We also get about 20-25 legit POST requests (that result in 2xx response). These would start to produce errors, so clients would have to switch to HTTPS.

We also get the DELETE, OPTIONS, PROPFIND, PUT and TRACE requests, usually less than 10 per day. Most of them either result in 404, or produce the same response as GET, so we can ignore them.

PR #46, opened 2017-05-18
Support specifying KeyVault certificate directly rather than by thumbprint

scottbommarito:
The V3 monitoring job must use another means to find the certificate rather than by the thumbprint. This change will allow me to find the certificate myself and then pass it to my `KeyVaultReader`.

### DIFF
--- a/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
+++ b/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
@@ -11,12 +11,7 @@ namespace NuGet.Services.KeyVault
         public string VaultName { get; }
         public string ClientId { get; }
         public X509Certificate2 Certificate { get; }
-
-        public KeyVaultConfiguration(string vaultName, string clientId, string certificateThumbprint, StoreName storeName, StoreLocation storeLocation, bool validateCertificate)
-            : this(vaultName, clientId, FindCertificateByThumbprint(storeName, storeLocation, certificateThumbprint, validateCertificate))
-        {
-        }
-
+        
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -10,12 +10,14 @@ public class KeyVaultConfiguration
     {
         public string VaultName { get; }
         public string ClientId { get; }
-        public string CertificateThumbprint { get; }
-        public StoreName StoreName { get; }
-        public StoreLocation StoreLocation { get; set; }
-        public bool ValidateCertificate { get; }
+        public X509Certificate2 Certificate { get; }
```
skofman1: A configuration class is not supposed to contain any logic, but to be a simple collection of parameters for the class. Please move FindCertificateByThumbprint back.

scottbommarito: I don't think it makes sense in `KeyVaultReader` anymore, given that the class now takes the certificate from this configuration object. Should I make a static utility class to put this function in instead?

joelverhagen: I think that would be better, yes. `CertificateUtility` or something like that.

joelverhagen: After this change there should only be one ctor whose params are 1:1 with properties.

scottbommarito: done

PR #47, opened 2017-05-22
Tweaked TelemetryResponseCodeProcessor 

loic-sharma:
Application Insight's configuration file doesn't play nice with configurable properties that are arrays. This changes the `TelemetryResponseCodeProcessor` to use an `IList` to follow the pattern set by Application Insight's `RequestTrackingTelemetryModule`'s `Handlers` configurable property.

### DIFF
--- a/src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
+++ b/src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
@@ -2,10 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.ApplicationInsights.Extensibility;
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -2,10 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DataContracts;
+using System.Collections.Generic;
```
scottbommarito: `System` before everything else

PR #49, opened 2017-05-22
ForceSslMiddleware now allows to exclude certain paths from redirection.

agr:
Added the exclusion list to the ForceSslMiddleware.
Added tests for the change.

### DIFF
--- a/src/NuGet.Services.Owin/ForceSslMiddleware.cs
+++ b/src/NuGet.Services.Owin/ForceSslMiddleware.cs
@@ -2,8 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Owin;
 
@@ -12,17 +15,34 @@ namespace NuGet.Services.Owin
     public class ForceSslMiddleware : OwinMiddleware
     {
         private readonly int _sslPort;
+        private readonly List<Regex> _exclusionPathPatterns;
 
-        public ForceSslMiddleware(OwinMiddleware next, int sslPort) : base(next)
+        public ForceSslMiddleware(OwinMiddleware next, int sslPort)
+            : this(next, sslPort, Enumerable.Empty<Regex>())
+        {
+        }
+
+        public ForceSslMiddleware(
+            OwinMiddleware next, 
+            int sslPort, 
+            IEnumerable<Regex> exclusionPathPatterns)
+            : base(next ?? throw new ArgumentNullException(nameof(next)))
         {
             _sslPort = sslPort;
+            _exclusionPathPatterns = exclusionPathPatterns.ToList();
         }
 
         public override async Task Invoke(IOwinContext context)
         {
-            if (!context.Request.IsSecure)
+            bool shouldPassThrough = context.Request.IsSecure 
+                || (context.Request.Path.HasValue && IsExcludedPath(context.Request.Path.Value));
+            if (shouldPassThrough)
             {
-                if (context.Request.Method == HttpMethod.Get.Method || context.Request.Method == HttpMethod.Head.Method)
+                await Next.Invoke(context);
+            }
+            else
+            {
+                if (IsAllowedMethod(context.Request.Method))
                 {
                     context.Response.Redirect(new UriBuilder(context.Request.Uri)
                     {
@@ -36,10 +56,16 @@ namespace NuGet.Services.Owin
                     context.Response.ReasonPhrase = "SSL Required";
                 }
             }
-            else
-            {
-                await Next.Invoke(context);
-            }
+        }
+
+        private bool IsExcludedPath(string path)
+        {
+            return _exclusionPathPatterns.Any(p => p.IsMatch(path));
+        }
+
+        private static bool IsAllowedMethod(string method)
+        {
+            return method == HttpMethod.Get.Method || method == HttpMethod.Head.Method;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.Owin/ForceSslMiddleware.cs
```diff
@@ -36,10 +55,16 @@ public ForceSslMiddleware(OwinMiddleware next, int sslPort) : base(next)
...
+        }
+
+        private bool IsExcludedPath(string path)
+        {
+            return _exclusionPathPatterns.Contains(path);
+        }
+
+        private static bool IsAllowedMethod(string method)
+        {
+            return method == HttpMethod.Get.Method || method == HttpMethod.Head.Method;
```
scottbommarito: might be a bit more elegant if you store this as a list

```
private static string[] AllowedMethods
{
    HttpMethod.Get.Method,
    HttpMethod.Head.Method
}
```

and then above do something like
```
if (AllowedMethods.IndexOf(context.Request.Method) >= 0)
```

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
```diff
@@ -55,6 +57,38 @@ public class ForceSslMiddlewareFacts
...
+            await middleware.Invoke(context);
+
+            Assert.Equal((int)HttpStatusCode.OK, context.Response.StatusCode);
+        }
+
+        [Theory]
+        [InlineData("http://localhost/", new[] { "/health" })]
+        [InlineData("http://localhost/search/diag", new[] { "/" })]
+        [InlineData("http://localhost/somepath?something=somevalue", new[] { "/", "/SomeOtherPath" })]
+        public async Task RedirectsNotExludedUrls(string url, IEnumerable<string> excludedPaths)
```
scottbommarito: typo: `RedirectsNotExludedUrls` -> `RedirectsNotExcludedUrls`

src/NuGet.Services.Owin/ForceSslMiddleware.cs
```diff
@@ -12,17 +14,34 @@ namespace NuGet.Services.Owin
...
 
         public override async Task Invoke(IOwinContext context)
         {
-            if (!context.Request.IsSecure)
+            bool shouldPassThrough = context.Request.IsSecure 
+                || (context.Request.Path.HasValue && IsExcludedPath(context.Request.Path.Value));
+            if (shouldPassThrough)
             {
-                if (context.Request.Method == HttpMethod.Get.Method || context.Request.Method == HttpMethod.Head.Method)
+                await Next.Invoke(context);
```
scottbommarito: to reduce nesting

```
if (shouldPassThrough)
{
    return await Next.Invoke(context);
}

if (IsAllowedMethod(context.Request.Method))
{
    ...
}
```

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
```diff
@@ -55,6 +57,38 @@ public class ForceSslMiddlewareFacts
...
         }
 
+        [Theory]
+        [InlineData("http://localhost/", new[] { "/" })]
+        [InlineData("http://localhost/Search/Diag", new[] { "/", "/search/diag" })]
+        [InlineData("http://localhost/somepath?something=somevalue", new[] { "/", "/SomePath" })]
+        public async Task RespectsExclusionList(string url, IEnumerable<string> excludedPaths)
+        {
+            var uri = new Uri(url);
+            var context = CreateOwinContext("GET", uri);
```
scottbommarito: should we test this with other http methods as well?

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
```diff
@@ -55,6 +57,38 @@ public class ForceSslMiddlewareFacts
...
+        }
+
+        [Theory]
+        [InlineData("http://localhost/", new[] { "/health" })]
+        [InlineData("http://localhost/search/diag", new[] { "/" })]
+        [InlineData("http://localhost/somepath?something=somevalue", new[] { "/", "/SomeOtherPath" })]
+        public async Task RedirectsNotExludedUrls(string url, IEnumerable<string> excludedPaths)
+        {
+            var uri = new Uri(url);
+            var context = CreateOwinContext("GET", uri);
```
scottbommarito: should we test this with other http methods as well?

src/NuGet.Services.Owin/ForceSslMiddleware.cs
agr: Since the method is `async` and its return type is simply `Task` I can't really do that. Compiler does not expect a `Task` as a value to return and produces the following error:
```
Since 'ForceSslMiddleware.Invoke(IOwinContext)' is an async method that returns 'Task', a return keyword must not be followed by an object expression. Did you intend to return 'Task<T>'?
```
I could make it non-`async`, but then I'd have to add some ugly return statement at the end of the method. I chose to leave it like that.

joelverhagen: You would `await` on one line then `return;` on the next.

src/NuGet.Services.Owin/ForceSslMiddleware.cs
joelverhagen: If you do this use an `ISet` for O(1) lookup time. `IndexOf` is O(n).

agr: I have thread safety concerns with that proposal: I can't find any documentation stating that `IndexOf` or `Contains` methods are thread safe. Could you point me to one?

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
agr: done

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
agr: fixed

tests/NuGet.Services.Owin.Tests/ForceSslMiddlewareFacts.cs
agr: done

PR #51, opened 2017-06-02
Reduce exception telemetry noise

dtivel:
Resolve NuGet/NuGetGallery#4041.


src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -29,18 +33,91 @@ public TelemetryResponseCodeProcessor(ITelemetryProcessor next)
...
-            var request = item as RequestTelemetry;
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            ITelemetry nextItem = null;
+
+            if (TryProcess(item as RequestTelemetry))
+            {
```
xavierdecoster: Empty branch in `if-else` could be rewritten into a single `if` branch:

```csharp
if (!TryProcess(item as RequestTelemetry) 
    && TryProcess(item as ExceptionTelemetry, out nextItem))
{
  item = nextItem;
}

_next.Process(item);
```

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -29,18 +33,89 @@ public TelemetryResponseCodeProcessor(ITelemetryProcessor next)
 
         public void Process(ITelemetry item)
         {
-            var request = item as RequestTelemetry;
+            if (item == null)
```
skofman1: I wonder if we should just ignore the case when item==null and succeed

dtivel: I think a `null` argument would be a bug.  Application Insights internally inserts its own telemetry processor as the last in the chain of telemetry processors.  This processor would start [here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/309ff76563b7db90cb26c049f9c671c5a61fb8dd/src/Core/Managed/Shared/Extensibility/Implementation/TransmissionProcessor.cs#L40) and then dereference `null` [here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/309ff76563b7db90cb26c049f9c671c5a61fb8dd/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryDebugWriter.cs#L36), so it's reasonable to assert that `item` is not `null`.

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -29,18 +33,89 @@ public TelemetryResponseCodeProcessor(ITelemetryProcessor next)
...
+
+                    return true;
                 }
             }
 
-            _next.Process(item);
+            return false;
+        }
+
+        private static bool TryProcess(ExceptionTelemetry telemetry, out ITelemetry nextTelemetry)
```
loic-sharma: Why are `TryProcess` and `Convert` static?

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -29,18 +33,89 @@ public TelemetryResponseCodeProcessor(ITelemetryProcessor next)
...
+                    var httpCode = httpException.GetHttpCode();
+
+                    if (httpCode < 500)
+                    {
+                        // Logging exception telemetry for a request that resulted in an HTTP response code under 500
+                        // adds noise to exception telemetry analysis.  Log it as trace telemetry instead.
+                        nextTelemetry = Convert(telemetry, httpException);
+                    }
+
+                    return true;
```
loic-sharma: Shouldn't you only return `true` if the `httpCode` is less than 500? If you make that change, you can move the `nextTelemetry` assignment at the beginning of the method to the very end and avoid an unnecessary assignment.

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
```diff
@@ -29,18 +33,89 @@ public TelemetryResponseCodeProcessor(ITelemetryProcessor next)
...
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            ITelemetry nextItem = null;
+
+            if (!TryProcess(item as RequestTelemetry) &&
+                TryProcess(item as ExceptionTelemetry, out nextItem))
+            {
+                item = nextItem;
```
loic-sharma: What if we have AI initializers/processors registered before this `TelemetryResponseCodeProcessor` that have special logic `TraceTelemetry`? Could we instead drop the current `ExceptionTelemetry` item and emit a new `TraceTelemetry` through AI's `TrackTrace`?

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
dtivel: Because instance methods are unnecessary.  The methods are stateless.

src/NuGet.Services.Logging/TelemetryResponseCodeProcessor.cs
dtivel: Each `TryProcess(...)` method attempts to process an `ITelemetry` instance as a specific implementation (e.g.:  `RequestTelemetry`, `ExceptionTelemetry`, etc.).  This method is for processing `ExceptionTelemetry` instances.  Additional handling could be added here in the future.  If an `ITelemetry` instance is processed as an `ExceptionTelemetry` but we return `false` as you suggest, we shouldn't then try processing it as some other implementation (e.g.:  `TraceTelemetry`).

loic-sharma: Ah, gotcha, that makes sense.

chenriksson: Sounds like it's worth putting this in a comment

loic-sharma: Yes, either that or split up the `TelemetryResponseCodeProcessor` - one for `RequestTelemetry` and one for `ExceptionTelemetry`.

PR #54, opened 2017-08-08
CachingSecretReader should cache secrets as SecureString

scottbommarito:
`SecureString` encrypts the strings so that they are encrypted at rest in memory.

https://github.com/NuGet/Engineering/issues/667

This appeared to be the only thing in this repository that was caching secrets.

### DIFF
--- a/src/NuGet.Services.KeyVault/CachingSecretReader.cs
+++ b/src/NuGet.Services.KeyVault/CachingSecretReader.cs
@@ -3,6 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Security;
 using System.Threading.Tasks;
 
 namespace NuGet.Services.KeyVault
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -23,12 +26,20 @@ public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec =
...
 
             _internalReader = secretReader;
-            _cache = new Dictionary<string, Tuple<string, DateTime>>();
+            _cache = new Dictionary<string, Tuple<SecureString, DateTime>>();
 
             _refreshIntervalSec = refreshIntervalSec;
         }
 
-        public virtual bool IsSecretOutdated(Tuple<string, DateTime> cachedSecret)
+        ~CachingSecretReader()
```
agr: Destructors in C#? Blasphemy!

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -39,10 +50,39 @@ public virtual bool IsSecretOutdated(Tuple<string, DateTime> cachedSecret)
...
             }
 
-            return _cache[secretName].Item1;
+            return SecureStringToString(_cache[secretName].Item1);
+        }
+
+        private static SecureString StringToSecureString(string input)
+        {
+            var output = new SecureString();
+            input.ToCharArray().ToList().ForEach(c => output.AppendChar(c));
```
agr: Shouldn't `input.ForEach(c => output.AppendChar(c));` also work? string already implements `IEnumerable<char>` why additional conversions?

PR #58, opened 2017-09-20
Add IValidator and ValidationStatus to Contracts project

joelverhagen:
And rename NuGet.Services.Abstractions to NuGet.Services.Contracts.

The validator interface may change as we continue in orchestrator implementation, but this is a first stab.


src/NuGet.Services.Contracts/Validation/IValidator.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="packageId">The package ID associated with this <paramref name="validationId"/>.</param>
+        /// <param name="packageVersion">
+        /// The package version (not necessarily normalized) associated with this <paramref name="validationId"/>.
+        /// </param>
+        /// <param name="nupkgUrl">The URL to the NuGet package content (the .nupkg).</param>
+        /// <returns>
+        /// A task returning the current validation status indicating that the validation has been started and possibly
+        /// already completed.
+        /// </returns>
+        Task<ValidationStatus> StartValidationAsync(Guid validationId, string packageId, string packageVersion, string nupkgUrl);
```
loic-sharma: >StartValidationAsync [](start = 31, length = 20)

Don't forget the package's key :)

joelverhagen: Added.

PR #59, opened 2017-09-20
Initialize validation DB context, entities, and migrations

joelverhagen:
Depends on https://github.com/NuGet/ServerCommon/pull/58. Supersedes https://github.com/NuGet/NuGetGallery/pull/4690.

After talking to @loic-sharma about his own validator needs and thinking more about monitoring, I think it's best to start out by putting validation entities in their own DB.

Monitoring query patterns or indexing needs by orchestrator should not put unnecessary load on the Gallery DB.

As we add more validators and are implementing, `GetStatusAsync` validators can choose to put their state in the new validator DB. VCS happens to put its state in blob storage, but other validators can use SQL for ease in monitoring which may necessitate complex query patterns which SQL can do easily.


src/NuGet.Services.Validation/Entities/PackageValidationSet.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// of the <see cref="ValidationTrackingId"/> are decoupled.
+        /// </summary>
+        public DateTime Created { get; set; }
+
+        /// <summary>
+        /// The time when the application last updated the package validation set. When the set is first created, the 
+        /// <see cref="Created"/> timestamp shares the same value. When a validation is added to the validation set
+        /// after the set is created, this timestamp is updated.
+        /// </summary>
+        public DateTime Updated { get; set; }
```
agr: Should we use a DB trigger to update this column?..

src/NuGet.Services.Validation/Entities/PackageValidation.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// there could be multiple validations within a set with the same type (but different <see cref="Key"/>).
+        /// This value has a maximum length of 255 characters.
+        /// </summary>
+        public string Type { get; set; }
+
+        /// <summary>
+        /// The time when the validation was started. This can be null if the validation has not been started yet, as
+        /// indicared by the <see cref="ValidationStatus.NotStarted"/> status.
+        /// </summary>
+        public DateTime? Started { get; set; }
```
chenriksson: Will this be relative to ValidationStatus, same as ValidationStatusTimestamp? i.e., are these started/completed timestamps for the current validation iteration? Or is this the absolute start time?

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Data.Entity;
+using System.Data.Entity.Infrastructure.Annotations;
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// The Entity Framework database context for validation entities.
+    /// </summary>
+    [DbConfigurationType(typeof(EntitiesConfiguration))]
+    public class ValidationContext : DbContext
```
chenriksson: nit: wondering if EntitiesContext (or ValidationEntitiesContext) would be better name. Validation is redundant with the namespace, and it's not apparent from the name that this is an EF related class.

src/NuGet.Services.Validation/Entities/PackageValidation.cs
joelverhagen: Both `Started` and `ValidationStatusTimestamp` are absolute `DateTime` values. They will have the same value when the validation record first moves to the `Incomplete` validation status.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
joelverhagen: Hm, good point. I like `ValidationEntitiesContext`.

src/NuGet.Services.Validation/Entities/PackageValidationSet.cs
joelverhagen: This is slightly a hack. We will use this to force an `UPDATE` and there trigger optimistic concurrency when implementing auto-retry.

src/NuGet.Services.Validation/Entities/PackageValidation.cs
joelverhagen: Concerning the "current validation iteration" - not exactly sure what you mean here. If an admin uses the "revalidate" gesture on a package or if orchestrator implementats "auto-retry", there will be an entirely new `PackageValidation` record.

chenriksson: Got it - thought a retry would use the same validation.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
joelverhagen: Fixed.

src/NuGet.Services.Validation/Entities/PackageValidation.cs
joelverhagen: Clarifed comments.

PR #60, opened 2017-09-25
Add ETag support to Storage APIs

scottbommarito:
Currently there's no way to do optimistic concurrency with our storage calls.

The V3 monitoring job uses the ETag support to allow multiple `MonitoringProcessor`s to be running at once.

### DIFF
--- a/src/NuGet.Services.Storage/AzureStorage.cs
+++ b/src/NuGet.Services.Storage/AzureStorage.cs
@@ -1,32 +1,55 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.WindowsAzure.Storage.RetryPolicies;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Blob;
 
 namespace NuGet.Services.Storage
 {
     public class AzureStorage : Storage
     {
-        private readonly ILogger<AzureStorage> _logger;
+        private readonly TimeSpan _defaultServerTimeout = TimeSpan.FromSeconds(30);
         private readonly CloudBlobDirectory _directory;
+        private readonly BlobRequestOptions _blobRequestOptions;
 
-        public AzureStorage(CloudStorageAccount account, string containerName, string path, Uri baseAddress, ILogger<AzureStorage> logger)
-            : this(account.CreateCloudBlobClient().GetContainerReference(containerName).GetDirectoryReference(path), baseAddress, logger)
+
+        public AzureStorage(CloudStorageAccount account,
+                            string containerName,
+                            string path,
+                            Uri baseAddress,
+                            ILogger<AzureStorage> logger)
+            : this(account, containerName, path, baseAddress, DefaultMaxExecutionTime, logger)
         {
         }
 
-        private AzureStorage(CloudBlobDirectory directory, Uri baseAddress, ILogger<AzureStorage> logger)
+        public AzureStorage(CloudStorageAccount account,
+                           string containerName,
+                           string path,
+                           Uri baseAddress,
+                           TimeSpan maxExecutionTime,
+                           ILogger<AzureStorage> logger)
+           : this(account.CreateCloudBlobClient().GetContainerReference(containerName).GetDirectoryReference(path),
+                 baseAddress,
+                 maxExecutionTime,
+                 logger)
+        {
+        }
+
+        private AzureStorage(CloudBlobDirectory directory, 
+                             Uri baseAddress, 
+                             TimeSpan maxExecutionTime, 
+                             ILogger<AzureStorage> logger)
             : base(baseAddress ?? GetDirectoryUri(directory), logger)
         {
-            _logger = logger;
             _directory = directory;
 
             if (_directory.Container.CreateIfNotExists())
@@ -35,11 +58,20 @@ namespace NuGet.Services.Storage
 
                 if (Verbose)
                 {
-                    _logger.LogInformation("Created {ContainerName} publish container", _directory.Container.Name);
+                    Logger.LogInformation("Created {ContainerName} publish container", _directory.Container.Name);
                 }
             }
 
             ResetStatistics();
+            this._blobRequestOptions = CreateBlobRequestOptions(maxExecutionTime);
+        }
+
+        public static TimeSpan DefaultMaxExecutionTime
+        {
+            get
+            {
+                return TimeSpan.FromSeconds(600);
+            }
         }
 
         public bool CompressContent
@@ -73,7 +105,7 @@ namespace NuGet.Services.Storage
             }
             if (Verbose)
             {
-                _logger.LogInformation("The blob {BlobUri} does not exist.", packageRegistrationUri);
+                Logger.LogInformation("The blob {BlobUri} does not exist.", packageRegistrationUri);
             }
             return false;
         }
@@ -87,13 +119,31 @@ namespace NuGet.Services.Storage
 
         private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
         {
-            var lastModified = (listBlobItem as CloudBlockBlob)?.Properties.LastModified?.UtcDateTime;
+            DateTime? lastModified = null;
+            string eTag = null;
 
-            return new StorageListItem(listBlobItem.Uri, lastModified);
+            var properties = (listBlobItem as CloudBlockBlob)?.Properties;
+            if (properties != null)
+            {
+                lastModified = properties.LastModified?.UtcDateTime;
+                eTag = properties.ETag;
+            }
+
+            return new StorageListItem(listBlobItem.Uri, lastModified, eTag);
         }
 
         //  save
-        protected override async Task OnSave(Uri resourceUri, StorageContent content, CancellationToken cancellationToken)
+        protected override Task OnSave(Uri resourceUri, StorageContent content, CancellationToken cancellationToken)
+        {
+            return OnSaveInternal(resourceUri, content, null, cancellationToken);
+        }
+
+        protected override Task OnSaveIfETag(Uri resourceUri, StorageContent content, string eTag, CancellationToken cancellationToken)
+        {
+            return OnSaveInternal(resourceUri, content, GetAccessConditionForETag(eTag), cancellationToken);
+        }
+
+        private async Task OnSaveInternal(Uri resourceUri, StorageContent content, AccessCondition accessCondition, CancellationToken cancellationToken)
         {
             string name = GetName(resourceUri);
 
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.Storage/Storage.cs
```diff
@@ -83,7 +99,39 @@ public override string ToString()
                 if (webException != null)
```
loic-sharma: >webException != null [](start = 20, length = 20)

FYI: You can simplify these two lines to `if (e.InnerException is WebException webException)`

src/NuGet.Services.Storage/Storage.cs
```diff
@@ -28,50 +30,64 @@ public override string ToString()
...
             catch (Exception e)
             {
-                string message = String.Format("SAVE EXCEPTION: {0} {1}", resourceUri, e.Message);
-                _logger.LogError("SAVE EXCEPTION: {ResourceUri} {Exception}", resourceUri, e);
-                throw new Exception(message, e);
+                throw LogException(nameof(Save), resourceUri, e);
+            }
+        }
+
+        public async Task SaveIfETag(Uri resourceUri, StorageContent content, string eTag, CancellationToken cancellationToken)
```
loic-sharma: >SaveIfETag [](start = 26, length = 10)

Are these methods supposed to support concurrency? If so, `SaveCount++` isn't atomic and should be fixed (for this and all other public methods).

scottbommarito: I'm following the existing pattern, but I can change that to `Interlocked.Increment`.

loic-sharma:
>             catch (StorageException e)

This code is a duplicate of the catch in `DeleteIfETag` - could we create a helper for something like `IsNotFoundStorageException`?

---
Refers to: src/NuGet.Services.Storage/Storage.cs:96 in b60fdfe. [](commit_id = b60fdfe3ba14c09e442c7ad8f4d3246a5e142155, deletion_comment = False)

src/NuGet.Services.Storage/FileStorage.cs
```diff
@@ -88,10 +86,20 @@ public string Path
...
         }
 
+        protected override async Task OnSaveIfETag(Uri resourceUri, StorageContent content, string eTag, CancellationToken cancellationToken)
+        {
+            if (await IsETagConflict(resourceUri, eTag, cancellationToken))
+            {
+                return;
+            }
+
+            await OnSave(resourceUri, content, cancellationToken);
```
loic-sharma: >OnSave [](start = 18, length = 6)

This isn't thread-safe:

1. Thread A performs the `IsETagConflict` check, it passes
2. A context switch happens to thread B
3. Thread B performs a `OnSave`/`OnSaveIfEtag` to completion on the same resource as thread A
4. A context switch happens to thread A
5. Thread A runs `OnSave` even though the ETag has now changed

Adding locks will only add in-process synchronization, different processes could run into the same problem. This isn't an easy problem to solve :(

I would clearly document this.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -114,17 +164,74 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                RetryPolicy = new ExponentialRetry()
+            };
+        }
+
+        /// <summary>
+        /// Take one snapshot only if there is not any snapshot for the specific blob
+        /// This will prevent the blob to be deleted by a not intended delete action
+        /// </summary>
+        /// <param name="blob"></param>
+        /// <returns></returns>
```
chenriksson: incomplete doc comments

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -114,17 +164,74 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+        private async Task<bool> TryTakeBlobSnapshotAsync(CloudBlockBlob blob)
+        {
+            if (blob == null)
+            {
+                //no action
+                return false;
+            }
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
```
chenriksson: Are we concerned with performance of creating snapshots? Is there concern that this will introduce a regression? If so, I would add comments about any concerns.

Stopwatch/logging is fine for development, but I think if we want to truly monitor our code performance that we should consider a feature to instrument our code with etw markers like VS does.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -114,17 +164,74 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
-                    _logger.LogInformation("Saved uncompressed blob {BlobUri} to container {ContainerName}", blob.Uri.ToString(), _directory.Container.Name);
+                    await blob.UploadFromStreamAsync(stream,
+                        accessCondition: accessCondition,
+                        options: _blobRequestOptions,
+                        operationContext: null,
+                        cancellationToken: cancellationToken);
+                    Logger.LogInformation("Saved uncompressed blob {BlobUri} to container {ContainerName}", blob.Uri.ToString(), _directory.Container.Name);
+                }
+            }
+            await TryTakeBlobSnapshotAsync(blob);
```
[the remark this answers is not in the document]
chenriksson: Do we not care if the snapshot fails? Wondering why we don't use the return value.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -114,17 +164,74 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+            }
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+            try
+            {
+                var allSnapshots = blob.Container.
+                                   ListBlobs(prefix: blob.Name,
+                                             useFlatBlobListing: true,
+                                             blobListingDetails: BlobListingDetails.Snapshots);
+                //the above call will return at least one blob the original
```
chenriksson: No other blobs will have the same prefix? Instead of relying on count, why not check CloudBlockBlob.IsSnapshot?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -165,25 +276,44 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                accessCondition: accessCondition, 
+                options: _blobRequestOptions, 
+                operationContext: null, 
+                cancellationToken: cancellationToken);
+        }
 
-            await blob.DeleteAsync(cancellationToken);
+        private AccessCondition GetAccessConditionForETag(string eTag)
+        {
+            return eTag != null ? AccessCondition.GenerateIfMatchCondition(eTag) : AccessCondition.GenerateIfNotExistsCondition();
```
chenriksson: I would remove this and move the AccessCondition to the corresponding overload... and add arg validation.

If a caller uses the etag overload, we should validate and enforce that the etag is provided. Allowing null/empty etags from the etag-specific overload just weakens the API.

src/NuGet.Services.Storage/FileStorage.cs
chenriksson: Is it really worth us implementing etag behavior for the file system? I wouldn't...

I think the thread safety is an issue even with the existing OnSave method, right? If that's a known issue, we should address that separately - maybe with introducing use a semaphore slims per file or something.

src/NuGet.Services.Storage/IStorage.cs
```diff
@@ -10,11 +11,14 @@ namespace NuGet.Services.Storage
     public interface IStorage
     {
         Task Save(Uri resourceUri, StorageContent content, CancellationToken cancellationToken);
+        Task SaveIfETag(Uri resourceUri, StorageContent content, string eTag, CancellationToken cancellationToken);
```
chenriksson: Why not keep naming consistent, just add an overload with eTag support? Same for all other APIs.

- Save(Uri, StorageContent, CancellationToken);
- Save(Uri, StorageContent, string, CancellationToken);

As mentioned previously, if you have an overload with eTag you should do arg validation on eTag to ensure it's provided... otherwise the other overload is unnecessary and the APIs are weakened.

chenriksson: Or instead of overloads, why not just put an etag property on the StorageContent? CloudStorage would support it, but FileStorage could ignore.

scottbommarito: I like the overload.

`StorageContent` has an `ETag` property, but its meant to be the `ETag` of the storage contents itself. If you wanted to save the `StorageContent` you received from one `Storage` to a different `Storage`, you would need to change the `ETag` field of the object, which seems very odd and confusing to me.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -87,13 +119,31 @@ public override bool Exists(string fileName)
...
+            {
+                lastModified = properties.LastModified?.UtcDateTime;
+                eTag = properties.ETag;
+            }
 
-            return new StorageListItem(listBlobItem.Uri, lastModified);
+            return new StorageListItem(listBlobItem.Uri, lastModified, eTag);
         }
 
         //  save
```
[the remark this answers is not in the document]
scottbommarito: was existing, but I can remove

src/NuGet.Services.Storage/AzureStorage.cs
scottbommarito: ETag can be null for "expect nothing to be there" or have any string value for "the ETag of the file MUST be this"; there really isn't any need for validation because it can be ANY valid string value.

src/NuGet.Services.Storage/AzureStorage.cs
scottbommarito: This code was from https://github.com/NuGet/NuGet.Services.Metadata/pull/177/files, I don't plan to touch it in this PR.

src/NuGet.Services.Storage/AzureStorage.cs
scottbommarito: This code was from https://github.com/NuGet/NuGet.Services.Metadata/pull/177, I don't plan to touch it in this PR.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -35,11 +58,20 @@ private AzureStorage(CloudBlobDirectory directory, Uri baseAddress, ILogger<Azur
...
 
             ResetStatistics();
+            this._blobRequestOptions = CreateBlobRequestOptions(maxExecutionTime);
+        }
+
+        public static TimeSpan DefaultMaxExecutionTime
+        {
+            get
+            {
+                return TimeSpan.FromSeconds(600);
```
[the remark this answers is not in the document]
scottbommarito: This code is from https://github.com/NuGet/NuGet.Services.Metadata/pull/222 and I don't intend to touch it in this PR.

[I will note that this same question was asked on the original PR.](https://github.com/NuGet/NuGet.Services.Metadata/pull/222/files#r131005158)

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -114,17 +164,74 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+        /// This will prevent the blob to be deleted by a not intended delete action
+        /// </summary>
+        /// <param name="blob"></param>
+        /// <returns></returns>
+        private async Task<bool> TryTakeBlobSnapshotAsync(CloudBlockBlob blob)
+        {
+            if (blob == null)
+            {
+                //no action
+                return false;
```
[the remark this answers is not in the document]
scottbommarito: This code was from https://github.com/NuGet/NuGet.Services.Metadata/pull/177, I don't plan to touch it in this PR.

src/NuGet.Services.Storage/AzureStorage.cs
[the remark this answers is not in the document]
scottbommarito: This code was from https://github.com/NuGet/NuGet.Services.Metadata/pull/177, I don't plan to touch it in this PR.

scottbommarito:
Closing this PR as it's not currently needed. If we want to implement storage ETags we can do so in the future.

PR #61, opened 2017-09-25
[Package Signature] Add database schema for Package Signatures

loic-sharma:
Fixes https://github.com/NuGet/Engineering/issues/789. Resulting migrations SQL:

```sql
CREATE TABLE [dbo].[ValidatorStatuses] (
    [ValidationId] [uniqueidentifier] NOT NULL,
    [PackageKey] [int] NOT NULL,
    [State] [int] NOT NULL,
    CONSTRAINT [PK_dbo.ValidatorStatuses] PRIMARY KEY ([ValidationId])
)

CREATE INDEX [IX_ValidatorStatuses_PackageKey] ON [dbo].[ValidatorStatuses]([PackageKey])

IF schema_id('signature') IS NULL
    EXECUTE('CREATE SCHEMA [signature]')

CREATE TABLE [signature].[Certificates] (
    [Key] [bigint] NOT NULL IDENTITY,
    [Thumbprint] [char](20) NOT NULL,
    [Status] [int] NOT NULL,
    [StatusUpdateTime] [datetime2](7),
    [NextStatusUpdateTime] [datetime2](7),
    [LastVerificationTime] [datetime2](7),
    [RevocationTime] [datetime2](7),
    [ValidationFailures] [int] NOT NULL,
    CONSTRAINT [PK_signature.Certificates] PRIMARY KEY ([Key])
)

CREATE UNIQUE INDEX [IX_Certificates_Thumbprint] ON [signature].[Certificates]([Thumbprint])

CREATE TABLE [signature].[PackageSignatures] (
    [Key] [bigint] NOT NULL IDENTITY,
    [PackageKey] [int] NOT NULL,
    [SignedAt] [datetime2](7) NOT NULL,
    [CreatedAt] [datetime2](7) NOT NULL,
    [Status] [int] NOT NULL,
    CONSTRAINT [PK_signature.PackageSignatures] PRIMARY KEY ([Key])
)

CREATE INDEX [IX_PackageSignatures_PackageKey] ON [signature].[PackageSignatures]([PackageKey])

CREATE INDEX [IX_PackageSignatures_Status] ON [signature].[PackageSignatures]([Status])

CREATE TABLE [signature].[PackageSigningStates] (
    [PackageKey] [int] NOT NULL,
    [PackageId] [nvarchar](128) NOT NULL,
    [PackageNormalizedVersion] [nvarchar](64) NOT NULL,
    [SigningStatus] [int] NOT NULL,
    CONSTRAINT [PK_signature.PackageSigningStates] PRIMARY KEY ([PackageKey])
)

CREATE INDEX [IX_PackageSigningStates_PackageId_PackageNormalizedVersion] ON [signature].[PackageSigningStates]([PackageId], [PackageNormalizedVersion])

CREATE TABLE [signature].[CertificateValidations] (
    [Key] [bigint] NOT NULL IDENTITY,
    [CertificateKey] [bigint] NOT NULL,
    [ValidationId] [uniqueidentifier] NOT NULL,
    [Status] [int],
    CONSTRAINT [PK_signature.CertificateValidations] PRIMARY KEY ([Key])
)

CREATE INDEX [IX_CertificateValidations_CertificateKey_ValidationId] ON [signature].[CertificateValidations]([CertificateKey], [ValidationId])

CREATE INDEX [IX_CertificateValidations_ValidationId] ON [signature].[CertificateValidations]([ValidationId])

CREATE TABLE [signature].[PackageSignatureCertificates] (
    [PackageSignatureKey] [bigint] NOT NULL,
    [CertificateKey] [bigint] NOT NULL,
    CONSTRAINT [PK_signature.PackageSignatureCertificates] PRIMARY KEY ([PackageSignatureKey], [CertificateKey])
)

CREATE INDEX [IX_PackageSignatureKey] ON [signature].[PackageSignatureCertificates]([PackageSignatureKey])

CREATE INDEX [IX_CertificateKey] ON [signature].[PackageSignatureCertificates]([CertificateKey])

ALTER TABLE [signature].[PackageSignatures] ADD CONSTRAINT [FK_signature.PackageSignatures_signature.PackageSigningStates_PackageKey] FOREIGN KEY ([PackageKey]) REFERENCES [signature].[PackageSigningStates] ([PackageKey]) ON DELETE CASCADE

ALTER TABLE [signature].[CertificateValidations] ADD CONSTRAINT [FK_signature.CertificateValidations_signature.Certificates_CertificateKey] FOREIGN KEY ([CertificateKey]) REFERENCES [signature].[Certificates] ([Key]) ON DELETE CASCADE

ALTER TABLE [signature].[PackageSignatureCertificates] ADD CONSTRAINT [FK_signature.PackageSignatureCertificates_signature.PackageSignatures_PackageSignatureKey] FOREIGN KEY ([PackageSignatureKey]) REFERENCES [signature].[PackageSignatures] ([Key]) ON DELETE CASCADE

ALTER TABLE [signature].[PackageSignatureCertificates] ADD CONSTRAINT [FK_signature.PackageSignatureCertificates_signature.Certificates_CertificateKey] FOREIGN KEY ([CertificateKey]) REFERENCES [signature].[Certificates] ([Key]) ON DELETE CASCADE
```


src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -122,7 +131,138 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
 
+            RegisterPackageSigningEntities(modelBuilder);
+
             base.OnModelCreating(modelBuilder);
         }
+
+        private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<Package>()
+                .ToTable(nameof(Package), SignatureSchema)
```
joelverhagen: Please hard-code table names. Class refactoring should not effect schema.

loic-sharma: Talked offline. Refactoring a class that doesn't have a name explicitly set with `ToTable` will effect schema. Thus, using `nameof` keeps this behavior consistent.

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// The SHA1 thumbprint that uniquely identifies this certificate. This is a binary string with
+        /// a maximum length of 64 bytes.
+        /// </summary>
+        public byte[] Thumbprint { get; set;}
```
joelverhagen: Thoughts on making this a hex string for human readability / query-ability?

loic-sharma:
Hey folks, I tweaked the schema a little bit as it is unknown whether signed package's will have OCSP responses embedded within them. I changed the schema so that it will work even if the packages do not have OCSP responses included with them:

* `Certificate`s will now have a `Unknown` state if they haven't been verified with online revocation.
* `Certificate`s can now have null `StatusUpdateTime` and `NextStatusUpdateTime` columns (as that comes from the OCSP responses)

A signature with `Unknown` certificates will be considered to be in its grace period. See the spec linked in https://github.com/NuGet/Engineering/issues/789 for more information.

joelverhagen:
Could you provide the resulting SQL from executing this new migration on top of the existing migration (`Update-Database -Script`)?

loic-sharma:
Added SQL migration to original description.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -122,7 +131,138 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
...
+                .HasKey(p => p.PackageKey);
```
joelverhagen: Somehow this one should not be database-mastered. The value comes from the Gallery DB.

loic-sharma:
Updated schema to support full re-validations off a package's signature and its certificates.

src/NuGet.Services.Validation/Entities/Certificate.cs
loic-sharma: Done

src/NuGet.Services.Validation/Entities/ValidatorState.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// The state of an <see cref="IValidator"/>'s validation of a package. This should be used
+    /// by each <see cref="IValidator"/> to keep track of its own state.
+    /// </summary>
+    public class ValidatorState
```
joelverhagen: We should call this `ValidationStatus`. The validator itself should be stateless. The state is the validation.

Also, we can use the `ValidationId` as the key.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -122,7 +137,204 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                 .Property(pv => pv.RowVersion)
                 .IsRowVersion();
 
+            modelBuilder.Entity<ValidatorState>()
+                .HasKey(s => s.Key);
+
+            modelBuilder.Entity<ValidatorState>()
+                .Property(s => s.Key)
+                .IsRequired();
```
joelverhagen: This property is required by default, right?

src/NuGet.Services.Validation/Migrations/201709212249261_AddPackageSigningSchema.cs
```diff
@@ -0,0 +1,119 @@
...
+                .PrimaryKey(t => t.Key)
+                .Index(t => t.Thumbprint, unique: true, name: "IX_Certificates_Thumbprint");
+            
+            CreateTable(
+                "signature.CertificateValidations",
+                c => new
+                    {
+                        Key = c.Long(nullable: false, identity: true),
+                        CertificateKey = c.Long(nullable: false),
+                        ValidationId = c.Guid(nullable: false),
```
joelverhagen: This can be a foreign key right?

src/NuGet.Services.Validation/Entities/ValidationContext.cs
loic-sharma: Yes, the key is implicitly required already. I believe I saw this pattern used somewhere in the Gallery's context, but I can remove this line if you feel it's unnecessary.

src/NuGet.Services.Validation/Migrations/201709212249261_AddPackageSigningSchema.cs
loic-sharma: `ValidationId` cannot be a foreign key as the Revalidator job will also create its own `ValidationId`s, but will NOT be creating `ValidatorStatus` records. See the spec for more information :)

src/NuGet.Services.Validation/Entities/ValidationContext.cs
joelverhagen: I think it's unnecessary.

src/NuGet.Services.Validation/Migrations/201709212249261_AddPackageSigningSchema.cs
joelverhagen: Sounds good.

src/NuGet.Services.Validation/Entities/ValidatorState.cs
joelverhagen: `Validations` may even be better. Your call. I am bad at names.

src/NuGet.Services.Validation/Entities/ValidatorStatus.cs
```diff
@@ -6,28 +6,23 @@
...
+        /// The unique identifier for this validation. The Validation Orchestrator generates a unique
+        /// validation ID for each <see cref="IValidator"/> it runs on a single package.
         /// </summary>
-        public long Key { get; set; }
+        public Guid ValidationId { get; set; }
 
         /// <summary>
         /// The package key in the NuGet gallery database.
         /// </summary>
         public long PackageKey { get; set; }
```
joelverhagen: `PackageKey` is an `int` today.

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// NULL if online verification have never been completed for this certificate.
+        /// </summary>
+        public DateTime? StatusUpdateTime { get; set; }
+
+        /// <summary>
+        /// The time at or before which newer information will be available about the certificate's status,
+        /// according to the Certificate Authority. NULL if online verification have never been completed for
+        /// this certificate.
+        /// </summary>
+        public DateTime? NextStatusUpdateTime { get; set; }
```
chenriksson: This sounds like a second StatusUpdateTime, until you read the comment. I wonder if there's a better name... maybe CertificateAuthorityUpdateTime?

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// The last time this certificate's metadata was updated using online verification. NULL if online
+        /// verification have never been completed for this certificate.
```
chenriksson: nit: verification *has*

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// </summary>
+        public DateTime? LastVerificationTime { get; set; }
+
+        /// <summary>
+        /// The time at which the certificate was revoked. NULL if the certificate has not been revoked.
+        /// </summary>
+        public DateTime? RevocationTime { get; set; }
+
+        /// <summary>
+        /// The number of times online revocation checking failed without a "Good" or "Revoked" response.
```
chenriksson: It fails with a "Good" response?

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// The SHA1 thumbprint that uniquely identifies this certificate. This is a string with exactly 40 characters.
+        /// </summary>
+        public string Thumbprint { get; set;}
```
chenriksson: Since the length is known, should we use StringLength attribute here?

src/NuGet.Services.Validation/Entities/CertificateValidation.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// The validation performed for a given certificate.
+    /// </summary>
+    public class CertificateValidation
+    {
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
```
chenriksson: Is long necessary here? Do old cert validations need to be retained, or could we periodically delete them?

src/NuGet.Services.Validation/Entities/CertificateValidation.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// The unique identifier that represents this round of validation.
+        /// </summary>
+        public Guid ValidationId { get; set; }
+
+        /// <summary>
+        /// The determined <see cref="Certificate"/> Status after the validation finished. NULL
+        /// if the validation has not completed yet.
+        /// </summary>
+        public CertificateStatus? Status { get; set; }
```
chenriksson: Why null if validation not completed, as opposed to CertificateStatus.Unavailable? What's the difference?

src/NuGet.Services.Validation/Entities/PackageSignature.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        Invalid = 0,
+
+        /// <summary>
+        /// A signature is considered to be in its "grace period" if one of its certificate's status is unknown OR
+        /// its last known status is older than the signature itself. In other words, a signature is in
+        /// its grace period if one or more the signature's <see cref="Certificate"/>s' "StatusUpdateTime" is before
+        /// <see cref="PackageSignature"/>'s "SignedAt".
+        /// </summary>
+        InGracePeriod = 1,
```
chenriksson: Is the SignatureStatus package aggregate Valid if PackageSignatureStatuses for all signatures are either Valid or in GracePeriod?

SignatureStatus and PackageSignatureStatus naming is confusing, and their purpose is so similar... could we just combine them? On a related note, why not use InGracePeriod for package.SignatureStatus as well?

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -13,10 +13,25 @@ namespace NuGet.Services.Validation
     [DbConfigurationType(typeof(EntitiesConfiguration))]
     public class ValidationEntitiesContext : DbContext
     {
+        private const string SignatureSchema = "signature";
```
chenriksson: We use the default schema name everywhere else... why change?

src/NuGet.Services.Validation/Entities/ValidatorState.cs
chenriksson: Not ValidationStatus, which conflicts with the status in Contracts.

I like Validations... though, does the Orchestrator also track these in some other table?

src/NuGet.Services.Validation/Entities/ValidatorState.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// The database-mastered identifier for this validator state.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// The package key in the NuGet gallery database.
+        /// </summary>
+        public long PackageKey { get; set; }
```
chenriksson: Shouldn't this be int to match Gallery schema?

src/NuGet.Services.Validation/Entities/ValidatorStatus.cs
loic-sharma: Oops, my bad. Good catch!

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// <summary>
+    /// A X.509 Certificate used by one or more <see cref="PackageSignature" />s for one or more <see cref="Package"/>s.
+    /// </summary>
+    public class Certificate
+    {
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
```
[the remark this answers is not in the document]
loic-sharma: I would expect that there will be more packages than certificates. However, there are over a million packages today, I think it's reasonable to expect that we will hit 2 billion packages at some point. `PackageKey` should be a long (but that's out of the scope of this change).

src/NuGet.Services.Validation/Entities/Certificate.cs
loic-sharma: That's set on line 184 of `ValidationContext.cs` (in the `RegisterPackageSigningEntities` method). Would you like me to be explicit on the model as well?

src/NuGet.Services.Validation/Entities/Certificate.cs
loic-sharma: Are you suggesting that I rename `NextStatusUpdateTime` to `CertificateAuthorityUpdateTime` but keep `StatusUpdateTime` as is? Both of these fields come from OCSP responders, so if possible, I'd like to keep their names similar. Maybe `CertificateAuthorityUpdateTime` and `NextCertificateAuthorityUpdateTime`? This seems quite verbose though...

src/NuGet.Services.Validation/Entities/CertificateValidation.cs
loic-sharma: The `Status` is `null` is when the certificate validation hasn't finished yet. `Unknown` is a possible outcome for the validation (for things like OCSP responder not responding properly, or network errors). The spec has more detailed information if you're interested.

src/NuGet.Services.Validation/Entities/ValidatorState.cs
loic-sharma: Fixed! :)

src/NuGet.Services.Validation/Entities/Certificate.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// The <see cref="Certificate" /> has not been revoked.
+        /// </summary>
+        Good = 1,
+
+        /// <summary>
+        /// The <see cref="Certificate" /> has been revoked.
+        /// </summary>
+        Revoked = 2,
+    }
```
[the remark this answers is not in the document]
loic-sharma: `Services.Contracts` is shared between projects that do not depend on (and do not know anything about) the `ValidationContext`. `CertificateStatus` and the other enums will only be used by projects that use the `ValidationContext`. Hence, I think it's okay to leave these enums in their current location, but if you feel strongly about this I can move them over.

src/NuGet.Services.Validation/Entities/Certificate.cs
loic-sharma: Good catch, thanks! :)

src/NuGet.Services.Validation/Entities/ValidationContext.cs
loic-sharma: We're reusing the Validation Orchestrator's database for the package signature validations. I have some tables like `Packages` that are "broadly" named but are very specific to package signatures. This schema name change is essentially a separate namespace for the package signature's tables.

src/NuGet.Services.Validation/Entities/Certificate.cs
loic-sharma: I'll rephrase this to make it a little clearer :)

src/NuGet.Services.Validation/Entities/PackageSignature.cs
loic-sharma: Yup, `Package`'s `SignatureStatus` is an aggregate of `PackageSignature`s' `Status` field. The `SignatureStatus` is valid if-and-only-if all `PackageSignature`s' `Status` are `Valid` or `InGracePeriod`.

I'll rename `PackageSignatureStatus` to `PackageSigningStatus` and `SignatureStatus` to `SigningStatus` to make the distinction between `SignatureStatus` and `PackageSigningStatus` a little clearer. Let me know if you think that's still confusing.

We could add `InGracePeriod` to `Package`'s `SigningStatus`, but nothing would use this value today (we only use `Package`'s `SigningStatus` on the initial push validation, whereas we reuse `PacakgeSignature`'s `Status` constantly to do revalidations).

src/NuGet.Services.Validation/Entities/CertificateValidation.cs
loic-sharma: Talked offline with Christy, Andrei, and Joel. Old certification validations do not need to be retained, and we may periodically delete them in the future. If in the future we add support for periodically dropping validations, this will done along side the orchestrator.

src/NuGet.Services.Validation/Entities/Package.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// All of the package's signatures are valid.
+        /// </summary>
+        Valid = 2,
+    }
+
+    /// <summary>
+    /// The metadata for a package's signature. A package may be signed with one or more <see cref="PackageSignature"/>s
+    /// using one or more <see cref="Certificate"/>.
+    /// </summary>
+    public class Package
```
agr: `Package` name does not sound right here. The table would not contain packages. Signature validation requests maybe?

agr: Or package signature set...

src/NuGet.Services.Validation/Migrations/201709212249261_AddPackageSigningSchema.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            CreateTable(
+                "dbo.ValidatorStatus",
+                c => new
+                    {
+                        ValidationId = c.Guid(nullable: false),
+                        PackageKey = c.Int(nullable: false),
+                        State = c.Int(nullable: false),
+                    })
+                .PrimaryKey(t => t.ValidationId)
```
agr: That will cause terrible table fragmentation unless sequential GUIDs are used.

src/NuGet.Services.Validation/Entities/Package.cs
loic-sharma: I disagree, this table *does* contain package's metadata. Each package that is validated (even if it has no signature) will create a record in the `Packages` table, and there will only ever be one entry per package.

src/NuGet.Services.Validation/Migrations/201709212249261_AddPackageSigningSchema.cs
agr: Nevermind, it seems it is going to reference the `PackageValidation.Key` which is generated using sequential GUIDs, so fragmentation should not be a concern.

src/NuGet.Services.Validation/Entities/Package.cs
agr: We already have an entity named "Package" in Gallery, here it gets overridden with different meaning. This is confusing.
It contains secondary information about some aspect of the package, so "Package" seems like overly generic name.

src/NuGet.Services.Validation/Entities/Certificate.cs
chenriksson: Missed that - was just looking at examples from the Gallery. I don't think we're consistent one way or another.

src/NuGet.Services.Validation/Entities/Certificate.cs
chenriksson: I was, but if the naming is based on OCSP fields then it's probably best to keep as-is.

src/NuGet.Services.Validation/Entities/Package.cs
chenriksson: Maybe PackageReference is appropriate?

src/NuGet.Services.Validation/Entities/ValidationContext.cs
chenriksson: I would argue that Packages should be in the default schema, as other validators may need it too. Although in that case, maybe the orchestrator should control the population of this for validators. Just concerned about reuse here.

src/NuGet.Services.Validation/Entities/Package.cs
loic-sharma: Naming is hard! I will change the `Packages` table to `PackageSigningStates` :)

src/NuGet.Services.Validation/Entities/ValidationContext.cs
loic-sharma: Made the `Packages` table's name was made more specific as it isn't intend to be reused by other validators.

PR #62, opened 2017-09-26
Add StorageQueue wrapper for queues

scottbommarito:
A wrapper for interacting with queues. Currently, Azure storage queues are implemented--additionally work will need to be done to support other types of queues.


src/NuGet.Services.Storage/JsonStorageQueueMessageSerializer.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            return JsonConvert.SerializeObject(contents, _settings);
+        }
+
+        public override T Deserialize(string contents)
+        {
+            return JsonConvert.DeserializeObject<T>(contents, _settings);
+        }
+
+        public JsonStorageQueueMessageSerializer(JsonSerializerSettings settings)
```
loic-sharma: >JsonStorageQueueMessageSerializer [](start = 15, length = 33)

Nit: move constructor to top of class

src/NuGet.Services.Storage/JsonStorageQueueMessageSerializer.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            _settings = settings;
```
loic-sharma: >settings [](start = 24, length = 8)

Add null check?

src/NuGet.Services.Storage/IStorageQueueMessage.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Storage
+{
+    public interface IStorageQueueMessage
```
loic-sharma: >IStorageQueueMessage [](start = 21, length = 20)

There doesn't seem to be a need for interfaces on the message. Consider removing the interfaces and keeping just the concrete `StorageQueueMessage`/`StorageQueueMessage<T>` types.

src/NuGet.Services.Storage/JsonStorageQueueMessageSerializer.cs
scottbommarito: `JsonConvert` uses the default `JsonSerializerSettings` if it's null so no need to check.

src/NuGet.Services.Storage/IStorageQueue.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <param name="contents">The contents of a message to be added to the queue.</param>
+        /// <param name="token">A token to cancel the task with.</param>
+        Task Add(T contents, CancellationToken token);
+
+        /// <summary>
+        /// Adds a message to the queue.
+        /// </summary>
+        /// <param name="message">A message to add to the queue.</param>
+        /// <param name="token">A token to cancel the task with.</param>
+        Task Add(IStorageQueueMessage<T> message, CancellationToken token);
```
skofman1: if u dont use this, delete this

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,68 @@
...
+        private CloudQueue _queue;
+
+        private TimeSpan _visibilityTimeout = TimeSpan.FromMinutes(5);
+
+        public AzureStorageQueue(CloudStorageAccount account, string queueName, IStorageQueueMessageSerializer<T> serializer)
+            : base(serializer)
+        {
+            _queue = account.CreateCloudQueueClient().GetQueueReference(queueName);
+            _queue.CreateIfNotExists();
+        }
```
skofman1: the best practice for ctors is that they shouldn't do any network communications/long running actions. You should move the queue initialization to OnAdd/OnRemove if the form of if(notInitialized) {DoInit()}

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public override Task OnAdd(IStorageQueueMessage message, CancellationToken token)
+        {
+            CloudQueueMessage cloudMessage;
+
+            if (message is AzureStorageQueueMessage)
+            {
+                cloudMessage = ((AzureStorageQueueMessage)message).Message;
```
skofman1: Is queueing an CloudQueueMessage something you use now?
If not, I recommend removing this code, until the time we actually need it.
If you do use it, consider adding to this class an explicit method for CloudQueueMessage adding. The user of the class will know he is using Azure, and the abstraction is not needed.

src/NuGet.Services.Storage/StorageQueue.cs
```diff
@@ -0,0 +1,66 @@
...
+    public abstract class StorageQueue<T> : IStorageQueue<T>
+    {
+        private IStorageQueueMessageSerializer<T> _serializer;
+
+        /// <summary>
+        /// Adds a <see cref="IStorageQueueMessage"/> to the queue.
+        /// </summary>
+        /// <param name="message">The message to add.</param>
+        /// <param name="token">A token to cancel the task with.</param>
+        public abstract Task OnAdd(IStorageQueueMessage message, CancellationToken token);
```
skofman1: Should a user of this class call OnAdd directly? If not all the OnX methods should be protected

src/NuGet.Services.Storage/AzureStorageQueue.cs
loic-sharma: Or `Lazy<T>` for thread safety.

src/NuGet.Services.Storage/AzureStorageQueue.cs
scottbommarito: This code is entirely used.

I'd rather not add a `CloudQueueMessage` because it would be unused--all uses of this class are through the abstraction.

src/NuGet.Services.Storage/StorageQueue.cs
scottbommarito: good catch!

src/NuGet.Services.Storage/IStorageQueue.cs
scottbommarito: done

src/NuGet.Services.Storage/AzureStorageQueue.cs
scottbommarito: used `Lazy<T>`

I will note that I'm pretty sure I've seen this (ctors that do network calls) many places elsewhere in our code base.

src/NuGet.Services.Storage/IStorageQueue.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Storage
+{
+    /// <summary>
+    /// Represents a queue to add <see cref="IStorageQueueMessage{T}"/>s to and receive <see cref="IStorageQueueMessage{T}"/>s from.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public interface IStorageQueue<T>
```
joelverhagen: I don't think queues should be strongly typed. They should give back a message with `string` or `byte[]` that the caller deserializes.

There are too many deserialization concerns for it to be lumped in with the enqueueing/dequeueing/leasing/etc concerns that a queue already has.

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace NuGet.Services.Storage
+{
+    public class AzureStorageQueue<T> : StorageQueue<T>
+    {
+        private Lazy<CloudQueue> _queue;
+
+        private TimeSpan _visibilityTimeout = TimeSpan.FromMinutes(5);
+
+        public AzureStorageQueue(CloudStorageAccount account, string queueName, IStorageQueueMessageSerializer<T> serializer)
```
joelverhagen: The wrapper around `CloudQueue` should be much thinner and ideally has the same shape as `ICloudQueue`. This logic is fine, but it should depend on an `ICloudQueue` for caching, creating of the queue, deserialization, etc.

src/NuGet.Services.Storage/IStorageQueue.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Adds a message containing <paramref name="contents"/> to the queue.
+        /// </summary>
+        /// <param name="contents">The contents of a message to be added to the queue.</param>
+        /// <param name="token">A token to cancel the task with.</param>
+        Task Add(T contents, CancellationToken token);
```
joelverhagen: `AddAsync`

src/NuGet.Services.Storage/IStorageQueue.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Removes a message from the queue.
+        /// </summary>
+        /// <param name="message">The message to be removed from the queue.</param>
+        /// <param name="token">A token to cancel the task with.</param>
+        /// <remarks>
+        /// <paramref name="message"/> MUST have been previously returned by <see cref="GetNextAsync(CancellationToken)"/>.
+        /// </remarks>
+        Task Remove(IStorageQueueMessage<T> message, CancellationToken token);
```
joelverhagen: `RemoveAsync`

joelverhagen:
In general, I think the abstractions are in the wrong place here. Wrappers around `WindowsAzure.Storage` types should be razor thin if possible and deserialization should be outside of the queue integration.

src/NuGet.Services.Storage/JsonMessageSerializer.cs
```diff
@@ -6,24 +6,23 @@
...
-        public JsonStorageQueueMessageSerializer(JsonSerializerSettings settings)
+        public JsonMessageSerializer(JsonSerializerSettings settings)
         {
             _settings = settings;
         }
 
-        public override string Serialize(T contents)
+        public string Serialize(T contents)
         {
             return JsonConvert.SerializeObject(contents, _settings);
```
joelverhagen: We should have a schema version + message type in here. If the schema version or message type is not the expected value, throw. This means when we roll a new message format, old queue processors will reject new queue messages and new queue processors will migrate old messages to the new model.

scottbommarito: Added `TypedMessageSerializer<T>`. The user can pass `JsonMessageSerializer<T>` to it in order to get the functionality you've described.

joelverhagen: This should be the only implementation. People shouldn't be allowed to queue untyped/unversioned messages.

scottbommarito: i will make `StorageQueue<T>` use this implementation by default

joelverhagen: Why do we even have `JsonMessageSerializer<T>`?

scottbommarito: For people who want to use a different method of serialization. I suppose it's not likely that will happen but the cost of leaving it extensible and removing the extensibility are roughly the same.

scottbommarito: Also, `TypedMessageSerializer` uses `JsonMessageSerializer` internally

chenriksson:
Is there a plan to add NuGet.Services.Storage.Tests?

scottbommarito:
@chenriksson We should, but it should be backlogged.

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Queue;
+
+namespace NuGet.Services.Storage
+{
+    public class AzureStorageQueue : IStorageQueue
+    {
+        private Lazy<Task<CloudQueue>> _queueTask;
+
+        private TimeSpan _visibilityTimeout = TimeSpan.FromMinutes(5);
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public AzureStorageQueue(CloudStorageAccount account, string queueName)
+        {
```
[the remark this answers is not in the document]
scottbommarito: done

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -0,0 +1,65 @@
...
+            });
+        }
+
+        public async Task AddAsync(string contents, CancellationToken token)
+        {
+            var azureMessage = new AzureStorageQueueMessage(contents);
+            await (await _queueTask.Value).AddMessageAsync(azureMessage.Message, token);
+        }
+
+        public async Task<IStorageQueueMessage> GetNextAsync(CancellationToken token)
```
loic-sharma: >GetNextAsync [](start = 48, length = 12)

You should probably put a comment somewhere saying that the message will be visible after 5 minutes if it isn't removed with RemoveAsync.

src/NuGet.Services.Storage/TypedMessageSerializer.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    /// <remarks>
+    /// For this to work properly, the version property of the constructor must be incremented as changes to <typeparamref name="T"/> occur.
+    /// </remarks>
+    public class TypedMessageSerializer<T> : IMessageSerializer<T>
+    {
+        private IMessageSerializer<T> _contentsSerializer;
+        private IMessageSerializer<TypedMessage> _messageSerializer;
+
+        private string _type = typeof(T).FullName;
```
loic-sharma: >private [](start = 8, length = 7)

This can be readonly.

src/NuGet.Services.Storage/AzureStorageQueue.cs
scottbommarito: yup, done

src/NuGet.Services.Storage/TypedMessageSerializer.cs
scottbommarito: yup

src/NuGet.Services.Storage/TypedMessageSerializer.cs
```diff
@@ -0,0 +1,56 @@
...
+        private int _version;
+
+        public TypedMessageSerializer(IMessageSerializer<T> contentsSerializer, IMessageSerializer<TypedMessage> messageSerializer, int version)
+        {
+            _contentsSerializer = contentsSerializer ?? throw new ArgumentNullException(nameof(contentsSerializer));
+            _messageSerializer = messageSerializer ?? throw new ArgumentNullException(nameof(messageSerializer));
+            _version = version;
+        }
+
+        public string Serialize(T contents)
```
loic-sharma: >Serialize [](start = 22, length = 9)

Considering `Serialize` and `Deserialize` are public methods, I would consider adding null checks.

src/NuGet.Services.Storage/StorageQueue.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public Task RemoveAsync(IStorageQueueMessage<T> message, CancellationToken token)
+        {
+            return _queue.RemoveAsync(SerializeMessage(message), token);
+        }
+
+        private IStorageQueueMessage SerializeMessage(IStorageQueueMessage<T> message)
+        {
+            if (message is DeserializedStorageQueueMessage<T>)
```
loic-sharma: >message is DeserializedStorageQueueMessage<T> [](start = 16, length = 45)

Nit: you can do `if (message is DeserializedStorageQueueMessage<T> deserializedMessage) { return deserializedMessage.Message }`

src/NuGet.Services.Storage/TypedMessageSerializer.cs
scottbommarito: what if you want to push a null message?

it sounds dumb but I don't think there's a reason to implicitly reject it when the serializers and deserializers handle it fine

src/NuGet.Services.Storage/StorageQueue.cs
scottbommarito: nice

src/NuGet.Services.Storage/AzureStorageQueue.cs
scottbommarito: (it's on `_visibilityTimeout`)

src/NuGet.Services.Storage/IStorageQueueMessage.cs
scottbommarito: fixed

PR #63, opened 2017-09-26
For storage APIs, use interfaces whenever possible

scottbommarito:
For whatever reason, we were declaring storage interfaces (`IStorage`, `IStorageFactory`) but not using them everywhere and instead using the base implementation (`Storage`, `StorageFactory`). This PR fixes all instances of this.

It also adds `Exists` to `IStorage`, because it seemed ridiculous not to have it when every subclass implements it.

### DIFF
--- a/src/NuGet.Services.Storage/IStorage.cs
+++ b/src/NuGet.Services.Storage/IStorage.cs
@@ -15,6 +15,7 @@ namespace NuGet.Services.Storage
         Task<string> LoadString(Uri resourceUri, CancellationToken cancellationToken);
         Uri BaseAddress { get; }
         Uri ResolveUri(string relativeUri);
+        bool Exists(string fileName);
         Task<IEnumerable<StorageListItem>> List(CancellationToken cancellationToken);
     }
 }
--- a/src/NuGet.Services.Cursor/DurableCursor.cs
+++ b/src/NuGet.Services.Cursor/DurableCursor.cs
@@ -11,10 +11,10 @@ namespace NuGet.Services.Cursor
     public class DurableCursor : ReadWriteCursor<DateTimeOffset>
     {
         Uri _address;
-        Storage.Storage _storage;
+        IStorage _storage;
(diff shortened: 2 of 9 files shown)


src/NuGet.Services.Storage/IStorage.cs
```diff
@@ -15,6 +15,7 @@ public interface IStorage
         Task<string> LoadString(Uri resourceUri, CancellationToken cancellationToken);
         Uri BaseAddress { get; }
         Uri ResolveUri(string relativeUri);
+        bool Exists(string fileName);
```
loic-sharma: >Exists [](start = 13, length = 6)

Nit: considering changing `fileName` to `relativeUri` to make its purpose a little more clear.

skofman1:
@scottbommarito , this change affects all V3 jobs. What's the benefit in it?

scottbommarito:
@skofman1 
The benefit is code style, testability, and better polymorphism. In general, interfaces are easier to work with than abstract classes.

I was about to update all of the V3 jobs to use `NuGet.Services.Storage` (as required by the V3 monitoring jobs changes), which is slightly out of sync with `NuGet.Services.Metadata.Catalog` as is. I thought I might as well make this change now before I swap libraries because it will be less work in the long run. TBH I'm not sure what even uses `NuGet.Services.Storage` right now because none of the V3 jobs currently use it.

scottbommarito:
@skofman1 Also, because the namespace is called `NuGet.Services.Storage`, unless we use `IStorage` everywhere, every usage of `Storage` must be updated to `Storage.Storage`, which I think is a more useless change than this.

skofman1:
@scottbommarito , please don't use NuGet.Services.Storage for V3 jobs right now. Like you said, the code is out of sync, and changing libs will require a significant testing investment.

scottbommarito:
@skofman1 Agreed. I think this is still a valuable PR, however, but for the backlog.

PR #65, opened 2017-10-04
[Package Signing] Miscellaneous improvements to ValidationContext

loic-sharma:
Updated the ValidationContext:

* Added a `IValidationEntitiesContext` to allow us to mock the validation context in unit tests.
* Added the property `ValidatorName` to `ValidatorStatus`. This will be used in the revalidation flow to check if the validation request's package has already been validated by the current `IValidator`.
* Added a `IsFinished` helper to `CertificateValidation` as `CertificateValidation.IsFinished()` is clearer than `!CertificateValidation.Status.HasValue`


src/NuGet.Services.Validation/Entities/IValidationContext.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Data.Entity;
+using System.Data.Entity.Infrastructure.Annotations;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Validation
+{
+    public interface IValidationEntitiesContext
+    {
+        IDbSet<PackageValidationSet> PackageValidationSets { get; set; }
```
joelverhagen: Do we need `set;` here?

src/NuGet.Services.Validation/Entities/IValidationContext.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Data.Entity;
+using System.Data.Entity.Infrastructure.Annotations;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Validation
+{
+    public interface IValidationEntitiesContext
```
joelverhagen: Beautiful. 🥇

src/NuGet.Services.Validation/Entities/IValidationContext.cs
loic-sharma: Removing `set;` builds and passes (my unpublished) unit tests. I also tested removing the `set;` from the Gallery's entity interface, that passed manual testing. So... it seems fine? :)

PR #68, opened 2017-10-09
Add GetTrafficManagerPropertiesAsync to AzureManagementAPIWrapper

scottbommarito:
Added `GetTrafficManagerPropertiesAsync` to `AzureManagementAPIWrapper` so that we can fetch Traffic Manager properties from Azure.


src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -26,7 +26,7 @@ public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2
             
             VaultName = vaultName;
             ClientId = clientId;
-            Certificate = certificate ?? throw new ArgumentNullException(nameof(certificate));
+            Certificate = certificate/* ?? throw new ArgumentNullException(nameof(certificate))*/;
```
skofman1: undo this.. it was only for test

scottbommarito: fixed, woops

PR #70, opened 2017-10-11
Added generic Service Bus serializer

loic-sharma:
Added `BrokedMessageSerializer`, a JSON serializer that enforces versioned schemas. Updated the `PackageValidationMessageData` serializer to use this new serializer. This will be used by `PackageSigningValidator` and `PackageCertificatesValidator` to queue messages to their downstream validation jobs.


src/NuGet.Services.ServiceBus/SchemaVersionAttribute.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace NuGet.Services.ServiceBus
+{
+    /// <summary>
+    /// The version of the sche
```
joelverhagen: `sche`

tests/NuGet.Services.Validation.Tests/ServiceBusMessageSerializerTests.cs
```diff
@@ -16,7 +16,7 @@ public class ServiceBusMessageSerializerTests
         private const string PackageId = "NuGet.Versioning";
         private const string PackageVersion = "4.3.0";
         private static readonly Guid ValidationTrackingId = new Guid("14b4c1b8-40e2-4d60-9db7-4b7195e807f5");
-        private const string PackageValidationMessageDataType = "PackageValidationMessageData";
+        private const string PackageValidationMessageDataType = "PackageValidationMessageData1";
```
scottbommarito: `nameof(PackageValidationMessageData1`)`?

src/NuGet.Services.ServiceBus/BrokedMessageSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+
+namespace NuGet.Services.ServiceBus
+{
+    /// <summary>
+    /// Serializes objects into Service Bus <see cref="IBrokeredMessage"/>. This serializer will
+    /// throw <see cref="FormatException"/> if the message does not contain a message with the expected
+    /// type and schema version.
+    /// </summary>
+    /// <typeparam name="TMessage">A type decorated with a <see cref="SchemaVersionAttribute"/>.</typeparam>
+    public class BrokedMessageSerializer<TMessage>
```
joelverhagen: Hopefully not broked 🤞 
`Brokered`

src/NuGet.Services.ServiceBus/BrokedMessageSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var schemaAttribute = typeof(SchemaVersionAttribute);
+            var attributes = typeof(TMessage).GetCustomAttributes(schemaAttribute, inherit: false);
+
+            if (attributes == null || attributes.Length != 1)
+            {
+                throw new InvalidOperationException($"{typeof(TMessage)} must have exactly one {nameof(SchemaVersionAttribute)}");
+            }
+
+            MessageType = typeof(TMessage).Name;
```
joelverhagen: This misses the point of `MessageType`. This is meant to be constant for all versions so that caller can detect that he is listening on the wrong queue (or is using a shared queue). If it is different for each schema version of the same type, it doesn't make sense to have two properties since one implies the other.

`MessageType` needs to be a parameter or if we must use reflection then an attribute (new one or expand `SchemaVersionAttribute`).

tests/NuGet.Services.Validation.Tests/ServiceBusMessageSerializerTests.cs
joelverhagen: This value should never change. If it changes, in-flight messages will become unreadable.

src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
```
joelverhagen: Copyright

src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        private static readonly string MessageType;
+        private static readonly int SchemaVersion;
+
+        static BrokeredMessageSerializer()
+        {
+            var schemaAttribute = typeof(SchemaVersionAttribute);
+            var attributes = typeof(TMessage).GetCustomAttributes(schemaAttribute, inherit: false);
+
+            if (attributes == null || attributes.Length != 1)
```
joelverhagen: How could this array be null?

src/NuGet.Services.Validation/ServiceBusMessageSerializer.cs
```diff
@@ -2,99 +2,35 @@
...
-            };
-
-            var brokeredMessage = Serialize(body, PackageValidationType, SchemaVersion1);
-
-            return brokeredMessage;
+            });
         }
 
         public PackageValidationMessageData DeserializePackageValidationMessageData(IBrokeredMessage message)
         {
```
joelverhagen: How do you expect this code to work when there are two versions of `PackageValidationMessageData`? Two different `_serializers` called in succession with try/catch? This is a bit nasty since the version check and the serialization are done in the same method (`_serializer.Deserialize`).

tests/NuGet.Services.Validation.Tests/ServiceBusMessageSerializerTests.cs
loic-sharma: I think it's better to not use `nameof` here simply because this *should* never change even if the type changes.

src/NuGet.Services.ServiceBus/BrokedMessageSerializer.cs
loic-sharma: Good point. I'll add an attribute to the name.

tests/NuGet.Services.Validation.Tests/ServiceBusMessageSerializerTests.cs
loic-sharma: Correct. I will change this back such that we used a set name instead of the internal schematized type.

src/NuGet.Services.Validation/ServiceBusMessageSerializer.cs
loic-sharma: Yes, this is an issue in the existing code as well. This change isn't intended to fix the versioning problem, but rather, to make service bus serialization/deserialization generic enough to unblock package signing validation. In the future, we could add a method to try deserialize - or we could leverage something that handles the versioning problem properly, like Bond.

joelverhagen: Previously, the schema version property was an implementation detail of `ServiceBusMessageSerializer`. Now it's not. It seems odd to give `BrokedMessageSerializer` a shape that will need to change when we have > 1 version, but I agree it does not need to be solved now.

src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
loic-sharma: Ah, I followed an example I found online. The MSDN docs says the array will have a length of 0 if the attribute doesn't exist, I will remove the null check.

src/NuGet.Services.Validation/ServiceBusMessageSerializer.cs
```diff
@@ -2,99 +2,35 @@
...
-
-            return brokeredMessage;
-        }
-
-        private static T Deserialize<T>(string json)
-        {
-            return JsonConvert.DeserializeObject<T>(json);
-        }
-
+        [Schema(Name = "PackageValidationMessageData", Version = 1)]
```
joelverhagen: nit: constant the Name

tests/NuGet.Services.ServiceBus.Tests/BrokeredMessageSerializer.cs
```diff
@@ -0,0 +1,185 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Moq;
+using Xunit;
+
+namespace NuGet.Services.ServiceBus.Tests
+{
+    public class BrokeredMessageSerializer
```
joelverhagen: Call this class `Facts` or `Tests`

joelverhagen: And make the file name match

src/NuGet.Services.Validation/ServiceBusMessageSerializer.cs
```diff
@@ -2,99 +2,35 @@
...
 using System;
-using Newtonsoft.Json;
 using NuGet.Services.ServiceBus;
 
 namespace NuGet.Services.Validation
 {
     public class ServiceBusMessageSerializer : IServiceBusMessageSerializer
     {
-        private const string SchemaVersionKey = "SchemaVersion";
-        private const string TypeKey = "Type";
```
joelverhagen: nit: this is a breaking change since `Type` was renamed to `SchemaName`. I think it's fine since we've not gone to PROD with this but we need to be careful with this sort of change.

loic-sharma: That's true. This requires that both the Gallery and NuGet.Jobs repositories are linked against the latest version of ServerCommon. I'll make sure to update both.

@agr FYI

src/NuGet.Services.ServiceBus/SchemaAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// The attribute used to define a schema.
+    /// </summary>
+    public class SchemaAttribute : Attribute
+    {
+        /// <summary>
+        /// The name of a message's schema. This should NEVER change for a single message.
+        /// </summary>
+        public string Name { get; set; }
```
joelverhagen: Why `set;`? Does this interface really need mutability?

loic-sharma: It's to allow the `[Schema(Name = "X")]` syntax. You can't do it without the `set;`

joelverhagen: named parameters were designed for this sort of clarity, but it's not a big deal. Doesn't matter too much though since someone would really have to go out of their way to mutate an attribute.

PR #72, opened 2017-10-13
[Package Signing] Store a package signature's end certificate instead of the full chain

loic-sharma:
A single certificate may have multiple valid certificate chains (see the "Certificate chains and cross-certification" section [here](https://en.wikipedia.org/wiki/X.509)). Therefore, we should not store the full certificate chain when validating a certificate as this may change. Instead, the certificate's chain should be rebuilt every time the certificate is validated.

I also added the `Invalid` `CertificateStatus`. Unlike `Revoked`, an `Invalid` status invalidates ALL signatures that depends on this certificate.

These are breaking changes, however, we haven't run the "AddPackageSigninigSchema" migration anywhere.


src/NuGet.Services.Validation/Entities/PackageSignature.cs
```diff
@@ -22,8 +21,13 @@ public class PackageSignature
...
+        /// </summary>
+        public long CertificateKey { get; set; }
+
+        /// <summary>
         /// The time at which this signature was created. A signature is valid as long as it was signed
-        /// before its certificates were revoked and/or expired. This timestamp SHOULD come from a trusted
+        /// before its certificates were revoked and/or expired. This timestamp MUST come from a trusted
         /// timestamp authority.
         /// </summary>
         public DateTime SignedAt { get; set; }
```
dtivel: Consider using `DateTimeOffset` instead wherever you use `DateTime`.

loic-sharma: Talked to Damon offline. Using `DateTimeOffset` properties changes the SQL Server column type from `datetime2` to `datetimeoffset`. This `datetimeoffset` type [stores both the UTC timestamp and the client's offset from the UTC timestamp](https://stackoverflow.com/questions/4715620/storing-datetime-utc-vs-storing-datetimeoffset). Considering that all other entities use `DateTime`, we decided it would be better to stick to `DateTime` and then enforce that all code uses UTC timestamp.

PR #73, opened 2017-10-17
Introduced IPackageSignaturesValidationEnqueuer implementation.

xavierdecoster:
Related to https://github.com/NuGet/Engineering/issues/785

* separate message type for enqueueing signatures validation (even though their schema looks the same *currently*)
* separate interface and implementation for enqueuer

### DIFF
--- a/src/NuGet.Services.Validation/IPackageSignaturesValidationEnqueuer.cs
+++ b/src/NuGet.Services.Validation/IPackageSignaturesValidationEnqueuer.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Validation
+{
+    public interface IPackageSignaturesValidationEnqueuer
+    {
+        /// <summary>
+        /// Enqueues signatures validation of the specified package to start ASAP.
+        /// </summary>
+        /// <param name="message">Package information</param>
+        Task StartSignaturesValidationAsync(PackageSignaturesValidationMessageData message);
+
+        /// <summary>
+        /// Enqueues signatures validation of the specified package to start no sooner than specified time
+        /// </summary>
+        /// <param name="message">Package information</param>
+        /// <param name="postponeProcessingTill">The time till which validation processing should be postponed.</param>
+        Task StartSignaturesValidationAsync(PackageSignaturesValidationMessageData message, DateTimeOffset postponeProcessingTill);
+    }
+}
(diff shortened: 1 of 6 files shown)


loic-sharma:
Hey, I talked to @joelverhagen and we both think that `IPackageSignaturesValidationEnqueuer`, `PackageSignaturesValidationMessageData`, and `PackageSignaturesValidationMessageData1` should all be in the NuGet.Jobs repository as these types will not be shared across multiple repositories.

src/NuGet.Services.Validation/IPackageSignaturesValidationEnqueuer.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// </summary>
+        /// <param name="message">Package information</param>
+        Task StartSignaturesValidationAsync(PackageSignaturesValidationMessageData message);
+
+        /// <summary>
+        /// Enqueues signatures validation of the specified package to start no sooner than specified time
+        /// </summary>
+        /// <param name="message">Package information</param>
+        /// <param name="postponeProcessingTill">The time till which validation processing should be postponed.</param>
+        Task StartSignaturesValidationAsync(PackageSignaturesValidationMessageData message, DateTimeOffset postponeProcessingTill);
```
loic-sharma: Out of curiosity, what will `postponeProcessingTill` be used for?

src/NuGet.Services.Validation/PackageSignaturesValidationEnqueuer.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public PackageSignaturesValidationEnqueuer(ITopicClient topicClient, IServiceBusMessageSerializer serializer)
+        {
+            _topicClient = topicClient;
+            _serializer = serializer;
+        }
+
+        public async Task StartSignaturesValidationAsync(PackageSignaturesValidationMessageData message)
+        {
+            await StartSignaturesValidationAsync(message, DateTimeOffset.MinValue);
```
loic-sharma: Nit: no need for await here, you can just return the `Task` directly and make this method non-`async`

xavierdecoster:
Thanks for the review @loic-sharma and @joelverhagen! will close this one, process your feedback, and take it along in the upcoming job PR.

xavierdecoster:
@loic-sharma @joelverhagen did you notice there were extensions to this existing `ServiceBusMessageSerializer`?

loic-sharma:
The changes to `ServiceBusMessageSerializer` looked good! However, I would recommend that you move the methods into their own class as the new [`SubscriptionProcessor`](https://github.com/NuGet/ServerCommon/pull/74) requires a `IBrokeredMessageSerializer<TMessage>` (which in your case would be `IBrokeredMessageSerializer<PackageSignaturesValidationMessageData>`). While its possible to jam two of the same interface with a different generic parameter, I would advise against that to keep things clean :)

xavierdecoster:
@loic-sharma but wouldn't the `PackageSignaturesValidationMessageData` (and the schema type) still need to be shared between the orchestrator's `PackageSignatureVerifier` (sending the message), and the new extract-n-validate job (receiving the message)? Those types would still be best put into a common assembly.

loic-sharma:
@xavierdecoster Agreed! That being said, that common assembly should be in NuGet.Jobs as that is the only repository that will need to consume the assembly.

PR #74, opened 2017-10-17
[Service Bus] Add SubscriptionProcessor

loic-sharma:
Added a `SubscriptionProcessor` to the `NuGet.Services.ServiceBus` project. Its purpose is to facilitate processing a specific message type from a given Service Bus subscription. This is a generic version of @agr's [`Orchestrator`](https://github.com/NuGet/NuGet.Jobs/pull/226/files#diff-0bcef9abb72b34c9e7c76d551f9ae9cd), with more tests. The pattern to use this class should follow the Orchestrator’s code:

1. A job should call the `SubscriptionProcessor`’s `Start` method within its `Run` method
2. A job should “block” the `Run`’s thread
3. On shutdown, a job should:
  a. Call the `SubscriptionProcessor`’s `StartShutdownAsync` method
  b. Wait until the `SubscriptionProcessor`’s `NumberOfMessagesInProgress` goes to 0

This will be used by the Package Signing jobs to listen to Service Bus queues and process validation requests.


src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Constructs a new subscription processor.
+        /// </summary>
+        /// <param name="client">The client used to receive messages from the subscription.</param>
+        /// <param name="serializer">The serializer used to deserialize received messages.</param>
+        /// <param name="handler">The handler used to handle received messages.</param>
+        /// <param name="logger">The logger used to record debug information.</param>
+        public SubscriptionProcessor(
+            ISubscriptionClient client,
+            IBrokeredMessageSerializer<TMessage> serializer,
```
loic-sharma: Note: `TMessage` is the type of the message, **not its schema**. This means that this `IBrokedMessageSerializer`'s `TMessage` is NOT a schema type (ie: NOT [`PackageValidationMessageData1`](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Validation/ServiceBusMessageSerializer.cs#L36), but rather, [`PackageValidationMessageData`](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Validation/PackageValidationMessageData.cs))

src/NuGet.Services.ServiceBus/project.json
```diff
@@ -1,5 +1,6 @@
 ﻿{
   "dependencies": {
+    "Microsoft.Extensions.Logging": "1.0.0",
```
agr: We could use more recent version. 1.1.2?

loic-sharma:
🔔 Tweaked processing logic to allow handlers to requeue messages. Please review again :)

PR #82, opened 2017-11-08
[Package Signing] Added a Scoped Service Bus Message Handler

loic-sharma:
Often times, Service Bus message handlers will need to use some sort of Entity Framework context. This requires that the handlers' callbacks are executed within their own dependency injection scope. This change creates a new generic Service Bus message handler `ScopedMessageHandler` that, upon receiving a message, creates a new dependency injection scope.

This change also makes `CachingSecretReader` thread-safe, as otherwise, the `ScopedMessageHandler` would need a lock.

### DIFF
--- a/src/NuGet.Services.KeyVault/CachingSecretReader.cs
+++ b/src/NuGet.Services.KeyVault/CachingSecretReader.cs
@@ -25,7 +25,12 @@ namespace NuGet.Services.KeyVault
 
         public async Task<string> GetSecretAsync(string secretName)
         {
-            // If the cache contains the secret and it is not expired, returned the cached value.
+            if (string.IsNullOrEmpty(secretName))
+            {
+                throw new ArgumentException("Null or empty secret name", nameof(secretName));
+            }
+
+            // If the cache contains the secret and it is not expired, return the cached value.
             if (_cache.TryGetValue(secretName, out CachedSecret result))
             {
                 if (!IsSecretOutdated(result))
@@ -51,7 +56,7 @@ namespace NuGet.Services.KeyVault
         /// <summary>
         /// A cached secret.
         /// </summary>
-        private struct CachedSecret
+        private class CachedSecret
         {
             /// <summary>
             /// The value of the cached secret.
--- a/src/NuGet.Services.KeyVault/CachingSecretReader.cs
+++ b/src/NuGet.Services.KeyVault/CachingSecretReader.cs
@@ -17,7 +17,7 @@ namespace NuGet.Services.KeyVault
 
         public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec)
         {
-            _internalReader = secretReader ?? throw new ArgumentNullException(nameof(secretReader)); ;
+            _internalReader = secretReader ?? throw new ArgumentNullException(nameof(secretReader));
             _cache = new ConcurrentDictionary<string, CachedSecret>();
 
             _refreshInterval = TimeSpan.FromSeconds(refreshIntervalSec);
@@ -31,26 +31,22 @@ namespace NuGet.Services.KeyVault
             }
 
             // If the cache contains the secret and it is not expired, return the cached value.
-            if (_cache.TryGetValue(secretName, out CachedSecret result))
+            if (_cache.TryGetValue(secretName, out CachedSecret result)
+                && !IsSecretOutdated(result))
             {
-                if (!IsSecretOutdated(result))
-                {
-                    return result.Value;
-                }
+                return result.Value;
             }
 
             // The cache does not contain a fresh copy of the secret. Fetch and cache the secret.
-            result.Value = await _internalReader.GetSecretAsync(secretName);
-            result.CacheTimeUtc = DateTime.UtcNow;
+            var updatedValue = new CachedSecret(await _internalReader.GetSecretAsync(secretName));
 
-            var updatedResult = _cache.AddOrUpdate(secretName, result, (key, old) => result);
-
-            return updatedResult.Value;
+            return _cache.AddOrUpdate(secretName, updatedValue, (key, old) => updatedValue)
+                         .Value;
         }
 
         private bool IsSecretOutdated(CachedSecret secret)
         {
-            return (DateTime.UtcNow - secret.CacheTimeUtc) >= _refreshInterval;
+            return (DateTime.UtcNow - secret.CacheTime) >= _refreshInterval;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,66 @@
...
-            if (secretReader == null)
+            _internalReader = secretReader ?? throw new ArgumentNullException(nameof(secretReader)); ;
+            _cache = new ConcurrentDictionary<string, CachedSecret>();
+
+            _refreshInterval = TimeSpan.FromSeconds(refreshIntervalSec);
+        }
+
+        public async Task<string> GetSecretAsync(string secretName)
+        {
+            // If the cache contains the secret and it is not expired, returned the cached value.
```
xavierdecoster: typo: `returned` > `return`

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,66 @@
...
         {
-            if (secretReader == null)
+            _internalReader = secretReader ?? throw new ArgumentNullException(nameof(secretReader)); ;
+            _cache = new ConcurrentDictionary<string, CachedSecret>();
+
+            _refreshInterval = TimeSpan.FromSeconds(refreshIntervalSec);
+        }
+
+        public async Task<string> GetSecretAsync(string secretName)
+        {
```
xavierdecoster: Should we have a null-check here?

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,66 @@
...
         {
-            return DateTime.UtcNow.Subtract(cachedSecret.Item2).TotalSeconds >= _refreshIntervalSec;
+            return (DateTime.UtcNow - secret.CacheTimeUtc) >= _refreshInterval;
         }
 
-        public async Task<string> GetSecretAsync(string secretName)
+        /// <summary>
+        /// A cached secret.
+        /// </summary>
+        private struct CachedSecret
```
xavierdecoster: Not sure whether `struct` is the right construct here, may be better off using `class`. This is not a short-lived object, is not immutable, and is likely larger than 16 bytes (`DateTime` is 8 bytes, `string` value most likely grows beyond 8 bytes - Unicode is 2 bytes per character).

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct

loic-sharma: Hm. My gut feeling is that since we know the lifetime of the `CachedSecret` and the `CachedSecret` *is* immutable after creation, a struct is better here. I haven't measured performance here, so I'll play it safe and use a class instead.

P.S.: `string` is implemented as [a `DWORD` and a `WCHAR` pointer](https://github.com/dotnet/coreclr/blob/master/src/vm/object.h#L1136), so a string is only 16 bytes on a 64-bit machine.

joelverhagen: Additionally, it's an anti-pattern to make mutable structs. I agree, `class` is more appropriate here.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,71 @@
...
     {
         public const int DefaultRefreshIntervalSec = 60 * 60 * 24; // 1 day
-        private readonly int _refreshIntervalSec;
 
         private readonly ISecretReader _internalReader;
-        private readonly Dictionary<string, Tuple<string, DateTime>> _cache;
+        private readonly ConcurrentDictionary<string, CachedSecret> _cache;
+        private readonly TimeSpan _refreshInterval;
 
         public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec)
```
joelverhagen: Why doesn't this ctor just accept a `TimeSpan`?

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,71 @@
...
+            if (string.IsNullOrEmpty(secretName))
+            {
+                throw new ArgumentException("Null or empty secret name", nameof(secretName));
+            }
+
+            // If the cache contains the secret and it is not expired, return the cached value.
+            if (_cache.TryGetValue(secretName, out CachedSecret result))
             {
-                throw new ArgumentNullException(nameof(secretReader));
+                if (!IsSecretOutdated(result))
```
joelverhagen: nit nested `if` can be collapsed to a single if with an `&&`.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -2,47 +2,71 @@
...
+            /// <summary>
+            /// The value of the cached secret.
+            /// </summary>
+            public string Value { get; set; }
 
-            return _cache[secretName].Item1;
+            /// <summary>
+            /// The time at which the secret was cached.
+            /// </summary>
+            public DateTime CacheTimeUtc { get; set; }
```
joelverhagen: Use `DateTimeOffset` so you don't need to specify UTC in the property name

src/NuGet.Services.ServiceBus/ScopedMessageHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Handle the message in its own dependency injection scope.
+        /// </summary>
+        /// <param name="message">The received message.</param>
+        /// <returns>Whether the message has been handled. If false, the message will be requeued to be handled again later.</returns>
+        public async Task<bool> HandleAsync(TMessage message)
+        {
+            // Create a new scope for this message.
+            using (var scope = _scopeFactory.CreateScope())
```
joelverhagen: Looks great! 👍

src/NuGet.Services.KeyVault/CachingSecretReader.cs
loic-sharma: Yes, that'd be ideal. Unfortunately, that's a breaking change to a public API, and I don't know the implications of this. I'd like to do the minimal possible work possible to unblock package signing.

joelverhagen: We are introducing a new type which must be adopted. Also, cleaning up our code will never be done unless make improvements along the way. Your call, tho.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
loic-sharma: Fixed

PR #86, opened 2017-11-16
Add ParentCertificate support

xavierdecoster:
A _ParentCertificate_ is referenced by an _EndCertificate_. We don't track hierarchy (not required). The combined set of EndCertificate and ParentCertificates allows us to reconstruct the certificate chain.

The relationship between these entities is described in the `CertificateChainLink` type.

image


src/NuGet.Services.Validation/Entities/ParentCertificate.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// The key to the end-<see cref="Certificate"/> of the certificate chain this parent-certificate is part of.
+        /// </summary>
+        public long EndCertificateKey { get; set; }
```
joelverhagen: Shouldn't this be called "Root"?

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// The SHA1 thumbprint that uniquely identifies this certificate. This is a string with exactly 40 characters.
+        /// </summary>
+        public string Thumbprint { get; set; }
```
joelverhagen: Isn't the thumbprint already on the `EndCertificate`? Why copy the value?

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// A X.509 Intermediary or Root Certificate used by one or more end-<see cref="Certificate" />s, 
+    /// together forming a certificate chain, used by one or more <see cref="PackageSignature"/>s.
+    /// </summary>
+    public class ParentCertificate
```
joelverhagen: Could this data also be represented as a an additional column on `Certificate`?

src/NuGet.Services.Validation/Migrations/201711161418120_AddParentCertificatesTable.cs
```diff
@@ -0,0 +1,31 @@
+namespace NuGet.Services.Validation
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddParentCertificatesTable : DbMigration
+    {
+        public override void Up()
```
joelverhagen: I think @loic-sharma has just been recreating the second (?) migration since this isn't deployed.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
loic-sharma: A `Certificate` will usually have 2+ `ParentCertificate`s - the intermediary Certificate Authority's certificate and the root Certificate Authority's certificate. However, this isn't always true since you can have as many intermediary CAs as you would like. To make matters worse, a certificate can have multiple valid chains.

TLDR: There is a 1-to-many relationship between a `Certificate` and `ParentCertificate`s, so a single column isn't enough.

src/NuGet.Services.Validation/Migrations/201711161418120_AddParentCertificatesTable.cs
xavierdecoster: That's indeed a pending question I have with @loic-sharma (via e-mail): should this be merged into the existing migration, or not?

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
loic-sharma: `EndCertificateKey` is correct as the "end certificate" is the `Certificate` that was used to sign the package.

Some background here: a signature *must* provide the certificate used to sign the package. This is called the "end certificate" and is represented by the `Certificate` entity. The signature may also provide additional certificates that should be used when validating the "end certificate" - these are the `ParentCertificate` entities. We don't care so much about the hierarchy of the end certificate and its parent certificates, we just to need to keep track of which parent certificates a end certificate has.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
loic-sharma: This is the thumbprint of the `ParentCertificate`. We need this to download the parent certificate from blob storage.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
xavierdecoster: @joelverhagen Unless I am missing something, only the end-certificate is stored in the `Certificates` table. The CA Root or Intermediary (parent) certs of the chain are linked to the end-certificate as a 1-to-many and stored in the `ParentCertificates` table. The thumbprint in the `ParentCertificates` table is that of the ParentCertificate itself.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
xavierdecoster: Nope, this is a FK to the end-certificate stored in `Certificates` table.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
xavierdecoster: Not a copy, this is the thumbprint of the parent certificate itself.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -337,6 +340,38 @@ private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
...
+                .HasColumnAnnotation(
+                    IndexAnnotation.AnnotationName,
+                    new IndexAnnotation(new[]
+                    {
+                        new IndexAttribute(ParentCertificatesEndCertificateKeyThumbprintIndex, 0)
+                    }));
+
+            modelBuilder.Entity<ParentCertificate>()
+                .Property(c => c.Thumbprint)
+                .HasMaxLength(20)
```
loic-sharma: This should be `40`, right? I think I goofed this value on the `Certificate` entity too... Could you fix that too?

xavierdecoster:
Looks like @loic-sharma and I were processing comments simultaneously, sorry about the duplicates :)

Question: perhaps we can avoid some confusion by renaming `Certificate` to `EndCertificate`, so we make it more explicit that we refer to the end-certificate, and not just _a_ certificate?

loic-sharma:
`EndCertificate` sounds good to me

src/NuGet.Services.Validation/Migrations/201711161418120_AddParentCertificatesTable.cs
loic-sharma: Yup, please merge the migrations together if possible. Thanks! :)

xavierdecoster:
Renamed `Certificate` (and related) to `EndCertificate` to avoid ambiguity.

image

src/NuGet.Services.Validation/Migrations/201711161418120_AddParentCertificatesTable.cs
xavierdecoster: Done

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The database-mastered identifier for this certificate.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// The key to the end-<see cref="Validation.EndCertificate"/> of the certificate chain this parent-certificate is part of.
+        /// </summary>
+        public long EndCertificateKey { get; set; }
```
joelverhagen: This should be a many to many relationship right?

loic-sharma: Good point. We shouldn't need duplicate entries for the same parent certificate used by different packages.

src/NuGet.Services.Validation/Entities/EndCertificate.cs
```diff
@@ -7,9 +7,9 @@
 namespace NuGet.Services.Validation
 {
     /// <summary>
-    /// A X.509 Certificate used by one or more <see cref="PackageSignature" />s for one or more <see cref="PackageSigningState"/>s.
+    /// A X.509 end-certificate used by one or more <see cref="PackageSignature" />s for one or more <see cref="PackageSigningState"/>s.
```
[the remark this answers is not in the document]
xavierdecoster: Fixed.

src/NuGet.Services.Validation/Entities/ParentCertificate.cs
xavierdecoster: Introducing an intermediary `CertificateChainLink` table to describe the relationship and avoid duplication.

image

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -287,79 +293,131 @@ private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
...
+            modelBuilder.Entity<ParentCertificate>()
+                .Property(c => c.Thumbprint)
+                .HasMaxLength(40)
+                .HasColumnType("char")
+                .IsRequired()
+                .HasColumnAnnotation(
+                    IndexAnnotation.AnnotationName,
+                    new IndexAnnotation(new[]
+                    {
+                        new IndexAttribute(ParentCertificatesThumbprintIndex)
```
joelverhagen: Thumbprint is unique now, right?

src/NuGet.Services.Validation/Entities/ValidationContext.cs
```diff
@@ -287,79 +293,131 @@ private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
...
+                .HasKey(t => t.Key);
+
+            modelBuilder.Entity<CertificateChainLink>()
+                .Property(v => v.EndCertificateKey)
+                .IsRequired()
+                .HasColumnAnnotation(
+                    IndexAnnotation.AnnotationName,
+                    new IndexAnnotation(new[]
+                    {
+                        new IndexAttribute(CertificateChainLinkEndCertificateKeyParentCertificateKeyIndex, 0)
```
joelverhagen: `(ParentCertificateKey, EndCertificateKey)` is unique, right?

src/NuGet.Services.Validation/Entities/EndCertificate.cs
```diff
@@ -78,6 +79,12 @@ public class Certificate
...
         /// of all validations performed for this certificate.
         /// </summary>
-        public virtual ICollection<CertificateValidation> Validations { get; set; }
+        public virtual ICollection<EndCertificateValidation> Validations { get; set; }
+
+        /// <summary>
+        /// An end-certificate is linked to its <see cref="ParentCertificate"/>s (Intermediary and/or Root certificates) by <see cref="CertificateChainLink"/>s.
+        /// Combined, this allows for revalidation of the certificate chain.
+        /// </summary>
+        public virtual ICollection<CertificateChainLink> CertificateChainLinks { get; set; }
```
loic-sharma: What do you think of using a direct many-to-many relationship between `EndCertificate` and `ParentCertificate`? Positives:

* No need for a `CertificateChainLink` property on `EndCertificate` - you can access `ParentCertificate`s directly.
* No need for a `CertificateChainLink` entity
* Simplifies the code to register the entities

Drawbacks:

* The joining table created by EF will be less flexible. In this current model, we could add more columns to the `CertificateChainLink` entity if we want to.

xavierdecoster: I kinda liked the idea of explicitly having a type representing the relationship, which going forward even allows us to expand and add columns that could e.g. describe hierarchy etc. Even though a bit more work to store/traverse the entities, it feels more correct to me.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
xavierdecoster: Yep, will add.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
xavierdecoster: Yep

xavierdecoster: Fixed.

src/NuGet.Services.Validation/Entities/ValidationContext.cs
xavierdecoster: Fixed.

xavierdecoster:
@joelverhagen you happy with the changes?

PR #87, opened 2017-11-16
Added ValidationResult to the contracts

loic-sharma:
This will be used by validators to report errors to the Gallery.

### DIFF
--- a/src/NuGet.Services.Contracts/Validation/ValidationResult.cs
+++ b/src/NuGet.Services.Contracts/Validation/ValidationResult.cs
@@ -8,56 +8,16 @@ namespace NuGet.Services.Validation
     /// <summary>
     /// The result of an asynchronous validation.
     /// </summary>
-    public class ValidationResult
+    public interface IValidationResult
     {
-        /// <summary>
-        /// Represents a validation result that has not been started.
-        /// </summary>
-        public static readonly ValidationResult NotStarted = new ValidationResult(ValidationStatus.NotStarted);
-
-        /// <summary>
-        /// Represents a validation result that has started but not succeeded or failed yet.
-        /// </summary>
-        public static readonly ValidationResult Incomplete = new ValidationResult(ValidationStatus.Incomplete);
-
-        /// <summary>
-        /// A successful validation result.
-        /// </summary>
-        public static readonly ValidationResult Succeeded = new ValidationResult(ValidationStatus.Succeeded);
-
-        /// <summary>
-        /// A failed validation result with no errors.
-        /// </summary>
-        public static readonly ValidationResult Failed = new ValidationResult(ValidationStatus.Failed);
-
-        /// <summary>
-        /// Create a new validation result with the given status.
-        /// </summary>
-        /// <param name="status">The result's status.</param>
-        public ValidationResult(ValidationStatus status)
-        {
-            Status = status;
-            Errors = new string[0];
-        }
-
-        /// <summary>
-        /// Create a new failed validation result with the given errors.
-        /// </summary>
-        /// <param name="errors">The errors that detail why the validation failed.</param>
-        public ValidationResult(params string[] errors)
-        {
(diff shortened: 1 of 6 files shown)


src/NuGet.Services.Contracts/Validation/ValidationResult.cs
```diff
@@ -0,0 +1,63 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// The result of an asynchronous validation.
+    /// </summary>
+    public class ValidationResult
```
joelverhagen: Only interfaces in `Contracts`. This implementation could easily live in `NuGet.Services.Validation`. The idea is that we can mock these integration points between components.

src/NuGet.Services.Contracts/Validation/ValidationResult.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// Represents a validation result that has not been started.
+        /// </summary>
+        public static readonly ValidationResult NotStarted = new ValidationResult(ValidationStatus.NotStarted);
```
joelverhagen: public fields are non-idiomatic.

src/NuGet.Services.Contracts/Validation/ValidationResult.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// The status of the validation.
+        /// </summary>
+        public ValidationStatus Status { get; }
+
+        /// <summary>
+        /// The errors that were encountered if the validation failed.
+        /// </summary>
+        public IEnumerable<string> Errors { get; }
```
joelverhagen: I'd prefer `IReadOnlyList` so that caller doesn't have to worry about re-enumeration.

tests/NuGet.Services.Contracts.Tests/ContractsFacts.cs
```diff
@@ -13,6 +14,7 @@ public void ShouldOnlyHaveInterfacesAndEnums()
         {
             // Arrange
             var assembly = typeof(ValidationStatus).Assembly;
+            var exceptions = new[] { typeof(ValidationResult) };
```
joelverhagen: no exceptions! Or `Exceptions`s for that matter

src/NuGet.Services.Contracts/Validation/ValidationResult.cs
loic-sharma: Oh. Sorry, I thought that we settled on a class from our discussion.

joelverhagen: 😊 I must have been unclear. Sorry about the confusion.

src/NuGet.Services.Validation/ValidationResult.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            Status = status;
+            Errors = new string[0];
+        }
+
+        /// <summary>
+        /// Create a new failed validation result with the given errors.
+        /// </summary>
+        /// <param name="errors">The errors that detail why the validation failed.</param>
+        public ValidationResult(params string[] errors)
```
agr: Add another constructor that accepts `IEnumerable<string>` and copy the contents maybe?

agr: Also, a constructor that defaults to failed status seems confusing. Make it a static method maybe?

joelverhagen: Ah good point!! The `string[]` can be mutated after this ctor breaking the immutability of this type. It should be copied or this ctor should accept `IReadOnlyList<string>`.

PR #89, opened 2017-11-17
Enforce sequential GUIDs in PackageValidations table in Validation DB

agr:
Updated AddValidationSchema migration to make sure PackageValidations table key default value is sequential guid

### DIFF
--- a/src/NuGet.Services.Validation/Migrations/201709202218143_AddValidationSchema.cs
+++ b/src/NuGet.Services.Validation/Migrations/201709202218143_AddValidationSchema.cs
@@ -11,7 +11,7 @@ namespace NuGet.Services.Validation
                 "dbo.PackageValidations",
                 c => new
                     {
-                        Key = c.Guid(nullable: false, identity: true),
+                        Key = c.Guid(nullable: false, identity: true, defaultValueSql: "newsequentialid()"),
                         PackageValidationSetKey = c.Long(nullable: false),
                         Type = c.String(nullable: false, maxLength: 255, unicode: false),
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.Validation/Migrations/201709202218143_AddValidationSchema.cs
```diff
@@ -11,7 +11,7 @@ public override void Up()
                 "dbo.PackageValidations",
                 c => new
                     {
-                        Key = c.Guid(nullable: false, identity: true),
+                        Key = c.Guid(nullable: false, identity: true, defaultValueSql: "newsequentialid()"),
```
joelverhagen: This should be done on the entity context, not the auto-generated migration.

agr: I'm not sure it's possible. Besides, this auto-generated migration cannot be regenerated (otherwise there wouldn't be a need to store it in source control), so we should be fine with updating it manually, assuming we don't mess it up so that up/down actions don't work properly.

joelverhagen: This is that the `.HasColumnType(...)` extension method is for.

Can't we just rebuild the database? There is no critical data there now. If this is too much work, feel free to reject.

joelverhagen: If you add that then the resulting migration should be the same right?

PR #90, opened 2017-11-21
Download nuget.exe v4.4.1

edwardmeng:
In the windows os with Simplified Chinese culture, there will be some strange errors when packaging with nuget.exe 4.1.0.
But packaging with the latest version nuget.exe, the errors disappeared.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -300,7 +300,7 @@ Function Install-NuGet {
     }
     
     Trace-Log 'Downloading latest prerelease of nuget.exe'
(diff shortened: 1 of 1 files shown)


skofman1:
Thanks for the PR @edwardmeng ! We want our builds to depend on a specific nuget.exe version. This guarantees reproducible builds. Please update the PR to download v4.4.1 explicitly.

joelverhagen:
I moved the base branch to `dev`, which is our practice. Will merge momentarily.

joelverhagen:
Merged to `dev` with 2b1c9e9dc10bd675450d0195b55e2889abce6afe. Thanks @edwardmeng!

PR #91, opened 2017-11-24
Merging dev into master

xavierdecoster:
To create a stable release of the ServerCommon nupkgs for our NuGet.Jobs pending deployments.

Q: This contains breaking changes in terms of API (e.g. rename of `Certificate` to `EndCertificate` etc), should this be a v3.0.0 release of the nupkgs? (we should actually version our packages independently, but that's a different story)


loic-sharma:
The "breaking" changes are all happening to pre-released packages. I don't think we need to do a v3.0.0 release. Silly question: where do we do the version bump? Is it on VSTS after the master branch has been committed?

xavierdecoster:
Ok, v2.2.4 it is then :)

I'm not sure about the version bump location, but my guess would also be that this is tracked in the VSTS build definition.

xavierdecoster:
Confirmed: the version is set in a VSTS build definition variable.

PR #93, opened 2017-11-27
Revert "Added ValidationResult to the contracts"

loic-sharma:
Reverts NuGet/ServerCommon#87 to unblock `NuGet.Jobs`' `dev-package-signing` -> `dev` merge.


loic-sharma:
My bad, this also reverts #86. This was not intended.

PR #98, opened 2017-11-28
Change package signature schema to support SHA-256 fingerprints, instead of SHA-1

joelverhagen:
Per conversation with Barry, we're to use SHA-256 on the server side.

This is recreating the schema since there is no data we care about in the environments this is deployed to (DEV and INT).

I will release a new minor version (2.5.0?) of the ServerCommon packages to consume in NuGet.Jobs.


src/NuGet.Services.Validation/Entities/ValidationEntitiesContext.cs
```diff
@@ -304,7 +310,7 @@ private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
 
             modelBuilder.Entity<EndCertificate>()
                 .Property(c => c.Thumbprint)
-                .HasMaxLength(40)
+                .HasMaxLength(FingerprintLength)
                 .HasColumnType("char")
```
loic-sharma: If you'd like to move from `char` to `varchar`, feel free to do it in this PR :)

joelverhagen: Fixed.

joelverhagen:
This is progress on https://github.com/NuGet/Engineering/issues/935.

PR #101, opened 2017-11-30
Add Validation Error Messages to Contracts and Validation Entities

loic-sharma:
This change does two things:

1. Updates the Orchestrator's `IValidator` API. Instead of returning an `IValidationStatus`, validators will now return `IValidationResult`. A validation result has a validation status and 0 or more `IValidationError`s. Each `IValidationError` has an error code (`ValidationErrorCode`) and a dictionary of data so that the Gallery can properly format the error message.
2. Adds `PackageValidationError` entity. Each `PackageValidation` has 0 or more `PackageValidationError`.

The migration's SQL is:

```
CREATE TABLE [dbo].[PackageValidationIssues] (
    [Key] [bigint] NOT NULL IDENTITY,
    [PackageValidationKey] [uniqueidentifier] NOT NULL,
    [IssueCode] [int] NOT NULL,
    [Data] [nvarchar](max),
    CONSTRAINT [PK_dbo.PackageValidationIssues] PRIMARY KEY ([Key])
)
CREATE INDEX [IX_PackageValidationKey] ON [dbo].[PackageValidationIssues]([PackageValidationKey])
ALTER TABLE [dbo].[PackageValidationIssues] ADD CONSTRAINT [FK_dbo.PackageValidationIssues_dbo.PackageValidations_PackageValidationKey] FOREIGN KEY ([PackageValidationKey]) REFERENCES [dbo].[PackageValidations] ([Key]) ON DELETE CASCADE
```

**NOTE**: I'm merging this into my own topic branch `loshar-errors`.


src/NuGet.Services.Contracts/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// The error code for a type of asynchronous validation failure.
+    /// </summary>
+    public enum ValidationErrorCode
+    {
+        // Used when signed packages are not accepted.
+        PackageIsSignedError = 1,
```
joelverhagen: XML docs?

Also, we should document the set of arguments associated with this code. I think it would be best to have a new code every time the arguments change.

src/NuGet.Services.Validation/Entities/PackageValidationError.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// to an actual error message using this code.
+        /// </summary>
+        public ValidationErrorCode ErrorCode { get; set; }
+
+        /// <summary>
+        /// The error message's serialized data. This is used to store <see cref="Data"/> in the
+        /// database and is NOT intended to be used directly.
+        /// </summary>
+        [Column("Data")]
+        public string DataAsJson {
```
joelverhagen: IMO the deserialization should not live in the EF entity

src/NuGet.Services.Contracts/Validation/IValidationResult.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Validation
+{
+    /// <summary>
+    /// The result of an asynchronous validation.
+    /// </summary>
+    public interface IValidationResult
```
joelverhagen: Can we create one initial implementation of this for `PackageIsSignedError`?

src/NuGet.Services.Contracts/Validation/ValidationErrorCode.cs
loic-sharma: Fixed. I agree that a new code should be created on every argument change (unless in a backwards compatible way, like removing arguments)

src/NuGet.Services.Validation/Entities/PackageValidationError.cs
loic-sharma: Fixed. The deserialization logic is no longer on entity.

src/NuGet.Services.Contracts/Validation/IValidationResult.cs
loic-sharma: Good idea, added in latest commit.

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        [Fact]
+        public void TheSerializeMethodShouldSerializeAllPublicPropertiesExceptTheErrorCode()
+        {
+            // Arrange
+            var signedError = new PackageIsSignedError("Hello.World", "1.3.4");
+            var result = signedError.Serialize();
+
+            // Assert
+            Assert.Equal("{\"PackageId\":\"Hello.World\",\"PackageVersion\":\"1.3.4\"}", result);
```
scottbommarito: You should share this string with the `PackageIsSignedDeserialization` test below.

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class ValidationErrorFacts
+    {
+        [Fact]
+        public void TheErrorCodeTypesPropertyValuesAllExtendValidationError()
+        {
+            Assert.True(ValidationError.ErrorCodeTypes.Values.All(t => t.IsSubclassOf(typeof(ValidationError))));
+        }
+
+        [Fact]
+        public void TheSerializeMethodShouldSerializeAllPublicPropertiesExceptTheErrorCode()
```
scottbommarito: nit: I think this would be better if this was inside `public class TheSerializeMethod` and name the test `PackageIsSignedSerialization`

src/NuGet.Services.Contracts/Validation/ValidationErrorCode.cs
```diff
@@ -8,7 +8,14 @@ namespace NuGet.Services.Validation
     /// </summary>
     public enum ValidationErrorCode
```
scottbommarito: nit: I would link to `ErrorCodeTypes` in the XML doc for this.

src/NuGet.Services.Contracts/Validation/IValidationResult.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// The status of the validation.
+        /// </summary>
+        ValidationStatus Status { get; }
+
+        /// <summary>
+        /// The errors that were encountered if <see cref="Status"/> is <see cref="ValidationStatus.Failed"/>.
```
joelverhagen: We should qualify that this will be empty if `ValidationStatus` is not `Failed`

src/NuGet.Services.Errors/UnknownError.cs
```diff
@@ -0,0 +1,23 @@
...
+using NuGet.Services.Validation;
+
+namespace NuGet.Services.Errors
+{
+    /// <summary>
+    /// A placeholder error when the specific error is unknown.
+    /// </summary>
+    public class UnknownError : ValidationError
+    {
+        public static UnknownError Instance = new UnknownError();
```
joelverhagen: `readonly`

src/NuGet.Services.Errors/UnknownError.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        private UnknownError()
```
joelverhagen: `public`. Making this private doesn't give us much. `Newtonsoft.Json` will activate this (ignoring the `private`) but `Activator` will not.

src/NuGet.Services.Errors/ValidationError.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace NuGet.Services.Errors
+{
+    public abstract class ValidationError : IValidationError
+    {
+        /// <summary>
+        /// The map of error codes to the type that represents the error. The types MUST extend <see cref="ValidationError"/>.
+        /// </summary>
+        public static readonly Dictionary<ValidationErrorCode, Type> ErrorCodeTypes = new Dictionary<ValidationErrorCode, Type>
+        {
+            { ValidationErrorCode.PackageIsSignedError, typeof(PackageIsSignedError) },
```
joelverhagen: You can replace `typeof` with
```csharp
private static Type GetErrorType<T>() where T : ValidationError
{
    return typeof(T);
}
```

to help developers do the right thing.

Also, `IReadOnlyDictionary`

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using NuGet.Services.Validation;
+using Xunit;
+
+namespace NuGet.Services.Errors.Tests
+{
+    public class ValidationErrorFacts
+    {
```
joelverhagen: Test serialization of `UnknownError`. Alternatively, serialization of `UnknownError` should fail.

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using NuGet.Services.Validation;
+using Xunit;
+
+namespace NuGet.Services.Errors.Tests
```
joelverhagen: Assert the values of `ValidationErrorCode`. These must never change.

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
```diff
@@ -0,0 +1,66 @@
...
+    <Compile Include="PackageIsSignedError.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Properties\AssemblyInfo.*.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\NuGet.Services.Contracts\NuGet.Services.Contracts.csproj">
+      <Project>{6674B4B4-2D0E-4840-8CF0-2356ACDE8863}</Project>
+      <Name>NuGet.Services.Contracts</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\NuGet.Services.Validation\NuGet.Services.Validation.csproj">
```
joelverhagen: Ideally the presentation of errors is not bound to the persistence of errors. Additionally, this project name is very generic suggesting we can put other errors here too. Having the Validation project dependency is weird. AFAIK, the only reason you need this is the `FromPackageValidationError` method. This can be avoided by making `ErrorCode` and `Data` a parameter. Concerned about too many parameters in the future? Add a `IValidationErrorRecord` or `ISerializedValidationError` in the future, make the EF entity implement it, and make that the parameter although I don't see that happening.

src/NuGet.Services.Contracts/Validation/IValidationError.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// A validation error that failed an asynchronous validation.
+    /// </summary>
+    public interface IValidationError
+    {
+        /// <summary>
+        /// The code that classifies this error.
+        /// </summary>
+        ValidationErrorCode ErrorCode { get; }
```
joelverhagen: In the future, how do you think we'll represent validation warnings?

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
```diff
@@ -0,0 +1,66 @@
...
+      <Project>{c1e36a2c-1c1b-4521-b256-ad42505d9efb}</Project>
+      <Name>NuGet.Services.Validation</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json">
```
joelverhagen: Try packing this and make sure you get the write package dependencies and versions. I think I had to use `project.json` for some reason...

src/NuGet.Services.Errors/ValidationError.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using NuGet.Services.Validation;
+
+namespace NuGet.Services.Errors
+{
+    public abstract class ValidationError : IValidationError
```
joelverhagen: Suppose gallery gets N validation errors, should it be worried about de-duping? Like if two validates give the same error? 

The code that takes a package ID and version and returns all validation errors -- where will that live? Will it de-dupe? How? Does gallery even care?

joelverhagen:
Nice feature branch 🥇

src/NuGet.Services.Contracts/Validation/IValidationError.cs
joelverhagen: You can say out of scope, just wondering if you had any ideas that could effect this design. I haven't thought it through.

src/NuGet.Services.Errors/PackageIsSignedError.cs
```diff
@@ -0,0 +1,24 @@
...
+            PackageVersion = packageVersion;
+        }
+
+        public override ValidationErrorCode ErrorCode => ValidationErrorCode.PackageIsSignedError;
+
+        public string PackageId { get; }
+
+        public string PackageVersion { get; }
+
+        public override string GetMessage() => $"Package {PackageId} {PackageVersion} is signed.";
```
joelverhagen: nit: resource file

src/NuGet.Services.Validation/Entities/ValidationEntitiesContext.cs
```diff
@@ -171,6 +172,13 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                 .Property(pv => pv.RowVersion)
                 .IsRowVersion();
 
+            modelBuilder.Entity<PackageValidationError>()
+                .HasKey(e => e.Key);
+
+            modelBuilder.Entity<PackageValidationError>()
+                .Property(pv => pv.Key)
+                .HasDatabaseGeneratedOption(DatabaseGeneratedOption.Identity);
```
joelverhagen: Is `HasDatabaseGeneratedOption` really necessary? I thought that this was implicit for `long` keys

src/NuGet.Services.Validation/Errors/PackageIsSignedError.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Validation
+{
+    public class PackageIsSignedError : AbstractError
```
joelverhagen: How does this related to the other `PackageIsSignedError`?

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
joelverhagen: Feel free to "won't fix" this one. It's just odd to me to have a generic sounding library name referencing a specific component.

src/NuGet.Services.Errors/ValidationError.cs
loic-sharma: I don't think that a method adds that much protection as this is enforced by a unit test. Thoughts?

joelverhagen: Both. Unit tests are are a fail safe, helper method is to guide future you down the right path.

src/NuGet.Services.Validation/Errors/PackageIsSignedError.cs
loic-sharma: Sorry, this was supposed to be removed. Will fix.

src/NuGet.Services.Validation/Entities/ValidationEntitiesContext.cs
loic-sharma: Not sure. Even if it is implicit, I would rather keep it explicit.

src/NuGet.Services.Contracts/Validation/ValidationErrorCode.cs
loic-sharma: Hm, unfortunately `ValidationErrorCode` and `ErrorCodeTypes` are in separate projects and the `NuGet.Services.Contracts` project does not have a reference to `NuGet.Services.Validation.Errors`. AKA, this documentation link would require us to introduce a reference that we are trying to avoid adding. If you feel strongly about this, I can add the type's name in the documentation without an actual link.

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
loic-sharma: It seems that pack is skipping the `Newtonsoft.Json` dependency... Weird. I'll try using `project.json`.

src/NuGet.Services.Errors/UnknownError.cs
loic-sharma: I will make the constructor for this class public and remove the `Instance` property.

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
loic-sharma: Fixed

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
loic-sharma: Fixed (added resource file)

src/NuGet.Services.Contracts/Validation/IValidationResult.cs
loic-sharma: Fixed

src/NuGet.Services.Errors/UnknownError.cs
loic-sharma: Fixed

src/NuGet.Services.Errors/ValidationError.cs
loic-sharma: Fixed

NuGet.Services.Errors.Tests/ValidationErrorFacts.cs
loic-sharma: Fixed

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
loic-sharma: Fixed:

1. Renamed `FromPackageValidationError` to `Deserialize` that accepts `ErrorCode` and `Data`. I don't think more parameters should be necessary for deserialization, so it should be good enough for now.
2. Renamed the `NuGet.Services.Validation` project to `NuGet.Services.Validation.Errors` project as this project is just meant just for validation errors

src/NuGet.Services.Errors/NuGet.Services.Errors.csproj
loic-sharma: Fixed

src/NuGet.Services.Errors/PackageIsSignedError.cs
loic-sharma: Fixed

src/NuGet.Services.Validation/Errors/PackageIsSignedError.cs
loic-sharma: Fixed

src/NuGet.Services.Errors/ValidationError.cs
loic-sharma: Talked to Joel offline. Our current plan is for the Gallery to dedup off the generated messages. (AKA, if two validation issues have the same message, only one error will be shown).

src/NuGet.Services.Contracts/Validation/IValidationError.cs
loic-sharma: Talked to Joel offline. We renamed `IValidationError` to `IValidationIssue`. Issues include both errors and warnings. If in the future we need to differentiate errors and warnings, we can add a property to `ValidationIssue` like `IsWarning` or `IsError`.

src/NuGet.Services.Errors/ValidationError.cs
xavierdecoster: Q: should this type implement [`IEquatable<T>`](https://msdn.microsoft.com/en-us/library/ms131187(v=vs.110).aspx) so consumers can simply call into [`Enumerable.Distinct<ValidationError>()`](https://msdn.microsoft.com/en-us/library/bb348436(v=vs.110).aspx) to get default de-dupe behavior?

src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// An unknown issue has occurred.
+        /// </summary>
+        Unknown = 0,
+
+        /// <summary>
+        /// Signed packages are blocked.
+        /// </summary>
+        PackageIsSignedError = 1,
```
joelverhagen: Seems a bit weird to say "Error" in this code. Perhaps just `PackageIsSigned`?

src/NuGet.Services.Validation/Migrations/201712060116551_AddPackageValidationIssues.cs
```diff
@@ -0,0 +1,32 @@
...
+        public override void Up()
+        {
+            CreateTable(
+                "dbo.PackageValidationIssues",
+                c => new
+                    {
+                        Key = c.Long(nullable: false, identity: true),
+                        PackageValidationKey = c.Guid(nullable: false),
+                        IssueCode = c.Int(nullable: false),
+                        Data = c.String(),
```
joelverhagen: non-nullable? seems like errors with no data should have column value `{}`.

src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
loic-sharma: I don't feel strongly about this, but, I think the "Error" is useful as multiple severity levels could be placed within `ValidationIssueCode`. For example, we may other things like `IdAndTitleDontMatchWarning`. Thoughts?

src/NuGet.Services.Validation/Migrations/201712060116551_AddPackageValidationIssues.cs
loic-sharma: Fixed

src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
loic-sharma: Talked to Joel offline. We may decide to treat an issue as a warning/error depending on security policies. Hence, we should keep issues agnostic to the severity at which they were reported.

PR #102, opened 2017-12-07
Expose Exists API on IStorage + add Async overload

xavierdecoster:
All of the concrete implementations of `IStorage` already implemented the `Exists(string)` method, but it was not exposed on the interface. This PR adds it to the interface, and adds an async overload (which is useful when consuming a storage abstraction that supports async I/O).

Will be used by `ICertificateStore` implementation in validation job.

### DIFF
--- a/src/NuGet.Services.Storage/AzureStorage.cs
+++ b/src/NuGet.Services.Storage/AzureStorage.cs
@@ -78,6 +78,24 @@ namespace NuGet.Services.Storage
             return false;
         }
 
+        public override async Task<bool> ExistsAsync(string fileName, CancellationToken cancellationToken)
+        {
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -78,6 +78,24 @@ public override bool Exists(string fileName)
             return false;
         }
 
+        public override async Task<bool> ExistsAsync(string fileName, CancellationToken cancellationToken)
+        {
+            Uri packageRegistrationUri = ResolveUri(fileName);
```
loic-sharma: Can we use `var` here? :)

xavierdecoster: Fixed.

PR #106, opened 2017-12-12
Round trip bytes properly and allow no overwrite in Storage abstraction

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/5175.

Tests are skipped because they need Azure Storage emulator to run them.

This PR is currently blocking https://github.com/NuGet/NuGet.Jobs/pull/280.

### DIFF
--- a/src/NuGet.Services.Storage/AzureStorage.cs
+++ b/src/NuGet.Services.Storage/AzureStorage.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -161,29 +177,27 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
                 {
-                    using (var reader = new StreamReader(originalStream))
-                    {
-                        content = await reader.ReadToEndAsync();
-                    }
+                    content = originalStream;
                 }
 
-                return new StringStorageContent(content);
+                return new StreamStorageContent(content);
```
scottbommarito: I'd be very careful about this...need to enumerate through all of the uses of this library (you can ignore `Services.Metadata` for now) and guarantee that none do any `StringStorageContent` casting. I think `NuGetMonitoring` uses this package.

tests/NuGet.Services.Storage.Tests/AzureStorageFacts.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        private const string NoEmulator = "The Azure Storage emulator is not running on the CI.";
+        private const string ContainerName = "test";
+        private static readonly Uri BaseAddress = new Uri("http://example/dir/");
+        private const string BlobName = "blob.txt";
+        private const string InitialContent = "pre-existing content";
+
+        public class Load : BaseFacts
+        {
+            [Theory(Skip = NoEmulator)]
```
scottbommarito: I remember when I was working on the Queue API you suggested abstracting the actual storage calls in a way such that you could unit test `AzureStorage` without needing an actual Azure storage connection. Would it be difficult to do that so that we can run these tests with CI?

joelverhagen: I chose not to do this here since I would rather invest in the future in moving to a single abstraction (maybe NuGetGallery.Core's).

src/NuGet.Services.Storage/AzureStorage.cs
joelverhagen: Good idea, I'll take a look. That being said, this does not immediately break anyone (they will have to update package dependencies).

joelverhagen: No casting.

PR #107, opened 2017-12-13
Fix test failure due to signature change of `Storage.OnSave` method

xavierdecoster:
Yeay for reflection :)
@joelverhagen


joelverhagen:
Hm, my bad. Should have caught that.

PR #110, opened 2017-12-20
Better exception logging in SubscriptionProcessor.

agr:
Previously exception was logged in the message body converting the exception object to string, that caused truncation and loss of full stack trace.
By logging it this way we'll have structured exception information in AI entry.

### DIFF
--- a/src/NuGet.Services.ServiceBus/Event.cs
+++ b/src/NuGet.Services.ServiceBus/Event.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.Services.ServiceBus
+{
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.ServiceBus/Event.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace NuGet.Services.ServiceBus
+{
+    public static class Event
+    {
+        // should be sufficiently different from other event id's, but no strict
+        // requirement, numbers mean nothing so far
+        private const int EventIdBase = 600; 
+
+        public static EventId SubscriptionMessageHandlerException = new EventId(EventIdBase + 1, "Subscription event handler threw exception");
```
shishirx34: may be make an exception event for subscriptionMessage as 601 instead of using a base and use it here.

agr: Added a constant, left `EventId + 1` in it's definition as I think it better communicates the intent.

PR #111, opened 2018-01-04
Update "Package Is Signed" Validation Error Message

loic-sharma:
Updates the validation error message for packages that are signed. This needs to be merged before validation issues are deployed to the gallery.

NOTE: Previously, `PackageIsSignedIssue` was used by tests to validate deserialization of validation issues. As part of this change, `PackageIsSignedIssue` no longer stores any data in the database, and is no longer a good candidate for testing deserialization. Thus, this change adds an `ObsoleteTestingIssue` class that can be used to properly test deserialization.

@karann-msft


src/NuGet.Services.Validation.Issues/PackageIsSigned.cs
```diff
@@ -7,28 +7,11 @@ namespace NuGet.Services.Validation.Issues
...
-        }
-
         public override ValidationIssueCode IssueCode => ValidationIssueCode.PackageIsSigned;
 
-        public string PackageId { get; }
-
-        public string PackageVersion { get; }
-
-        public override string GetMessage() => $"Package {PackageId} {PackageVersion} is signed.";
+        public override string GetMessage() => "Package publishing failed: This package could not be published since it is signed. " +
```
scottbommarito: Should we note that
```
If you would still like to upload this package, you can upload an unsigned version of this package. Otherwise, please wait until the gallery accepts signed packages before publishing this package.
```

joelverhagen: Can we put this in a Strings.resx please?

tests/NuGet.Services.Validation.Issues.Tests/NuGet.Services.Validation.Issues.Tests.csproj
```diff
@@ -47,7 +47,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="Data\PackageIsSigned.json">
+    <None Include="Data\PackageIsSignedMessage.txt">
```
joelverhagen: If you really need this test (seems a little silly), make this a resource file to avoid current working directory weirdness.

src/NuGet.Services.Validation.Issues/PackageIsSigned.cs
loic-sharma: @scottbommarito: The message continues on lines 13, 14, and 15 (it also includes links to get more information). I think that the current message covers what your suggested message says. Thoughts?

tests/NuGet.Services.Validation.Issues.Tests/NuGet.Services.Validation.Issues.Tests.csproj
loic-sharma: Added a resource file. Created `ObsoleteTestingIssue` to properly test deserialization, and removed the unnecessary `PackageIsSigned` deserialization test (as that was what previously tested deserialization).

src/NuGet.Services.Validation.Issues/ObsoleteTestingIssue.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+
+namespace NuGet.Services.Validation.Issues
+{
+    /// <summary>
+    /// A validation issue used for testing purposes.
+    /// </summary>
+    [Obsolete("This issue should only be used for testing")]
+    public class ObsoleteTestingIssue : ValidationIssue
```
loic-sharma: This is used for unit testing `ValidationIssue` deserialization. See PR description for more info.

PR #118, opened 2018-01-12
Remove validation issue requiring exactly one signature

joelverhagen:
This is not possible per the spec and soon the not possible via the client APIs.

Spoke to @mishra14 and @dtivel about this offline.

Normally removing issues is not allowed, but this was never merged to `master` so it is safe.


loic-sharma:
Yay for simpler code! :)

PR #120, opened 2018-01-17
[Service Bus] Add timeout to shutdown

loic-sharma:
Today, the Orchestrator has custom code to ensure that:

1. New Service Bus messages are dropped once shutdown has started
2. Shutting down Service Bus does not block the current thread

This brings the logic to the `SubscriptionProcessor` so that all other jobs can also have these guarantees.

Part of https://github.com/NuGet/Engineering/issues/1115

### DIFF
--- a/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
+++ b/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
@@ -83,13 +83,29 @@ namespace NuGet.Services.ServiceBus
             }
         }
 
-        public async Task StartShutdownAsync()
+        public async Task<bool> StartShutdownAsync(TimeSpan timeout)
         {
             _logger.LogInformation(
                 "Shutting down the subscription listener with {NumberOfMessagesInProgress} messages in progress",
                 NumberOfMessagesInProgress);
 
-            await _client.CloseAsync();
+            // Start two tasks: one for the shutdown and one for the shutdown's timeout. The first task to finish will
+            // complete the "start shutdown" operation. Note that this method returns false if the timeout task completes first.
+            var shutdownTask = _client.CloseAsync();
+            var shutdownTimeoutTask = Task.Delay(timeout);
+
+            var shutdownResult = await Task.WhenAny(shutdownTask, shutdownTimeoutTask);
+
+            if (shutdownResult == shutdownTimeoutTask)
+            {
+                _logger.LogWarning(
+                    "Timeout reached when starting shutdown of subscription processor after {StartShutdownTimeout}",
+                    timeout);
+
+                return false;
+            }
+
+            return true;
         }
     }
 }
--- a/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
+++ b/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
@@ -16,6 +16,7 @@ namespace NuGet.Services.ServiceBus
         private readonly IMessageHandler<TMessage> _handler;
         private readonly ILogger<SubscriptionProcessor<TMessage>> _logger;
 
+        private bool _running;
         private int _numberOfMessagesInProgress;
 
         public int NumberOfMessagesInProgress => _numberOfMessagesInProgress;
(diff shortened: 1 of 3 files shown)


tests/NuGet.Services.ServiceBus.Tests/SubscriptionProcessorFacts.cs
```diff
@@ -157,11 +157,31 @@ public class TheStartShutdownAsyncMethod : Base
...
 
                 // Assert
                 _client.Verify(c => c.CloseAsync(), Times.Once);
             }
+
+            [Fact]
+            public async Task StopReturnsTrueIfClientCloseAsyncMethodFinishesFirst()
+            {
+                // Arrange
+                _client.Setup(c => c.CloseAsync()).Returns(Task.Delay(TimeSpan.FromMilliseconds(5)));
```
chenriksson: Does it even need a 5ms sleep, or can we reduce to 1 or 0?

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -83,13 +83,29 @@ public void Start()
...
         {
             _logger.LogInformation(
                 "Shutting down the subscription listener with {NumberOfMessagesInProgress} messages in progress",
                 NumberOfMessagesInProgress);
 
-            await _client.CloseAsync();
+            // Start two tasks: one for the shutdown and one for the shutdown's timeout. The first task to finish will
+            // complete the "start shutdown" operation. Note that this method returns false if the timeout task completes first.
+            var shutdownTask = _client.CloseAsync();
+            var shutdownTimeoutTask = Task.Delay(timeout);
```
chenriksson: This isn't really a timeout for CloseAsync, just a timeout for how long to wait.

It looks like there's a ServiceBus OperationTimeout, settable via the messaging factory. I wonder if that's really the approach we should be taking. Found the timeout via reflector, but see also stackoverflow:
https://stackoverflow.com/questions/24416435/setting-the-operationtimeout-property-for-a-service-bus-worker-role

tests/NuGet.Services.ServiceBus.Tests/SubscriptionProcessorFacts.cs
loic-sharma: Ideally we'd use `TimeSpan.MaxValue` and `TimeSpan.MinValue`, but, `Task.Delay` was throwing exceptions. A human cannot tell the difference between 1ms or 5ms, so I don't think there's any value in changing.

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
loic-sharma: That is correct. Let me know if you think the documentation on the interface for this method is confusing. It seems that `OperationTimeout` does not relate to the time the `CloseAsync` method takes, but rather, the time a thread can take to process a message. This fix is still necessary.

chenriksson: Documentation is clear. Assuming the process will just exit if we don't block on a hung `CloseAsync`?

Re: `OperationTimeout` does not relate to the time the `CloseAsync` method takes...
This is what I was wondering; whether OperationTimeout would prevent a `CloseAsync` hang due to processing of messages. Looks like it defaults to 1 min, so makes sense if this isn't related.

Seems ok for now, but do we know why `CloseAsync` hangs? Just want to make sure we clean up properly and don't have side effects.

loic-sharma: Kind of... the process won't end if we're blocked on `CloseAsync`, but, other factors may prevent the process from exiting.

We're not sure what causes the `CloseAsync` hang, @agr: is still investigating.

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -53,6 +57,12 @@ public void Start()
 
         private async Task OnMessageAsync(IBrokeredMessage brokeredMessage)
         {
+            if (_running == false)
```
joelverhagen: `!_running`?

PR #122, opened 2018-01-22
Add end certificate use column to make detection of multi-use dual cetificates easier

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5327.


src/NuGet.Services.Validation/Entities/EndCertificate.cs
```diff
@@ -29,6 +29,12 @@ public class EndCertificate
         /// </summary>
         public EndCertificateStatus Status { get; set; }
 
+        /// <summary>
+        /// The use of this end certificate. Today, this is a convenience property implied by
+        /// the presence of <see cref="PackageSignatures"/> and <see cref="TrustedTimestamps"/>.
```
loic-sharma: Nit: I think `and` should be `or` here, right?

joelverhagen: Yup, today that's true.

joelverhagen: Fixed.

PR #136, opened 2018-02-09
Scope around ServiceBus message processing

agr:
Added a scope around message processing in SubscriptionProcessor that should make it easier to correlate log entries across library boundaries.

A guid is generated per call, plus `DateTimeOffset.UtcNow.Ticks` is reported. The latter would allow grouping and ordering single call logs by start time. The former would allow to distinguish between calls in case they are close enough so that Ticks are the same.

### DIFF
--- a/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
+++ b/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
@@ -73,19 +73,24 @@ namespace NuGet.Services.ServiceBus
 
             try
             {
-                _logger.LogInformation("Received message from Service Bus subscription, processing");
-
-                var message = _serializer.Deserialize(brokeredMessage);
-
-                if (await _handler.HandleAsync(message))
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -73,19 +73,24 @@ public void Start()
 
             try
             {
-                _logger.LogInformation("Received message from Service Bus subscription, processing");
+                using (var scope = _logger.BeginScope($"{nameof(SubscriptionProcessor<TMessage>)}.{nameof(OnMessageAsync)} {{CallGuid}} {{CallStartTimestamp}}",
+                    Guid.NewGuid(),
+                    DateTimeOffset.UtcNow.Ticks))
```
joelverhagen: Why a timestamp? I presume the Guid is so you can correlate all logs related to this message.

joelverhagen: Nevermind, I see you comment about time. I would prefer an ISO timestamp over ticks.

agr: Done

PR #137, opened 2018-02-16
[Orchestrator] Expose the Service Bus delivery count

loic-sharma:
Adds the Service Bus message's delivery count to `PackageValidationMessageData`. This will allow the Orchestrator to make decisions based off how many times a validation message has been tried.

Progress on https://github.com/NuGet/Engineering/issues/1100

### DIFF
--- a/src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
+++ b/src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
@@ -10,7 +10,9 @@ namespace NuGet.Services.ServiceBus
     public interface IBrokeredMessage : IDisposable
     {
         int DeliveryCount { get; }
+        DateTimeOffset ExpiresAtUtc { get; }
         IDictionary<string, object> Properties { get; }
+        DateTimeOffset EnqueuedTimeUtc { get; }
         DateTimeOffset ScheduledEnqueueTimeUtc { get; set; }
         Task CompleteAsync();
         Task AbandonAsync();
--- a/src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
+++ b/src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
@@ -9,6 +9,7 @@ namespace NuGet.Services.ServiceBus
 {
     public interface IBrokeredMessage : IDisposable
     {
+        int DeliveryCount { get; }
         IDictionary<string, object> Properties { get; }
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
```diff
@@ -9,6 +9,7 @@ namespace NuGet.Services.ServiceBus
 {
     public interface IBrokeredMessage : IDisposable
     {
+        int DeliveryCount { get; }
```
joelverhagen: Perhaps we can expose `EnqueuedTimeUtc` at the same time?

joelverhagen: IIRC, `EnqueuedTimeUtc` is the actual enqueue time...

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -17,10 +29,12 @@ public PackageValidationMessageData(string packageId, string packageVersion, Gui
             PackageId = packageId ?? throw new ArgumentNullException(nameof(packageId));
             PackageVersion = packageVersion ?? throw new ArgumentNullException(nameof(packageVersion));
             ValidationTrackingId = validationTrackingId;
+            DeliveryCount = deliveryCount;
         }
 
         public string PackageId { get; }
         public string PackageVersion { get; }
         public Guid ValidationTrackingId { get; }
+        public int DeliveryCount { get; }
```
joelverhagen: What is the expected behavior for in-flight messages?

joelverhagen: Do we have tests that old JSON bodies can still be deserialized?

If not, we will need to introduce `PackageValidationMessageData2` or something.

loic-sharma: The serialized message hasn't changed at all as the `DeliveryCount` property comes from the Service Bus's properties, not from the serialized message. I didn't add a test to test "old" and "new" JSON bodies are they are identical. In-flight messages should work as expected.

src/NuGet.Services.Contracts/ServiceBus/IBrokeredMessage.cs
loic-sharma: Sure

loic-sharma: Added `ExpiresAtUtc` and `EnqueuedTimeUtc` 🎉

PR #139, opened 2018-03-01
Add IProcessor interface to enable validators that mutate packages

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1190

There may be better ways to do this, but I optimized for no casting and not adding properties to an interface that may never be used (e.g. adding `NupkgUrl` to `IValidationResult`). I considered a type parameter but orchestrator still needs to fork logic on `IValidationResult` vs. `IProcessorResult`.

Frankly, I am not a big fan of this approach and wouldn't mind merging `IValidator` and `IProcessor`...

Also, I renamed `IValidator.StartValidationAsync` to `IValidator.StartAsync`. The method name seemed redundant.


joelverhagen:
This implementation looks simple enough here but it gets nasty on the orchestrator side. Basically I either need to fork a lot of logic on `IValidator` vs. `IProcessor` or I need to come up with some abstraction that handles both cases in orchestrator... which is basically just merging `IValidator` and `IProcessor` as well as `IValidationResult` and `IProcessorResult`.

I think it's worth merging the interfaces and leaving `IValidationResult.NupkgUrl` as null on validators that do not mutate packages. We'd still want an `IsProcessor` property so orchestrator can fail concurrent processors.

Thoughts?

joelverhagen:
I think this could work if, on the orchestrator, we wrap all `IValidator` to `IProcessor`, just for convenience. But then why not just do it on this side and avoid the wrapping?

loic-sharma:
After doing some thinking, I agree that the benefit of having separate `IValidator` and `IProcessor` types does not justify the complexity of the Orchestrator handling two types. We may want to merge the interfaces later in the future.

src/NuGet.Services.Contracts/Validation/IProcessor.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <returns>
+        /// A task returning the current processor result indicating that the validation and mutation flow has been
+        /// started and possibly already completed.
+        /// </returns>
+        Task<IProcessorResult> StartAsync(IValidationRequest request);
+
+        /// <summary>
+        /// A method that can be used to clean up any state produced by <see cref="StartAsync(IValidationRequest)"/>.
+        /// If this method fails for some reason, an exception should be thrown. However, the caller can simply swallow
+        /// the exception making this method a best effort. This is simply an way, in the happy path, to keep storage
```
loic-sharma: `an way` -> `a way`

src/NuGet.Services.Validation/ValidationResult.cs
```diff
@@ -1,38 +1,32 @@
...
-        private static readonly IValidationResult notStarted = new ValidationResult(ValidationStatus.NotStarted);
-        private static readonly IValidationResult incomplete = new ValidationResult(ValidationStatus.Incomplete);
-        private static readonly IValidationResult succeeded = new ValidationResult(ValidationStatus.Succeeded);
-        private static readonly IValidationResult failed = new ValidationResult(ValidationStatus.Failed);
-
         /// <summary>
         /// Represents a validation result that has not been started.
         /// </summary>
-        public static IValidationResult NotStarted => notStarted;
+        public static IProcessorResult NotStarted { get; } = new ProcessorResult(ValidationStatus.NotStarted);
```
loic-sharma: Why are these `ProcessorResult`s? Shouldn't they be `ValidationResult`s?

joelverhagen: This allows the same statics to be used by both. Given `IProcessorResult` implements `IValidationResult`, they can also be used as `IValidationResult`.

loic-sharma: Makes sense. I had the type hierarchy backwards :)

joelverhagen:
If you're curious what this design implies in orchestrator, take a look at this:
https://github.com/NuGet/NuGet.Jobs/pull/353

Not too great. I think a better approach would be just enhancing `IValidator`.

src/NuGet.Services.Validation/ValidationResult.cs
```diff
@@ -1,38 +1,32 @@
...
 {
     public class ValidationResult : IValidationResult
     {
-        private static readonly IValidationResult notStarted = new ValidationResult(ValidationStatus.NotStarted);
-        private static readonly IValidationResult incomplete = new ValidationResult(ValidationStatus.Incomplete);
-        private static readonly IValidationResult succeeded = new ValidationResult(ValidationStatus.Succeeded);
-        private static readonly IValidationResult failed = new ValidationResult(ValidationStatus.Failed);
-
         /// <summary>
         /// Represents a validation result that has not been started.
```
agr: Can we somehow call out that those statics are supposed to be used for returning the result, and not for checking it in calling code?..
I.e. that one should check `ValidationResult.ValidationStatus` and shouldn't try to compare instance of `IValidationResult` with one of `ValidationResult.NotStarted`, `ValidationResult.Incomplete`, `ValidationResult.Succeeded` and `ValidationResult.Failed`.
I would probably make them `Create*` static methods instead of static properties.

joelverhagen: Hm, I like the `Create*` idea. The implementations could return shared instances but the semantics of a method are less confusing in the way you said.

joelverhagen:
@agr, @loic-sharma, I have merged the interfaces to illustrate the different approach. I like this once better 😄.

src/NuGet.Services.Validation/ValidationResult.cs
joelverhagen: Chose not to do this since NuGet.Jobs code has never has this problem. Preferring this since it's more terse.

PR #141, opened 2018-03-15
Add ITelemetryClient and TelemetryClientWrapper

joelverhagen:
This moves the implementation from NuGetGallery to ServerCommon.

This is helpful for unit tests `Track*` methods on `TelemetryService`.


joelverhagen:
See:
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/TelemetryClientWrapper.cs

PR #143, opened 2018-03-19
Add TrackDuration extension

loic-sharma:
Add a `TrackDuration` method. This allows you to easily instrument duration of code like so:

```cs
using (_telemetryClient.TrackDuration("my_service.time_to_download_package"))
{
   // Do some expensive operation whose duration should be tracked
  _packageDownloadService.DownloadPackage("My.Package", "1.0.0");
}
```

### DIFF
--- a/tests/NuGet.Services.Logging.Tests/TelemetryClientExtensionsTests.cs
+++ b/tests/NuGet.Services.Logging.Tests/TelemetryClientExtensionsTests.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Moq;
+using Xunit;
+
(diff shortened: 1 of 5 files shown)


tests/NuGet.Services.Logging.Tests/TelemetryClientExtensionsTests.cs
```diff
@@ -0,0 +1,49 @@
...
+                using (_target.Object.TrackDuration(name, properties))
+                {
+                }
+
+                _target.Verify(t => t.TrackMetric(name, It.Is<double>(d => d > 0), properties));
+            }
+        }
+    }
+
+    public class Base
```
joelverhagen: Nest in `TelemetryClientExtensionsTests`

loic-sharma: Good catch, fixed!

chenriksson:
Awesome - reminds me of the VS MeasurementBlock instrumentation, but for AI

PR #150, opened 2018-04-25
Support for AAD SQL authentication

chenriksson:
Infrastructure for automated SQL secret rotation:
- Reading certificates from AKV
- Refreshing secrets on demand, used for SqlConnection retry
- SqlConnections that support AAD auth with token using AKV cert

Note that AAD authentication requires additional configuration, which I've tried to incorporate into the existing DB connection string argument to simplify job configuration.


src/NuGet.Services.KeyVault/ISecretInjector.cs
```diff
@@ -7,6 +7,10 @@ namespace NuGet.Services.KeyVault
 {
     public interface ISecretInjector
     {
+        string GetSecretName(string input);
+
         Task<string> InjectAsync(string input);
+
+        Task<string> InjectAsync(string input, bool forceRefresh);
```
skofman1: we can add this param to the existing method and have the default value =false

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -38,6 +38,12 @@ public KeyVaultReader(KeyVaultConfiguration configuration)
             return secret.Value;
         }
 
+        public async Task<X509Certificate2> GetCertificateSecretAsync(string secretName, string password)
```
skofman1: should we use SecureString here?

src/NuGet.Services.KeyVault/SecretInjector.cs
```diff
@@ -46,13 +57,27 @@ public SecretInjector(ISecretReader secretReader, string frame)
...
+                var secretValue = await GetSecretAsync(secretName, forceRefresh);
                 output.Replace($"{_frame}{secretName}{_frame}", secretValue);
             }
 
             return output.ToString();
         }
 
+        private async Task<string> GetSecretAsync(string secretName, bool forceRefresh = false)
+        {
+            if (forceRefresh)
```
skofman1: is there an option to separate the refresh logic from SecretInjector? The injector is meant to be very simple...

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+
+        private string Ingest(string propertyName)
+        {
+            string result = string.Empty;
+            if (ContainsKey(propertyName))
+            {
+                result = this[propertyName] as string;
+                Remove(propertyName);
```
skofman1: why remove the property?

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System.Data.Common;
+using System.Globalization;
+
+namespace NuGet.Services.Sql
+{
+    /// <summary>
+    /// Custom connection string 
+    /// </summary>
+    public class AzureSqlConnectionStringBuilder : DbConnectionStringBuilder
```
skofman1: can you add a comment with the expected format of the connection string? Great idea by the way :)

src/NuGet.Services.KeyVault/ISecretInjector.cs
chenriksson: Removed - refresh will only be called directly on secret reader

src/NuGet.Services.KeyVault/KeyVaultReader.cs
chenriksson: Removed - generated AKV certs have private key stored as AKV secret without password

src/NuGet.Services.KeyVault/SecretInjector.cs
chenriksson: Reverted, but did expose SecretReader property off of SecretInjector to allow for refresh of underlying reader.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,100 @@
...
+            SecretInjector = new SecretInjector(secretReader);
+        }
+        
+        public async Task<SqlConnection> CreateAsync()
+        {
+            try
+            {
+                return await ConnectAsync();
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
chenriksson: Changed to look for ADAL exception

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
chenriksson: Removing my custom AAD properties so that the resulting connection string is compatible with SqlConnectionStringBuilder.

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
chenriksson: Added

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+using NuGet.Services.KeyVault;
+
+namespace NuGet.Services.Sql
+{
+    public class AzureSqlConnectionFactory : ISqlConnectionFactory
+    {
+        private const string AzureSqlResourceId = "https://database.windows.net/";
+
+        public AzureSqlConnectionStringBuilder ConnectionStringBuilder { get; }
```
agr: Why do we need all those public properties? They are not part of the interface.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+                    {
+                        cachingSecretReader.RefreshSecret(certSecret);
+                    }
+                }
+            }
+        }
+
+        private string GetSecretName(string input)
+        {
+            return SecretInjector.GetSecretNames(input).SingleOrDefault();
```
agr: That's a huuuge assumption that connection string will contain a single secret and that secret is going to be cert secret...

src/NuGet.Services.KeyVault/ISecretInjector.cs
```diff
@@ -1,12 +1,17 @@
...
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace NuGet.Services.KeyVault
 {
     public interface ISecretInjector
     {
+        ISecretReader SecretReader { get; }
+
+        IEnumerable<string> GetSecretNames(string input);
```
agr: It seems what you actually need is a `ITokenizer` from this PR: https://github.com/NuGet/ServerCommon/pull/152/files/998c776443ef2dcbf547e45ee84b847d681f2eab#diff-5e4872b12eaf8af2910a866ff61c8af5 :)

Passing the injector around just to be able to extract secret names from a string sounds like overkill.

src/NuGet.Services.KeyVault/ISecretInjector.cs
```diff
@@ -1,12 +1,17 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace NuGet.Services.KeyVault
 {
     public interface ISecretInjector
     {
+        ISecretReader SecretReader { get; }
```
agr: I think it's DI's job to pass instances around, we shouldn't do that through interfaces.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        protected virtual X509Certificate2 SecretToCertificate(string rawData)
+        {
+            return new X509Certificate2(Convert.FromBase64String(rawData), string.Empty);
+        }
+
+        private void RefreshSecrets()
+        {
+            var cachingSecretReader = SecretReader as ICachingSecretReader;
```
agr: This cast is bad. It creates very non-obvious dependency.
If this class requires `ICachingSecretReader` it should be explicitly requested through constructor argument.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (cachingSecretReader != null)
+            {
+                foreach (var secret in SecretInjector.GetSecretNames(ConnectionStringBuilder.ConnectionString))
+                {
+                    cachingSecretReader.RefreshSecret(secret);
+                }
+
+                if (!string.IsNullOrEmpty(ConnectionStringBuilder.AadCertificate))
+                {
+                    var certSecret = GetSecretName(ConnectionStringBuilder.AadCertificate);
```
agr: Introduce a `GetAadCertificateSecretName` method maybe? This call is done at least twice in the class and it is a little bit non-trivial thing happening here, which I think deserves its own method.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        private void RefreshSecrets()
+        {
+            var cachingSecretReader = SecretReader as ICachingSecretReader;   
+            if (cachingSecretReader != null)
+            {
+                foreach (var secret in SecretInjector.GetSecretNames(ConnectionStringBuilder.ConnectionString))
+                {
+                    cachingSecretReader.RefreshSecret(secret);
```
agr: If updating the certificate does not immediately makes old one invalid (I hope it is, otherwise secret rotation without downtime becomes complicated), we should use existing infrastructure to do secret value updates and not add yet another one.

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private string GetSecretName(string input)
+        {
+            return SecretInjector.GetSecretNames(input).SingleOrDefault();
+        }
+
+        private static bool IsAdalException(Exception e)
+        {
+            return (e is AdalException) ? true
+                : (e.InnerException != null) ? IsAdalException(e.InnerException) : false;
```
agr: `e is AdalException || (e.InnerException != null && IsAdalException(e.InnerException))`?

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Set and remove <see cref="AzureSqlConnectionFactory"/> properties that are not supported by SqlConnectionStringBuilder.
+        /// </summary>
+        private T Ingest<T>(string propertyName, T defaultValue = default(T))
+        {
+            T result = defaultValue;
+            if (ContainsKey(propertyName))
+            {
+                var value = this[propertyName] as string;
```
agr: How does it work for `bool`?

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
chenriksson: made private

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            var connectionString = await SecretInjector.InjectAsync(ConnectionStringBuilder.ConnectionString);
+            var connection = new SqlConnection(connectionString);
+
+            if (!string.IsNullOrWhiteSpace(ConnectionStringBuilder.AadAuthority))
+            {
+                var certSecret = GetSecretName(ConnectionStringBuilder.AadCertificate);
+                if (!string.IsNullOrEmpty(certSecret))
+                {
+                    var certSecretBytes = await SecretReader.GetSecretAsync(certSecret);
```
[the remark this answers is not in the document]
chenriksson: renamed

src/NuGet.Services.Configuration/ISecretDictionary.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System.Collections.Generic;
+using NuGet.Services.KeyVault;
+
+namespace NuGet.Services.Configuration
+{
+    public interface ISecretDictionary : IDictionary<string, string>
+    {
+        ISecretInjector SecretInjector { get; }
```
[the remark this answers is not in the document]
chenriksson: reverted

src/NuGet.Services.KeyVault/ISecretInjector.cs
chenriksson: reverted

src/NuGet.Services.KeyVault/ISecretInjector.cs
chenriksson: reverted

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
chenriksson: removed

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
chenriksson: removed

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
chenriksson: removed

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
chenriksson: done

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Set and remove <see cref="AzureSqlConnectionFactory"/> properties that are not supported by SqlConnectionStringBuilder.
+        /// </summary>
+        private T Ingest<T>(string propertyName, T defaultValue = default(T))
+        {
+            T result = defaultValue;
+            if (ContainsKey(propertyName))
```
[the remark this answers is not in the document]
chenriksson: done

src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs
chenriksson: Convert

PR #151, opened 2018-04-26
Scan-related tables.

agr:
New table and indexes.


src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Validation ID for which this operation is performed.
+        /// </summary>
+        public Guid PackageValidationKey { get; set; }
+
+        /// <summary>
+        /// The type of the operation performed.
+        /// </summary>
+        public int OperationType { get; set; }
```
loic-sharma: >OperationType [](start = 19, length = 13)

What values will be in this column? Can they be represented as an enum? If not, please add documentation

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// State of the scan.
+        /// </summary>
+        public int ScanState { get; set; }
```
loic-sharma: >ScanState [](start = 19, length = 9)

What values will be in this column? Can they be represented as an enum? If not, please add documentation

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -0,0 +1,69 @@
...
+        
+        /// <summary>
+        /// The url pointing to an operation output blob.
+        /// </summary>
+        public string ResultUrl { get; set; }
+
+        /// <summary>
+        /// Third party operation ID.
+        /// </summary>
+        public string OperationId { get; set; }
```
loic-sharma: >OperationId [](start = 22, length = 11)

What is this?

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
agr: It's the internal state of the scan service it is opaque for any external observer and is not going to be detailed in any way.

src/NuGet.Services.Validation/Entities/ValidationEntitiesContext.cs
```diff
@@ -517,5 +524,74 @@ private void RegisterPackageSigningEntities(DbModelBuilder modelBuilder)
...
+                    }));
+
+            modelBuilder.Entity<ScanOperationState>()
+                .Property(s => s.CreatedAt)
+                .IsRequired()
+                .HasColumnType("datetime2")
+                .HasColumnAnnotation(
+                    IndexAnnotation.AnnotationName,
+                    new IndexAnnotation(new[] {
+                        new IndexAttribute(ScanOperationStatesScanStateCreatedIndex, 1)
```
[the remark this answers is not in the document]
agr: For monitoring. We need to be able to figure out the ages of the operations in certain states, it's better to happen without the need to do table scan.

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
agr: Updated

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
agr: This and `ScanState` are implementation details of the service that will use that table. It's not going to be exposed in a library that used in multiple projects.

PR #153, opened 2018-05-02
Certificates:  add new validation issue

dtivel:
Add a new issue for the signature validation failure when a package is signed with a certificate that is not known to the package's signing owner(s).


src/NuGet.Services.Validation.Issues/ValidationIssue.cs
```diff
@@ -17,13 +17,15 @@ public abstract class ValidationIssue : IValidationIssue
         public static ValidationIssue OnlySignatureFormatVersion1Supported { get; } = new NoDataValidationIssue(ValidationIssueCode.OnlySignatureFormatVersion1Supported);
         public static ValidationIssue AuthorCounterSignaturesNotSupported { get; } = new NoDataValidationIssue(ValidationIssueCode.AuthorCounterSignaturesNotSupported);
         public static ValidationIssue PackageIsNotSigned { get; } = new NoDataValidationIssue(ValidationIssueCode.PackageIsNotSigned);
+        public static ValidationIssue PackageIsSignedWithUnauthorizedCertificate { get; }
```
loic-sharma: >PackageIsSignedWithUnauthorizedCertificate [](start = 38, length = 42)

Out of curiosity, what will use this property?

dtivel: Good catch.  Removed.

PR #163, opened 2018-06-15
PackageNormalizedVersion

cristinamanum:
Add the PackageNormalizedVersion to the PackageValidationMessageData.
This is the first step in replacing the PackageVersion with PackageNormalizedVersion.

https://github.com/NuGet/NuGetGallery/issues/6075


src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -10,14 +11,16 @@ public class PackageValidationMessageData
...
             Guid validationTrackingId)
-          : this(packageId, packageVersion, validationTrackingId, deliveryCount: 0)
+          : this(packageId, packageVersion, packageNormalizedVersion, validationTrackingId, deliveryCount: 0)
         {
         }
 
         internal PackageValidationMessageData(
             string packageId,
             string packageVersion,
+            string packageNormalizedVersion,
```
shishirx34: no need to pass this in constructor, you can simply add the field and `Normalize` it in the constructor from `packageVersion`

cristinamanum: This is correct - I will update.

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -27,14 +28,32 @@ public class PackageValidationMessageData
             }
 
             PackageId = packageId ?? throw new ArgumentNullException(nameof(packageId));
+            // This is the first step in the transition from PackageVersion to PackageNormalizedVersion
```
chenriksson: Wouldn't persist this in comments, unless you're planning to revisit in next phase

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -27,14 +28,32 @@ public class PackageValidationMessageData
...
         public string PackageVersion { get; }
+        public string PackageNormalizedVersion { get; }
         public Guid ValidationTrackingId { get; }
         public int DeliveryCount { get; }
+
+        private static string Normalize(string version)
+        {
+            if(version == null)
+            {
+                return null;
```
chenriksson: Why not throw `ArgumentNullException`?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -27,14 +28,32 @@ public class PackageValidationMessageData
...
+            }
+            NuGetVersion parsed;
+            if (!NuGetVersion.TryParse(version, out parsed))
+            {
+                return version;
```
chenriksson: Why isn't this an exception?

src/NuGet.Services.Validation/project.json
```diff
@@ -1,7 +1,8 @@
 ﻿{
   "dependencies": {
     "EntityFramework": "6.1.3",
-    "Newtonsoft.Json": "9.0.1"
+    "Newtonsoft.Json": "9.0.1",
+    "NuGet.Versioning": "4.3.0"
```
chenriksson: Gallery.Core depends on `NuGet.Versioning, 4.8.0-preview1.5156`. Why use `4.3.0` here?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -27,14 +28,32 @@ public class PackageValidationMessageData
             }
 
             PackageId = packageId ?? throw new ArgumentNullException(nameof(packageId));
+            // This is the first step in the transition from PackageVersion to PackageNormalizedVersion
             PackageVersion = packageVersion ?? throw new ArgumentNullException(nameof(packageVersion));
```
loic-sharma: Let's create a work item to track this and then link the issue in this comment so that we don't forget about it

src/NuGet.Services.Validation/PackageValidationMessageData.cs
chenriksson: Why not throw `ArgumentNullException`?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
chenriksson: Why isn't this an exception?

src/NuGet.Services.Validation/project.json
chenriksson: `Gallery.Core` depends on `Nuget.Versioning, 4.8.9-preview1.5156`. Why only `4.3.0` here?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
loic-sharma: Maybe I'm missing something, but the Gallery will be updated to pass the normalized version in the future, right? If so, the constructor is necessary...

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: Added https://github.com/NuGet/NuGetGallery/issues/6076

loic-sharma:
>         public class TheSerializePackageValidationMessageDataMethod : Base

This needs a test for both old  and new serialized messages

---
Refers to: tests/NuGet.Services.Validation.Tests/ServiceBusMessageSerializerTests.cs:24 in f5c048a. [](commit_id = f5c048a0e9dc9526af3a9570a84cd0bc6eb40a31, deletion_comment = False)

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: Make sense. I added an work item to track this issue. It should be enough. Removed the comments.

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: Refactored and use directly NuGetVersion.Parse(packageVersion).ToNormalizedString()

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: NuGetVersion.Parse(packageVersion).ToNormalizedString() (above comments)

src/NuGet.Services.Validation/project.json
cristinamanum: Good catch - I used https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Core/packages.config but we changed to project ref. I changed it.

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: I think that I addressed this comment above

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: I think that I addressed this comment above

src/NuGet.Services.Validation/project.json
cristinamanum: I think that I addressed this comment above

PR #164, opened 2018-06-20
Service bus telemetry

agr:
Added the `ISubscriptionProcessorTelemetryService` that would allow to export two metrics from the Service Bus client:
* Message delivery lag - the time between when the message was supposed to be enqueued for delivery to the client and the actual delivery time. Larger (>~1 second) values of the metric would indicate congestion when processing service bus messages (i.e. we are receiving more messages than processing);
* Enqueue lag - the time between when the message was supposed to be enqueued for delivery and actual enqueuing time. The values greater than zero here would indicate some sort of issues on Service Bus side, so I expect this metric to be zero or near-zero 100% of the time. But if it is not, it would hopefully save us time on debugging some funny issues.

Since it's a breaking change in a library, the default implementation of the new interface was added that does nothing, but would allow quick recovery from the breaking change to the "as before" state if needed. Otherwise callers are to proper implement the interface so the metrics are actually saved.

### DIFF
--- a/src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
+++ b/src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
@@ -20,7 +20,7 @@ namespace NuGet.Services.ServiceBus
         /// scheduled delivery time set to T+S and was actually received at T+S+L, the reported
         /// value is going to be: (T+S+L) - (T+S) = L.
         /// </remarks>
-        void TrackMessageDeliveryLag(TimeSpan deliveryLag);
+        void TrackMessageDeliveryLag<TMessage>(TimeSpan deliveryLag);
 
         /// <summary>
         /// Tracks the time difference between the scheduled enqueue time and actual enqueue time.
@@ -29,6 +29,6 @@ namespace NuGet.Services.ServiceBus
         /// This time being noticeably greater than zero might indicate perf issues on Service Bus
         /// side, if we ever to see any.
         /// </remarks>
-        void TrackEnqueueLag(TimeSpan enqueueLag);
+        void TrackEnqueueLag<TMessage>(TimeSpan enqueueLag);
     }
 }
--- a/src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
+++ b/src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.ServiceBus
+{
+    /// <summary>
+    /// Interface for the telemetry emitted by <see cref="SubscriptionProcessor{TMessage}"/>.
+    /// </summary>
+    public interface ISubscriptionProcessorTelemetryService
+    {
+        /// <summary>
+        /// Tracks the time difference between the time a message was received and the time
+        /// it was scheduled to be enqueued.
+        /// </summary>
+        /// <remarks>
+        /// Scheduled enqueueing set by <see cref="IBrokeredMessage.ScheduledEnqueueTimeUtc"/> 
+        /// affects the reported time: if message was enqueued by the code at time T, had
+        /// scheduled delivery time set to T+S and was actually received at T+S+L, the reported
+        /// value is going to be: (T+S+L) - (T+S) = L.
+        /// </remarks>
+        void TrackMessageDeliveryLag(TimeSpan deliveryLag);
+
+        /// <summary>
+        /// Tracks the time difference between the scheduled enqueue time and actual enqueue time.
+        /// </summary>
+        /// <remarks>
+        /// This time being noticeably greater than zero might indicate perf issues on Service Bus
+        /// side, if we ever to see any.
+        /// </remarks>
+        void TrackEnqueueLag(TimeSpan enqueueLag);
+    }
+}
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -156,5 +162,17 @@ public void Start()
...
+        private void TrackMessageLags(IBrokeredMessage brokeredMessage)
+        {
+            try
+            {
+                _telemetryService.TrackMessageDeliveryLag(DateTimeOffset.UtcNow - brokeredMessage.ScheduledEnqueueTimeUtc);
+                // we expect the "enqueue lag" to be zero or really close to zero pretty much all the time, logging it just in case it is not
+                // and for historical perspective if we need one.
+                _telemetryService.TrackEnqueueLag(brokeredMessage.EnqueuedTimeUtc - brokeredMessage.ScheduledEnqueueTimeUtc);
+            }
+            catch { }
```
loic-sharma: We assume everywhere else that telemetry won't throw. I would just skip the try/catch

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -20,10 +19,12 @@ public class SubscriptionProcessor<TMessage> : ISubscriptionProcessor<TMessage>
         private readonly ISubscriptionClient _client;
         private readonly IBrokeredMessageSerializer<TMessage> _serializer;
         private readonly IMessageHandler<TMessage> _handler;
+        private readonly ISubscriptionProcessorTelemetryService _telemetryService;
         private readonly ILogger<SubscriptionProcessor<TMessage>> _logger;
 
         private bool _running;
         private int _numberOfMessagesInProgress;
+        private Action<TimeSpan> _messageLagCallback;
```
loic-sharma: This looks unused

src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Tracks the time difference between the time a message was received and the time
+        /// it was scheduled to be enqueued.
+        /// </summary>
+        /// <remarks>
+        /// Scheduled enqueueing set by <see cref="IBrokeredMessage.ScheduledEnqueueTimeUtc"/> 
+        /// affects the reported time: if message was enqueued by the code at time T, had
+        /// scheduled delivery time set to T+S and was actually received at T+S+L, the reported
+        /// value is going to be: (T+S+L) - (T+S) = L.
+        /// </remarks>
+        void TrackMessageDeliveryLag(TimeSpan deliveryLag);
```
loic-sharma: Could we record the `TMessage` type as a dimension? Each topic is separate, so I would expect that each subscription have independent lag

src/NuGet.Services.ServiceBus/SubscriptionProcessorNoTelemetryService.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+
+namespace NuGet.Services.ServiceBus
+{
+    /// <summary>
+    /// The implementation of the <see cref="ISubscriptionProcessorTelemetryService"/>
+    /// that does not send any telemetry.
+    /// </summary>
+    public class SubscriptionProcessorNoTelemetryService : ISubscriptionProcessorTelemetryService
```
loic-sharma: Who will use this type?

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
agr: Because that's a library and should be defensive about external code.

src/NuGet.Services.ServiceBus/SubscriptionProcessorNoTelemetryService.cs
agr: Whoever needs to make their service to work *fast* after upgrade, with planned proper implementation later.

agr: It was added because the `SubscriptionProcessor` constructor had a breaking change (required argument was added), and might require some work to implement all plumbing properly.

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
agr: Removed

src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
loic-sharma: Never mind, I misunderstood the code. This can be done by the implementation.

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -38,15 +39,18 @@ public class SubscriptionProcessor<TMessage> : ISubscriptionProcessor<TMessage>
             ISubscriptionClient client,
```
shishirx34: nit: Add the new `param` to comments above.

src/NuGet.Services.ServiceBus/ISubscriptionProcessorTelemetryService.cs
agr: Updated

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
agr: Removed

src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
agr: Added

PR #165, opened 2018-06-20
Fix Clear-PackageCache for NuGet 3.6+

scottbommarito:
See https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders
`packages-cache` became `http-cache` after 3.5


joelverhagen (approved):
Fix comment then :shipit:

PR #166, opened 2018-06-27
Add ValidatingEntityType

cristinamanum:
Add ValidatingEntityType to the Package Validation Set and add it to the ValidationMessage Data.
This will allow:
- The ValidationMessage Handlers can choose to reject any validation types that they do not expect.
- The Package Validation Sets can be filtered only for the validation sets specific for a type.


src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -8,18 +8,28 @@ namespace NuGet.Services.Validation
 {
     public class PackageValidationMessageData
```
loic-sharma: >PackageValidationMessageData [](start = 17, length = 28)

Should we rename this to `ValidationMessageData`?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -8,18 +8,28 @@ namespace NuGet.Services.Validation
...
     {
+        public PackageValidationMessageData(
```
loic-sharma: >PackageValidationMessageData [](start = 15, length = 28)

Should we make the `validatingEntityType` parameter required?

src/NuGet.Services.Validation/Entities/PackageValidationSet.cs
```diff
@@ -71,5 +71,11 @@ public class PackageValidationSet
         /// The package validations in the validation set.
         /// </summary>
         public virtual ICollection<PackageValidation> PackageValidations { get; set; }
+
+        /// <summary>
+        /// The entity type to be validated.
+        /// </summary>
+        public ValidatingEntityType ValidatingEntityType { get; set; }
```
loic-sharma: >ValidatingEntityType [](start = 36, length = 20)

What will the Orchestrator's query patterns be like? Do we need any indexes?

Also, I think a shorter property name could be fine here - what do you think of `Type`?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: I think that if will be more cases when this needs to be used for other types is worth the work of renaming. Symbol is very close to the package structure, If not strong objection I will keep like it is.

src/NuGet.Services.Validation/PackageValidationMessageData.cs
cristinamanum: I think that if the name type nae is PacakgeValidationMessageData having the default to package makes sense.  If not strong objection I will keep like it is.

src/NuGet.Services.Validation/Entities/PackageValidationSet.cs
cristinamanum: Good point for the indexes.I will look into this. I shortened the name to ValidatingType .

PR #168, opened 2018-07-05
[Telemetry] Add duration metric with properties

loic-sharma:
Enable metrics that use this pattern:

```csharp
using (var operation = _telemetry.TrackMyOperation())
{
  // Do complicated work that may fail...

  operation.Properties.Success = true;
  operation.Properties.SomeValueNotKnownUntilEndOfOperation = "Hello world";
}
```

This will be used by the revalidation job to track the result of each revalidation it starts.


agr:
Need a default implementation for `serializeFunc`?

loic-sharma:
@agr I'd like to, but building a fast object to dictionary converter is difficult

PR #169, opened 2018-07-06
Add the Killswitch client

loic-sharma:
Killswitches allow you to turn off services' features remotely. This will be used to turn off the revalidation job if it causes pressure on the ingestion pipeline. The API is simple:

```csharp
if (!_killswitchClient.IsActive("MyService.MyFeature"))
{
  _logger.LogWarning("My feature is disabled, don't do anything!");
  return;
}
```

Like the [`ITelemetryClient`](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs), services are expected to wrap the `IKillswitchClient` with higher-level abstractions. For example:

```csharp
if (!_killswitchService.ShouldEnqueueRevalidation())
{
  _logger.LogWarning("Revalidations have been disabled!");
  return;
}
```

# Killswitch Endpoint

Killswitches work by polling an endpoint that returns a JSON list of activated killswitches:

```json
[
  "Revalidation.Enqueue",
  "Gallery.SymbolUploads"
  "Orchestrator.RepositorySigning"
]
```

This endpoint will be a JSON file hosted on Azure Blob Storage. An admin panel will be created on the Gallery to modify this file.

# Design Decisions

## Client uses HTTP endpoint instead of Azure Blob Storage

The fact that we use blob storage on the backend is an implementation detail. With blob storage, we'd need to add storage credentials to all services, which would increase the cost to on-board. As for privacy concerns, we could make the killswitch blob private but generate a URI with read-only access for services to consume.

Lastly, an HTTP implementation can support a push model through long-polling. As far as I know, this cannot be done using Azure Blob Storage.

## Background Refresh

In this implementation, `IsActive` does not refresh the killswitch cache. Instead, a background task refreshes the cache periodically (every 5 minutes by default). This lets consumers of the killswitch client call `IsActive` as frequently as they'd like, including in hot paths. 

# Compared to Gallery's [`ContentObjectService`](https://github.com/NuGet/NuGetGallery/blob/4622ebed4e10a25987adc41cc94fe6dab9d57508/src/NuGetGallery/Services/ContentObjectService.cs)

The `HttpKillswitchClient` is very similar to the `ContentObjectService`:

* Both refresh their state in the background
* Both store their state as JSON on an endpoint
* Both allow fast lookups of the currently cached values

The differences are:

* The `ContentObjectService` gives you high-level objects, allowing you to build complex feature flag logic. The `HttpKillswitchClient` can only determine whether a features should be on or off
* To add a new feature, the `ContentObjectService` requires code changes in both the Gallery as well as the service that owns the feature
* The `ContentObjectService` uses APIs specific to ASP.NET for background refreshes, and cannot be used by jobs (see [this](https://github.com/NuGet/NuGetGallery/blob/4622ebed4e10a25987adc41cc94fe6dab9d57508/src/NuGetGallery/App_Start/OwinStartup.cs#L71))
* The `ContentObjectService` fails silently if the latest settings cannot be fetched/parsed
* The `ContentObjectService` depends on Azure Blob Storage

Part of https://github.com/NuGet/Engineering/issues/1440

### DIFF
--- a/tests/NuGet.Services.Killswitch.Tests/NuGet.Services.Killswitch.Tests.csproj
+++ b/tests/NuGet.Services.Killswitch.Tests/NuGet.Services.Killswitch.Tests.csproj
@@ -1,0 +1,59 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{4F73955E-3A98-4403-8BBE-43271FED1CA3}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>NuGet.Services.Killswitch.Tests</RootNamespace>
+    <AssemblyName>NuGet.Services.Killswitch.Tests</AssemblyName>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <TargetFrameworkProfile />
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="HttpKillswitchClientFacts.cs" />
+    <Compile Include="Support\KillswitchIntegrationTestCollection.cs" />
+    <Compile Include="Support\KillswitchIntegrationTestFixture.cs" />
+    <Compile Include="Support\KillswitchTestServer.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\NuGet.Services.KillSwitch\NuGet.Services.KillSwitch.csproj">
+      <Project>{bc8d8216-8a19-4402-9da3-238d0162b0a6}</Project>
+      <Name>NuGet.Services.KillSwitch</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="project.json" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+</Project>
--- a/NuGet.Server.Common.sln
+++ b/NuGet.Server.Common.sln
@@ -62,6 +62,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.Sql", "src\N
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.Sql.Tests", "tests\NuGet.Services.Sql.Tests\NuGet.Services.Sql.Tests.csproj", "{FA2B3447-7242-495F-A8E1-D94181C0C9A5}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.KillSwitch", "src\NuGet.Services.KillSwitch\NuGet.Services.KillSwitch.csproj", "{BC8D8216-8A19-4402-9DA3-238D0162B0A6}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Services.Killswitch.Tests", "tests\NuGet.Services.Killswitch.Tests\NuGet.Services.Killswitch.Tests.csproj", "{4F73955E-3A98-4403-8BBE-43271FED1CA3}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -168,6 +172,14 @@ Global
 		{FA2B3447-7242-495F-A8E1-D94181C0C9A5}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{FA2B3447-7242-495F-A8E1-D94181C0C9A5}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{FA2B3447-7242-495F-A8E1-D94181C0C9A5}.Release|Any CPU.Build.0 = Release|Any CPU
+		{BC8D8216-8A19-4402-9DA3-238D0162B0A6}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{BC8D8216-8A19-4402-9DA3-238D0162B0A6}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{BC8D8216-8A19-4402-9DA3-238D0162B0A6}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{BC8D8216-8A19-4402-9DA3-238D0162B0A6}.Release|Any CPU.Build.0 = Release|Any CPU
+		{4F73955E-3A98-4403-8BBE-43271FED1CA3}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{4F73955E-3A98-4403-8BBE-43271FED1CA3}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{4F73955E-3A98-4403-8BBE-43271FED1CA3}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{4F73955E-3A98-4403-8BBE-43271FED1CA3}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -198,6 +210,8 @@ Global
 		{E29F54DF-DFB8-4E27-940D-21ECCB9B6FC1} = {7783A106-0F4C-4055-9AB4-413FB2C7B8F0}
 		{F5121B0A-669F-48BD-86DC-27C546D1A825} = {8415FED7-1BED-4227-8B4F-BB7C24E041CD}
 		{FA2B3447-7242-495F-A8E1-D94181C0C9A5} = {7783A106-0F4C-4055-9AB4-413FB2C7B8F0}
+		{BC8D8216-8A19-4402-9DA3-238D0162B0A6} = {8415FED7-1BED-4227-8B4F-BB7C24E041CD}
+		{4F73955E-3A98-4403-8BBE-43271FED1CA3} = {7783A106-0F4C-4055-9AB4-413FB2C7B8F0}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {AA413DB0-5475-4B5D-A3AF-6323DA8D538B}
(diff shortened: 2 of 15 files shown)


src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+            _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _active = new HashSet<string>();
+        }
+
+        public void Start(CancellationToken cancellationToken)
+        {
+            Task.Run(async () =>
```
loic-sharma: TODO: Make this async, call `RefreshAsync` once before returning to ensure cache is primed.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+            var content = await response.Content.ReadAsStringAsync();
+
+            var activeNames = JsonConvert.DeserializeObject<List<string>>(content);
+
+            _active = new HashSet<string>(activeNames, StringComparer.OrdinalIgnoreCase);
+
+            _logger.LogInformation("Refreshed killswitches");
+        }
+
+        public bool IsActive(string name) => _active.Contains(name);
```
loic-sharma: TODO: Throw if not started.

tests/NuGet.Services.Killswitch.Tests/HttpKillswitchServiceFacts.cs
```diff
@@ -0,0 +1,79 @@
...
+        public HttpKillswitchServiceFacts()
+        {
+            _httpClient = new HttpClient();
+            _testServer = new Lazy<Task<KillswitchTestServer>>(KillswitchTestServer.CreateAsync);
+        }
+
+        public Task<KillswitchTestServer> GetTestServerAsync() => _testServer.Value;
+
+        [Fact]
+        public async Task ReturnsTrueForActiveKillswitches()
```
loic-sharma: TODO: Test malformed responses, that `StartAsync` refreshes in background once, that `IsActive` throws if not started

tests/NuGet.Services.Killswitch.Tests/KillswitchTestServer.cs
```diff
@@ -0,0 +1,118 @@
...
+                {
+                    if (ex.ErrorCode == ErrorConstants.ERROR_OPERATION_ABORTED ||
+                        ex.ErrorCode == ErrorConstants.ERROR_INVALID_HANDLE ||
+                        ex.ErrorCode == ErrorConstants.ERROR_INVALID_FUNCTION ||
+                        RuntimeEnvironmentHelper.IsMono && ex.ErrorCode == ErrorConstants.ERROR_OPERATION_ABORTED_MONO)
+                    {
+                        return;
+                    }
+
+                    Console.WriteLine($"Unexpected error code:  {ex.ErrorCode}.  Exception:  {ex.ToString()}");
```
loic-sharma: `ex.ToString()` -> `ex`

tests/NuGet.Services.Killswitch.Tests/NuGet.Services.Killswitch.Tests.csproj
```diff
@@ -0,0 +1,76 @@
...
+    </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+    <Reference Include="xunit.abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c, processorArchitecture=MSIL">
```
loic-sharma: Get rid of packages.config references.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+        public void Start(CancellationToken cancellationToken)
+        {
+            Task.Run(async () =>
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    try
+                    {
+                        await RefreshAsync();
+                        await Task.Delay(_config.RefreshInterval);
```
skofman1: what's the intended RefreshInterval? How often will IsActive be called?
If RefreshInterval is small and IsActive is not frequent this approach is inefficient. A different approach is to refresh when IsActive is called.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+            ILogger<HttpKillswitchService> logger)
+        {
+            _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            _config = config ?? throw new ArgumentNullException(nameof(config));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _active = new HashSet<string>();
+        }
+
+        public void Start(CancellationToken cancellationToken)
```
skofman1: This is not thread safe. Start() can be called multiple times.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+            Task.Run(async () =>
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    try
+                    {
+                        await RefreshAsync();
+                        await Task.Delay(_config.RefreshInterval);
+                    }
+                    catch (Exception e)
```
skofman1: You are swallowing the exception and hiding issues. The correct behavior will be to add retries to protect against network issues, but throw on deserialization issues and other unrecoverable exceptions.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+                        _logger.LogError(0, e, "Failed to refresh killswitches due to exception");
+                    }
+                }
+
+                _logger.LogInformation("Cancelling killswitch refreshes");
+            },
+            cancellationToken);
+        }
+
+        public async Task RefreshAsync()
```
skofman1: why public if not part of the interface?

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+    public class HttpKillswitchService : IKillswitchService
+    {
+        private readonly HttpClient _httpClient;
+        private readonly HttpKillswitchConfig _config;
+        private readonly ILogger<HttpKillswitchService> _logger;
+
+        private HashSet<string> _active;
+
+        public HttpKillswitchService(
+            HttpClient httpClient,
```
skofman1: using HttpClient you assume the config blob is public, but it's an internal service config and should be private. Use IStorage interface.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace NuGet.Services.KillSwitch
+{
+    /// <summary>
+    /// A killswitch service that queries an HTTP service periodically for the list of activate killswitches.
+    /// </summary>
+    public class HttpKillswitchService : IKillswitchService
```
skofman1: There will be two roles in the killswitch world: reader and writer. If you intend to separate them into two interfaces you should suffix this interface with Reader. You could also combine the two into a single KillSwitchService, however, since there's going to be only one writer (the gallery) I don't think it's a good idea.

skofman1:
I like the idea of having a system wide killswitch service, however this creates feature duplication in the gallery with the "feature flag" service. Both turn abilities on/off. We should discuss this on a team level.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace NuGet.Services.KillSwitch
+{
+    /// <summary>
+    /// A killswitch service that queries an HTTP service periodically for the list of activate killswitches.
```
shishirx34: nit: active killswitches

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: It may be useful for downstream consumers. Its not on the interface as not all killswitch services that support push model don't need a concept of refresh.

It also enables easy testing.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: This runs on a background thread, throwing would cause the background thread to end.

That being said, I agree that we should have some "hard fail" behavior when we repeatedly fail to reload the killswitches. I'll make `IsActive` throw if the cache's staleness reaches some threshold.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
shishirx34: I am assuming `IsActive` will be called very frequently, specifically for revalidation job.. this could get out of hand. A decent `RefreshInterval` should work fine. It could be tweaked as per the calling job via config. Or may be add another config to determine when to refresh? Either via interval or with `IsActive` call?

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Why isn't this thread-safe? As far as I can tell, it is. Reference assignment is atomic, and that's the only thing that the background thread interacts with.

Calling `Start` multiple times would cause the cache to be refreshed at a faster cadence. I can add a check to `Start` to ensure it's called just once, but I personally prefer simpler code.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Each job can have its own refresh cadence. I'm thinking 5 minutes for the revalidate job.

The idea is to make `IsActive` as cheap as possible so that it can be called *anywhere* including critical and hot paths. Moving the network request into `IsActive` would limit where we can call `IsActive`.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
skofman1: >> Calling Start multiple times would cause the cache to be refreshed at a faster cadence. I can add a check to Start to ensure it's called just once, but I personally prefer simpler code.

This is the reason. The code won't work as expected if Start() is called multiple times.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: I talked to @shishirx34, I will add `RefreshAsync` to the interface.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Hm, good point. What do you think of `HttpKillswitchClient` so that it's more inline with our naming in logging and service bus?

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: I considered this but decided against it. With blob storage, we'd need to add storage credentials to all jobs, which makes it harder to on-board a new service to killswitches.  I don't think there's a need for killswitches to be private and there's precedent for public blobs: search auxiliary data, cursors, etc.. Really, the fact that we will use blob storage on the backend is an implementation detail.

Lastly, an HTTP implementation can support for a push model through long-polling. This requires a backend that supports long polling.

agr: We can have a SAS URL to a non-public blob with expiration 100 years in the future, can't we?

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Added a check to ensure `Start` is called just once.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Made `IsActive` throw if the cache reaches the configured staleness.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Fixed

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: @agr Probably. That doesn't really fix any of the issues I brought up in the previous comment though.

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Updated

tests/NuGet.Services.Killswitch.Tests/NuGet.Services.Killswitch.Tests.csproj
loic-sharma: Fixed

tests/NuGet.Services.Killswitch.Tests/KillswitchTestServer.cs
loic-sharma: Fixed

tests/NuGet.Services.Killswitch.Tests/HttpKillswitchServiceFacts.cs
loic-sharma: Fixed

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Fixed

src/NuGet.Services.KillSwitch/HttpKillswitchService.cs
loic-sharma: Fixed

loic-sharma:
@skofman1 Sure. Let's talk about it after the next standup?

loic-sharma:
We would like to implement this using storage APIs. Once services use the same APIs, we can look into a universal killswitch system

PR #171, opened 2018-07-10
Adding the ability to specify MaxConcurrentCalls for the SubscriptionProcessor

agr:
Previously we used whatever default value ServiceBus library provided (which seem to be 1). Now, we'd be able to specify whatever we want.
Previous behavior is kept: `Start()` without arguments still uses the default library setting. New behavior is opt-in by means of `Start(int)` method.

### DIFF
--- a/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
+++ b/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
@@ -56,12 +56,29 @@ namespace NuGet.Services.ServiceBus
         {
             _logger.LogInformation("Registering the handler to begin listening to the Service Bus subscription");
 
-            _running = true;
-
-            _client.OnMessageAsync(OnMessageAsync, new OnMessageOptionsWrapper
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -56,14 +56,31 @@ public void Start()
...
-            _running = true;
+            StartInternal(new OnMessageOptionsWrapper
+            {
+                AutoComplete = false,
+            });
+        }
+
+        public void Start(int maxConcurrentCalls)
+        {
+            _logger.LogInformation("Registering the handler to begin listening to the Service Bus subscription with maxConcurrentCalls = {MaxConcurrentCalls}",
```
loic-sharma: Shoudl we maybe move this log into `StartInternal` and log the entire `OnMessageOptionsWrapper` object?

agr: Done.

PR #172, opened 2018-07-16
Replace default ADAL TokenCache

chenriksson:
**Changes:**
- Update to not use default ADAL TokenCache. see: [here](https://github.com/NuGet/ServerCommon/compare/master...chenriks-aadsql#diff-5c308c324b0f35bebf412e663485b68cR54)
- Cache ADAL authentication objects statically per ConnectionString
  - Authenticators cache is refreshed when cert data is updated (i.e., CachingSecretReader)
- Cache ADAL access token on authenticator
  - Access token refreshed when <30min remaining on lifetime (lifetime is 60m, so requests new one every 30m)

**Notes:**
- Default [TokenCache](https://docs.microsoft.com/en-us/azure/architecture/multitenant-identity/token-cache) is not thread-safe, which is why we need to replace it

- I believe (internal) SQL connection pooling is done based on connection string + identity (token in our case)

- Unlike password authentication, token-based AAD authentication has expiration which is configurable at the AD instance. see: [AD token lifetimes](https://docs.microsoft.com/en-us/azure/active-directory/active-directory-configurable-token-lifetimes)
  - Our access token lifetime is 60m, same as default

- ADAL v3 automatically uses refresh tokens behind the scenes, extending the lifetime of the token-based SqlConnections. This is also configurable at the AD instance, but I wasn't able to determine defaults / policies for these.
  - Refresh token lifetime
  - Refresh token inactive time

- In testing, I saw a single token-based SqlConnection (expiration = 1h) remain open for 4.5h while doing a simple query every 30s. Connection lifetime was shorter when I started increasing the inactive time between queries... but didn't capture enough data to determine policies / behavior here.

**Rollout:**
- NuGetGallery - SqlConnection per EF context, which is per request lifetime. OK to proceed since lifetime is short, and default TokenCache is now removed.
- Non-validation jobs - Usually 1 SqlConnection per query. Should be OK to proceed since queries are generally short, and default TokenCache is now removed. Longest query times I could determine from AI:
  - Stats.AggregateCdnDownloadsInGallery: 15m for Stats query, <60m for Gallery update query
  - Stats.CreateAzureCdnWarehouseReports: 15m for DownloadCountReport
- Validation jobs - I need to do more testing to verify EF context lifetimes are short-lived.


src/NuGet.Services.Sql/AzureSqlClientAuthenticator.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        private AuthenticationContext AuthenticationContext { get; }
+
+        private IClientAssertionCertificate ClientAssertionCertificate { get; }
+
+        /// <summary>
+        /// Cached AAD access token, because acquiring the access token takes an external request which can
+        /// add 2-3s latency to the connection.
+        /// </summary>
+        protected IAuthenticationResult AuthenticationResult { get; set; }
```
loic-sharma: >protected [](start = 8, length = 9)

What's the reason for `protected` here? A class could extend `AzureSqlClientAuthenticator` and set `AuthenticationResult` without acquiring the `AuthenticationResultLock`.

chenriksson:
@loic-sharma Updated since our discussion... can you review? /cc @scottbommarito 

It now acquires token on current thread only for the initial request(s), or if cached value is already expired. Otherwise, it refreshes in the background when access token `ExpiresOn` is < 30m, or if the AKV cert has been rotated. I didn't bother locking around the token acquisition, assuming last one in can win... let me know if you have concerns w/ this.

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+
+namespace NuGet.Services.Sql
+{
+    public class AccessTokenCache
```
scottbommarito: nit: This is intended to be used exclusively by `AzureSqlConnectionFactory`, and not by anybody who is using this library, correct? I would make the class internal to avoid confusion.

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+
+            throw new InvalidOperationException($"Failed to acquire access token for {connectionString.Sql.InitialCatalog}.");
+        }
+
+        /// <summary>
+        /// Refresh the access token for a specific connection string, on a background thread.
+        /// </summary>
+        private void TriggerBackgroundRefresh(AzureSqlConnectionStringBuilder connectionString, string clientCertificateData, ILogger logger = null)
```
loic-sharma: >null [](start = 143, length = 4)

It doesn't seem like the default value for the logger is necessary

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// Refresh the access token for a specific connection string, on the current thread.
+        /// </summary>
+        private async Task<bool> TryRefreshAccessTokenAsync(
+            AzureSqlConnectionStringBuilder connectionString,
+            string clientCertificateData,
+            ILogger logger = null)
+        {
+            try
+            {
+                using (logger?.BeginScope($"Refreshing access token for {connectionString.Sql.InitialCatalog}."))
```
loic-sharma: >$ [](start = 42, length = 1)

You should pass the log's parameters as parameters instead of formatting the string yourself:

1. Log implementations may do magic using the parameters (for example, AI shoves log parameters as custom dimensions, thereby enabling queries)
2. This forces string formatting to happen at log time, which may be expensive (for example, turning an enum value to string)

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    var accessToken = await AcquireAccessTokenAsync(connectionString, clientCertificateData);
+                    if (accessToken != null)
+                    {
+                        _cache.AddOrUpdate(connectionString.ConnectionString, accessToken, (k, v) => accessToken);
```
loic-sharma: >AddOrUpdate [](start = 31, length = 11)

I would consider adding a log line when the token has been refreshed so that we can determine how long the refresh took.

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+            out AccessTokenCacheValue accessToken)
+        {
+            return _cache.TryGetValue(connectionString.ConnectionString, out accessToken);
+        }
+
+        protected virtual async Task<AccessTokenCacheValue> AcquireAccessTokenAsync(
+            AzureSqlConnectionStringBuilder connectionString,
+            string clientCertificateData)
+        {
+            var certificate = new X509Certificate2(Convert.FromBase64String(clientCertificateData), string.Empty);
```
loic-sharma: >X509Certificate2 [](start = 34, length = 16)

Ideally, this certificate should be disposed. It looks like the current implementation leaks whenever a token is replaced. Feel free to ignore this if you think it'll complicate the code too much.

src/NuGet.Services.Sql/AccessTokenCache.cs
agr: Also, with the current Serilog version that we use, the Scope message is not saved anywhere.

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,145 @@
...
+            ILogger logger = null)
+        {
+            AccessTokenCacheValue accessToken;
+
+            if (TryGetValue(connectionString, out accessToken) && !IsExpired(accessToken))
+            {
+                // Refresh access token in background, if near expiry or client certificate has changed.
+                if (IsNearExpiry(accessToken) || ClientCertificateHasChanged(accessToken, clientCertificateData))
+                {
+                    TriggerBackgroundRefresh(connectionString, clientCertificateData, logger);
```
loic-sharma: >TriggerBackgroundRefresh [](start = 20, length = 24)

This will kick off X background tasks if `GetAsync` is called X times while `IsNearExpiry` is true but before any of the `TriggerBackgroundRefresh` finish. This could get out of hand if it takes a long time to refresh the token (say seconds).

I would consider enforcing a single active `TriggerBackgroundRefresh` at a time. I would also make `TryRefreshAccessTokenAsync` block asynchronously if another thread is refreshing the token, with some timeout.

chenriksson: @loic-sharma see what you think... I decided to add a lock w/ timeout on the refresh. BG threads would wait max of 250ms on the lock, FG threads will block for 5s max.

Also, I'm seeing the token request take ~300-500ms in testing.

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,195 @@
...
+        /// <summary>
+        /// Access token is near expiration, and should be refreshed soon.
+        /// </summary>
+        private bool IsNearExpiry(AccessTokenCacheValue accessToken)
+        {
+            return TokenExpiresIn(accessToken.AuthenticationResult, DefaultMinExpirationInMinutes);
+        }
+        
+        /// <summary>
+        /// AKV client certificate has been rotated, and client assertion and tokens should be refreshed soon. Old certificate
```
loic-sharma: >AKV [](start = 12, length = 3)

I'm guessing AKV stands for Azure Key Vault? I would expand the acronym to avoid confusion

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,195 @@
...
+            ILogger logger)
+        {
+            try
+            {
+                var start = DateTimeOffset.Now;
+                var accessToken = await AcquireAccessTokenAsync(connectionString, clientCertificateData);
+
+                Debug.Assert(accessToken != null);
+
+                logger?.LogInformation("Refreshed access token for {initialCatalog} in {elapsedMilliseconds}.",
```
loic-sharma: >initialCatalog [](start = 68, length = 14)

We've been using upper case parameters for log templates, like `{InitialCatalog}`

src/NuGet.Services.Sql/AccessTokenCache.cs
```diff
@@ -0,0 +1,195 @@
...
+                    connectionString.Sql.InitialCatalog,
+                    (DateTimeOffset.Now - start).TotalMilliseconds);
+
+                _cache.AddOrUpdate(connectionString.ConnectionString, accessToken, (k, v) => accessToken);
+
+                return true;
+            }
+            catch (Exception ex)
+            {
+                logger?.LogError("Failed to refresh access token for {initialCatalog}: {error}",
```
loic-sharma: >LogError [](start = 24, length = 8)

There's a LogError that accepts the exception directly. I think its `LogError(ex, 0, "Failed to refresh access token {InitialCatalog}");`

tests/NuGet.Services.Status.Tests/NuGet.Services.Status.Tests.csproj
```diff
@@ -67,5 +67,8 @@
       <Version>2.2.0</Version>
     </PackageReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
chenriksson: Ick... see https://github.com/Microsoft/vstest/issues/472

Going to leave in, since this already exists in other projects. We can clean up once the fix has propagated.

PR #173, opened 2018-07-17
Add RebootRoleInstanceAsync to allow programmatic rebooting of cloud service instance

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1527.

Make long lines more readable and used string interpolation instead of `string.Format`.

### DIFF
--- a/src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
+++ b/src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
@@ -41,7 +41,14 @@ namespace NuGet.Services.AzureManagement
             _clientCredential = new ClientCredential(configuration.ClientId, configuration.ClientSecret);
         }
 
-        public async Task<string> GetCloudServicePropertiesAsync(string subscription, string resourceGroup, string name, string slot, CancellationToken token)
+        public async Task RebootCloudServiceRoleInstanceAsync(
+            string subscription,
+            string resourceGroup,
+            string name,
+            string slot,
+            string role,
+            string roleInstance,
+            CancellationToken token)
         {
             if (string.IsNullOrEmpty(subscription))
             {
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
```diff
@@ -41,7 +41,14 @@ public AzureManagementAPIWrapper(IAzureManagementAPIWrapperConfiguration configu
             _clientCredential = new ClientCredential(configuration.ClientId, configuration.ClientSecret);
         }
 
-        public async Task<string> GetCloudServicePropertiesAsync(string subscription, string resourceGroup, string name, string slot, CancellationToken token)
+        public async Task RebootRoleInstanceAsync(
```
loic-sharma: >RebootRoleInstanceAsync [](start = 26, length = 23)

This API is cloud service specific, right? Should "CloudService" be in this method name? We may want to add rebooting of App Services later

joelverhagen: Sure, good idea.

joelverhagen: Fixed.

src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
```diff
@@ -63,14 +70,71 @@ public AzureManagementAPIWrapper(IAzureManagementAPIWrapperConfiguration configu
...
+            }
 
-            string requestUrl = string.Format(RequestUrlFormat, subscription, resourceGroup, name, slot);
+            if (string.IsNullOrEmpty(roleInstance))
+            {
+                throw new ArgumentException(nameof(roleInstance));
+            }
 
-            return await MakeAzureRequest(requestUrl, token);
+            var requestUrl = "https://management.azure.com" +
```
agr: Hostname should at least be a constant. Ideally it should be passed from outside (to a constructor?).

PR #174, opened 2018-07-19
SymbolErrorCodes

cristinamanum:
Add symbol error codes.

### DIFF
--- a/src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
+++ b/src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
@@ -63,6 +63,16 @@ namespace NuGet.Services.Validation
         PackageIsSignedWithUnauthorizedCertificate = 9,
 
         /// <summary>
+        /// Symbol checksum does not match with the binary assembly.
+        /// </summary>
+        ChecksumDoesNotMatch = 200,
+
+        /// <summary>
+        /// The pdb is not portable or did not have a binary assembly file associated.
+        /// </summary>
+        MatchingPortablePDBNotFound = 201,
+
+        /// <summary>
         /// Obsolete testing issue - do NOT use this!
         /// </summary>
         [Obsolete("This issue code should only be used for testing")]
--- a/src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
+++ b/src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
@@ -62,15 +62,17 @@ namespace NuGet.Services.Validation
         /// </summary>
         PackageIsSignedWithUnauthorizedCertificate = 9,
 
+        #region SymbolErrorCodes - reserved 200 - 299 segment
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
```diff
@@ -62,6 +62,16 @@ public enum ValidationIssueCode
         /// </summary>
         PackageIsSignedWithUnauthorizedCertificate = 9,
 
+        /// <summary>
```
chenriksson: Maybe worth putting a region or comments in to separate nupkg and snupkg issue codes?

Are nupkg issue codes always relevant for snupkg, and do snupkg codes always start at 200? If so, maybe it would be more user/ops friendly to use similar codes for the common scenarios...

e.g.:
```
  Unknown = 0,
  PackageIsZip64 = 3,
  OnlyAuthorSignaturesSupported = 4,
  ...
  SymbolsUnknown = 200,
  SymbolsPackageIsZip64 = 203
  SymbolsOnlyAuthorSignaturesSupported = 204,
  ...
  SymbolsCheckSumDoesNotMatch = 250,
  ...
```

src/NuGet.Services.Validation.Issues/ValidationIssue.cs
```diff
@@ -17,6 +17,8 @@ public abstract class ValidationIssue : IValidationIssue
         public static ValidationIssue OnlySignatureFormatVersion1Supported { get; } = new NoDataValidationIssue(ValidationIssueCode.OnlySignatureFormatVersion1Supported);
         public static ValidationIssue AuthorCounterSignaturesNotSupported { get; } = new NoDataValidationIssue(ValidationIssueCode.AuthorCounterSignaturesNotSupported);
         public static ValidationIssue PackageIsNotSigned { get; } = new NoDataValidationIssue(ValidationIssueCode.PackageIsNotSigned);
+        public static ValidationIssue ChecksumDoesNotMatch { get; } = new NoDataValidationIssue(ValidationIssueCode.ChecksumDoesNotMatch);
+        public static ValidationIssue MatchingPortablePDBNotFound { get; } = new NoDataValidationIssue(ValidationIssueCode.MatchingPortablePDBNotFound);
```
chenriksson: Due to shared code, do Symbols issues need to be combined in same class like this?
Can nupkg / snupkg issues at least be separated via region, comments, and/or naming?

src/NuGet.Services.Contracts/Validation/ValidationIssueCode.cs
cristinamanum: I like this idea. Reserved the first 50 segment for the common scenarios.

src/NuGet.Services.Validation.Issues/ValidationIssue.cs
cristinamanum: I separated by naming .

PR #175, opened 2018-07-23
Add incident API client classes to shared location

scottbommarito:
This is currently used only by the status aggregator, but I decided to move it here in case we ever need to access it elsewhere (e.g. in code that creates incidents).


NuGet.Services.Incidents/Incident.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace NuGet.Services.Incidents
+{
+    public class Incident
+    {
+        public string Id { get; set; }
+        public int Severity { get; set; }
+        public IncidentStatus Status { get; set; }
+        public DateTime CreateDate { get; set; }
```
loic-sharma: >DateTime [](start = 15, length = 8)

Can this be a DateTimeOffset?

NuGet.Services.Incidents/IncidentStatus.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NuGet.Services.Incidents
+{
+    public enum IncidentStatus
+    {
+        Holding,
```
loic-sharma: >Holding [](start = 8, length = 7)

Do we need all these statuses? If so, can we comment them? Its unclear what `Holding` or `Correlating` means.

agr:
Some tests maybe?

scottbommarito:
@agr I was going to add tests, but ultimately I feel the behavior here is so simple it's not really necessary.

PR #179, opened 2018-07-25
Add VSTS ServerRequest entity

cristinamanum:
Add VSTS ServerRequest entity


src/NuGet.Services.Validation/Entities/SymbolsServerRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// The key of the symbols package.
+        /// </summary>
+        public int SymbolsKey { get; set; }
+
+        /// <summary>
+        /// The request name used for ingestion.
+        /// </summary>
+        public string RequestName { get; set; }
```
chenriksson: We provide the VSTS name? How will we set it?

src/NuGet.Services.Validation/Migrations/201807252343358_AddSymbolsServerRequest.cs
```diff
@@ -0,0 +1,30 @@
...
+    public partial class AddSymbolsServerRequest : DbMigration
+    {
+        public override void Up()
+        {
+            CreateTable(
+                "dbo.SymbolsServerRequests",
+                c => new
+                    {
+                        SymbolsKey = c.Int(nullable: false, identity: true),
+                        RequestName = c.String(),
```
chenriksson: Should this be `nullable: false`?

src/NuGet.Services.Validation/Entities/SymbolsServerRequest.cs
cristinamanum: The ingestor job will set it.

src/NuGet.Services.Validation/Migrations/201807252343358_AddSymbolsServerRequest.cs
cristinamanum: Yes - correct . Updated

PR #180, opened 2018-07-28
Drop the previous migration and add a new one in order to drop the id…

cristinamanum:
Drop the previous migration and add a new one in order to drop the identity on the SymbolsKey column.


src/NuGet.Services.Validation/Entities/ValidationEntitiesContext.cs
```diff
@@ -675,6 +675,9 @@ private void RegisterSymbolEntities(DbModelBuilder modelBuilder)
         {
             modelBuilder.Entity<SymbolsServerRequest>()
                 .HasKey(r => r.SymbolsKey);
+            modelBuilder.Entity<SymbolsServerRequest>()
+              .Property(r => r.SymbolsKey)
```
loic-sharma: It looks like lines 679, 780, 683, and 684 are misaligned.

PR #183, opened 2018-08-16
Add Status.Table.Manual - classes that describe manual updates to status

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6119

These classes will be used by the status page and job to post manual updates in a geo-redundant way.

Depends on https://github.com/NuGet/NuGet.Jobs/pull/537, https://github.com/NuGet/NuGet.Status/pull/57


src/NuGet.Services.Status.Table/CursorEntity.cs
```diff
@@ -9,18 +9,20 @@ namespace NuGet.Services.Status.Table
...
         public CursorEntity()
         {
         }
 
-        public CursorEntity(DateTime value)
-            : base(DefaultPartitionKey, DefaultRowKey)
+        public CursorEntity(string name, DateTime value)
+            : base(DefaultPartitionKey, name)
         {
+            Name = name;
```
skofman1: check input

src/NuGet.Services.Status.Table/Manual/AddStatusEventManualChangeEntity.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace NuGet.Services.Status.Table.Manual
+{
+    public class AddStatusEventManualChangeEntity : ManualStatusChangeEntity
```
skofman1: file header

src/NuGet.Services.Status.Table/Manual/AddStatusMessageManualChangeEntity.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace NuGet.Services.Status.Table.Manual
+{
+    public class AddStatusMessageManualChangeEntity : ManualStatusChangeEntity
+    {
+        public AddStatusMessageManualChangeEntity()
+        {
+        }
+
+        public AddStatusMessageManualChangeEntity(
```
skofman1: headers everywhere

src/NuGet.Services.Status.Table/Manual/AddStatusMessageManualChangeEntity.cs
```diff
@@ -0,0 +1,32 @@
...
+            bool eventIsActive)
+            : base(ManualStatusChangeType.AddStatusMessage)
+        {
+            EventAffectedComponentPath = eventAffectedComponentPath;
+            EventStartTime = eventStartTime;
+            MessageContents = messageContents;
+            EventIsActive = eventIsActive;
+        }
+
+        public string EventAffectedComponentPath { get; set; }
```
skofman1: lots of duplication between classes with regards to properties. Add an abstract class.

src/NuGet.Services.Status.Table/Manual/ManualStatusChangeType.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Status.Table.Manual
+{
+    public enum ManualStatusChangeType
+    {
+        AddStatusEvent,
+        EditStatusEvent,
```
skofman1: I wouldn't bother with edit if I were you.. Edit=delete+add
But not a blocker

scottbommarito: I like the actions being explicitly like this. Also I've already done the work haha.

src/NuGet.Services.Status.Table/Manual/AddStatusMessageManualChangeEntity.cs
scottbommarito: I initially was going to do this, but it was really hard to determine an inheritance model that both made sense and provided value.

Ultimately, I settled on this, because the only meaningful inheritance is that every `ManualStatusChangeEntity` has a type and time that the change was applied.

Any other level of inheritance is just overcomplicating the design.

src/NuGet.Services.Status.Table/CursorEntity.cs
```diff
@@ -9,18 +9,20 @@ namespace NuGet.Services.Status.Table
...
             Value = value;
         }
 
+        public string Name { get; set; }
```
zivkan: given that name is passed in the constructor to the rowKey value, shouldn't the setter also set the base's RowKey property?  Or just make Name's getter and setter just call RowKey's getter and setter?

src/NuGet.Services.Status.Table/Manual/ManualStatusChangeType.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Status.Table.Manual
+{
+    public enum ManualStatusChangeType
```
zivkan: Since this enum is cast to int elsewhere and persisted to storage, each enum value should be assigned a numeric value. This makes sure if any values are added or deleted in the future, there's a reduced risk of deserialisation breaking.

src/NuGet.Services.Status.Table/CursorEntity.cs
scottbommarito: Removing setter and making getter return RowKey.

Name shouldn't be changed.

src/NuGet.Services.Status.Table/Manual/ManualStatusChangeEntity.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.WindowsAzure.Storage.Table;
+using System;
```
loic-sharma: Your `using`s seem out of order

PR #187, opened 2018-08-23
Status.Table - Add IncidentGroupEntity

scottbommarito:
Currently:
- each incident is linked to an event
- each event is associated with an component
- each message is linked to an event

This has become noisy, because we do not want to emit messages for every component (e.g. we shouldn't report for USSC).

With this change:
- each incident is linked to an incident group
- each incident group is associated with a component
- each incident group is linked to an event
- each event is associated with a component
- each message is linked to an event

This will allow us to control what components for which messages are posted, and improve the wording of the messages posted so that they are clearer to customers.


scottbommarito:
Closing this while I continue to work on it...sorry for the noise.

PR #190, opened 2018-08-30
[Revalidation] Add index on completed column

loic-sharma:
Adds an index on the completed column. This is necessary exclude completed validations when counting how many revalidations are pending.

Generates the following migration:

```sql
IF EXISTS (SELECT name FROM sys.indexes WHERE name = N'IX_PackageRevalidations_Enqueued' AND object_id = object_id(N'[dbo].[PackageRevalidations]', N'U'))
    DROP INDEX [IX_PackageRevalidations_Enqueued] ON [dbo].[PackageRevalidations]
CREATE INDEX [IX_PackageRevalidations_Enqueued_Completed] ON [dbo].[PackageRevalidations]([Enqueued], [Completed])
```

Part of https://github.com/NuGet/NuGetGallery/issues/6362


loic-sharma:
The admin panel's query that depends on this index is:

```sql
SELECT COUNT(*) FROM [PackageRevalidations] WHERE [Enqueued] IS NULL OR [Completed] = 1
```

PR #191, opened 2018-09-04
Set sendx5c flag

farehar:
Sendx5c needs to be set to 'true' to use Subject+Issuer based authentication. Added a new constructor to support it.


src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -11,8 +11,11 @@ public class KeyVaultConfiguration
         public string VaultName { get; }
         public string ClientId { get; }
         public X509Certificate2 Certificate { get; }
-        
-        public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2 certificate)
+        public bool SendX5c { get; }
+        public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2 certificate) : this(vaultName, clientId, certificate, false)
+        {
+        }
+        public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2 certificate, bool sendX5c)
```
skofman1: no need for a new ctor. You can use a single ctor and pass a default value of 'false' for sendX5c

PR #195, opened 2018-09-20
Add default and custom FxCop support

chenriksson:
Tracking: https://github.com/NuGet/Engineering/issues/822

Issues with our current way of running FxCop (VSTS build task):
- Single invocation for all solution assemblies, which can result in runtime dependency conflicts
- Does not set CODE_ANALYSIS flag, so code suppressions would not be preserved
- Target discovery assumes project and assembly names match
- Hard to debug on dev machines

Proposal here is to use the `RunCodeAnalysis` msbuild task to invoke FxCop. By default, it uses the instance installed with VS and the minimum ruleset. This also supports custom FxCop installation, project and ruleset - which will be used by our build agents.

Currently, I've only tested against `NuGet.Services.Metadata` but will be testing other repos before I merge.

### DIFF
--- a/build/nuget.codeanalysis.props
+++ b/build/nuget.codeanalysis.props
@@ -16,7 +16,7 @@
     <!-- Default includes 'Excluded' (suppressions) -->
     <CodeAnalysisSaveMessagesToReport>Active</CodeAnalysisSaveMessagesToReport>
     
-    <CodeAnalysisLogFile Condition="'$(FXCOP_OUTPUT_DIRECTORY)' != ''">$(FXCOP_OUTPUT_DIRECTORY)/$(TargetFileName).CodeAnalysisLog.xml</CodeAnalysisLogFile>
+    <CodeAnalysisLogFile Condition="'$(FXCOP_OUTPUT_DIRECTORY)' != ''">$(FXCOP_OUTPUT_DIRECTORY)/$(AssemblyName).CodeAnalysisLog.xml</CodeAnalysisLogFile>
   </PropertyGroup>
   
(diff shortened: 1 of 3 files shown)


build/nuget.codeanalysis.props
```diff
@@ -0,0 +1,36 @@
...
+    <RunCodeAnalysis>true</RunCodeAnalysis>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    
+    <CodeAnalysisSearchGlobalAssemblyCache>true</CodeAnalysisSearchGlobalAssemblyCache>
+    <CodeAnalysisIgnoreInvalidTargets>true</CodeAnalysisIgnoreInvalidTargets>
+    <CodeAnalysisIgnoreGeneratedCode>true</CodeAnalysisIgnoreGeneratedCode>
+    <CodeAnalysisIgnoreMissingIndirectReferences>true</CodeAnalysisIgnoreMissingIndirectReferences>
+    <CodeAnalysisOutputToConsole>true</CodeAnalysisOutputToConsole>
+    <CodeAnalysisVerbose>false</CodeAnalysisVerbose>
+    <CodeAnalysisSummary>true</CodeAnalysisSummary>
```
chenriksson: Matches config currently stored in build definition(s).

chenriksson:
Verified minRuleSet against the following repos:
- NuGet.Services.Metadata
- NuGet.Jobs
- NuGetGallery
- NuGet.Internal.Jobs
- NuGet.Shims
- NuGet.Services.EndToEnd
- NuGetMonitoring
- ServerCommon

PR #196, opened 2018-09-21
Add support for ReplyTo on EmailMessageData

xavierdecoster:


src/NuGet.Services.Messaging/EmailMessageData.cs
```diff
@@ -53,6 +55,7 @@ public class EmailMessageData
 
             CC = cc ?? new List<string>();
             Bcc = bcc ?? new List<string>();
+            ReplyTo = replyTo ?? new List<string>();
```
skofman1: what is the difference between "Reply To" and "From"?

xavierdecoster: When sending an email, you can specify that the ReplyTo address is different from the FromAddress.

Example: https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Services/MessageService.cs#L122-L123

PR #202, opened 2018-10-12
Add entity key to PackageValidationMessageData

shishirx34:
We need to pass `PackageKey` in the message from gallery so that `Orchestrator` can perform appropriate actions for revalidating a symbols package. The idea is we will not pass `PackageKey`(or null) in the message when a package is pushed for validation(first time) with revalidation we will pass in the `Key` for the entity. 

I am open to generalizing this and naming it `EntityKey` or `Key`, I kept it `PackageKey` in context to that of the `PackageId` and `PackageVersion` in the message.

### DIFF
--- a/src/NuGet.Services.Validation/PackageValidationMessageData.cs
+++ b/src/NuGet.Services.Validation/PackageValidationMessageData.cs
@@ -12,7 +12,7 @@ namespace NuGet.Services.Validation
            string packageId,
            string packageVersion,
            Guid validationTrackingId)
-         : this(packageId, packageVersion, validationTrackingId, validatingType: ValidatingType.Package, deliveryCount: 0, packageKey: null)
+         : this(packageId, packageVersion, validationTrackingId, validatingType: ValidatingType.Package, deliveryCount: 0, entityKey: null)
         {
         }
 
@@ -21,8 +21,8 @@ namespace NuGet.Services.Validation
             string packageVersion,
             Guid validationTrackingId,
             ValidatingType validatingType,
-            int? packageKey = null)
-          : this(packageId, packageVersion, validationTrackingId, validatingType, deliveryCount: 0, packageKey: packageKey)
+            int? entityKey = null)
+          : this(packageId, packageVersion, validationTrackingId, validatingType, deliveryCount: 0, entityKey: entityKey)
         {
         }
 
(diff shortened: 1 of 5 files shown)


tests/NuGet.Services.Validation.Tests/TestData.resx
```diff
@@ -121,12 +121,18 @@
     <value>{"PackageId":"NuGet.Versioning","PackageVersion":"4.3","ValidationTrackingId":"14b4c1b8-40e2-4d60-9db7-4b7195e807f5"}</value>
   </data>
   <data name="SerializedPackageValidationMessageData2" xml:space="preserve">
-    <value>{"PackageId":"NuGet.Versioning","PackageVersion":"4.3","PackageNormalizedVersion":"4.3.0","ValidationTrackingId":"14b4c1b8-40e2-4d60-9db7-4b7195e807f5"}</value>
+    <value>{"PackageId":"NuGet.Versioning","PackageVersion":"4.3","PackageNormalizedVersion":"4.3.0","ValidationTrackingId":"14b4c1b8-40e2-4d60-9db7-4b7195e807f5","PackageKey":123}</value>
```
cristinamanum: Add a new unit test to validate that old messages (without PackageKey) can be deserialized)

shishirx34: Already done in the same commit: 
See: https://github.com/NuGet/ServerCommon/pull/202/files#diff-c51e4ec72117c8bef027ab33e5947aa6R78
and: https://github.com/NuGet/ServerCommon/pull/202/files#diff-c51e4ec72117c8bef027ab33e5947aa6R72
and: https://github.com/NuGet/ServerCommon/pull/202/files#diff-c51e4ec72117c8bef027ab33e5947aa6R81
and: https://github.com/NuGet/ServerCommon/pull/202/files#diff-c51e4ec72117c8bef027ab33e5947aa6R121

src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -51,5 +54,6 @@ public class PackageValidationMessageData
         public Guid ValidationTrackingId { get; }
         public int DeliveryCount { get; }
         public ValidatingType ValidatingType { get; }
+        public int? PackageKey { get; }
```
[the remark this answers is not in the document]
shishirx34: I am fine with `EntityKey`.

PR #204, opened 2018-10-17
Changes to signing targets for job MicroBuild integration

scottbommarito:
https://github.com/NuGet/Engineering/issues/1728

This PR consists of two changes:
1 - We need to set `MicroBuild_NuPkgSigningEnabled` to false so that it doesn't automatically sign the nupkgs we produce (we are not required to sign them).
2 - Adding `sign.scripts.targets` that will be used by both `NuGet.Services.Metadata` and `NuGet.Jobs` to sign the `Scripts\` files.

### DIFF
--- a/build/sign.scripts.targets
+++ b/build/sign.scripts.targets
@@ -1,7 +1,7 @@
 <Project>
   <!--
   "EnumerateScriptFilesToSign" is a custom target that signs deployment scripts and executables.
-  The process is non-trivial because only files in $(OutDir) can be signed..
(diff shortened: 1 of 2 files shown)


build/sign.scripts.targets
```diff
@@ -0,0 +1,47 @@
+<Project>
+  <!--
+  "EnumerateScriptFilesToSign" is a custom target that signs deployment scripts and executables.
+  The process is non-trivial because only files in $(OutDir) can be signed..
```
joelverhagen: nit: remove duplicate `.`

PR #205, opened 2018-10-19
Configure MicroBuild to sign ServerCommon

scottbommarito:
https://github.com/NuGet/Engineering/issues/1728

Basically the same as the last couple PRs.

Note that we don't have to sign any third-party dependencies in this project because the `nupkg`s built only contain the output DLLs.

Also, we have some `packages.config` and `project.json` projects in here, so the reference to MicroBuild aren't always `PackageReference`.


tests/NuGet.Services.Contracts.Tests/NuGet.Services.Contracts.Tests.csproj
```diff
@@ -58,5 +58,9 @@
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- 
+  Even though this is a test project, it needs to be strong-named because it tests internal fields of NuGet.Indexing.
```
joelverhagen: NuGet.Indexing? If we are copying comment then perhaps they are not that informative.

scottbommarito: Woops!

I think it's still nice to have the comment, I just should fix it so that it says the right project.

PR #209, opened 2018-10-24
Support Key Vault secrets with Expiry Time

farehar:
Key vault returns SAS tokens for keyvault managed storage accounts. The SAS tokens are valid for a certain period of time. Updated the ISecretReader interface to support returning both the value and expiration time for a secret. Also updated the CacheSecretReader to refresh the secret value if it's near expiry or expired(interval configurable via constructor)


src/NuGet.Services.KeyVault/ISecretReader.cs
```diff
@@ -2,11 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using System;
```
chenriksson: nit: sort usings

src/NuGet.Services.KeyVault/ISecretReader.cs
```diff
@@ -2,11 +2,14 @@
...
 
 namespace NuGet.Services.KeyVault
 {
     public interface ISecretReader
     {
         Task<string> GetSecretAsync(string secretName);
+
+        Task<Tuple<string, DateTime?>> GetSecretValueAndExpiryAsync(string secretName);
```
chenriksson: I'd suggest returning a strong type, to allow more flexibility in the future. For instance, if we wanted to return the secret type, secret version, cache time, etc.

```
// based on https://docs.microsoft.com/en-us/previous-versions/azure/mt149351%28v%3dazure.100%29
interface ISecret {
    public string Value { get; }
    public DateTime? Expiration { get; }
    // public string ContentType { get; } // not needed now
}
interface ISecretReader
{
    Task<string> GetSecretAsync(string secretName);
    Task<ISecret> GetSecretObjectAsync(string secretName);
}
```

Also suggesting more general name like `GetSecretObjectAsync` in case we return more secret properties / attributes in the future.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -34,30 +43,33 @@ public async Task<string> GetSecretAsync(string secretName)
...
         {
-            return (DateTime.UtcNow - secret.CacheTime) >= _refreshInterval;
+            return (((DateTime.UtcNow - secret.CacheTime) >= _refreshInterval) ||
+                (secret.ExpiryDate != null &&  (secret.ExpiryDate - DateTime.UtcNow) <= _refreshIntervalBeforeExpiry));
         }
 
         /// <summary>
         /// A cached secret.
         /// </summary>
         private class CachedSecret
```
chenriksson: If we return an `ISecret`, we could have `KeyVaultSecret` and `CachedSecret` implementations... one for the KV specific properties/attributes, and this one adding the CacheTime property.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -10,20 +10,29 @@ namespace NuGet.Services.KeyVault
...
+        public const int DefaultRefreshIntervalBeforeExpirySec = 60 * 30 ; // 30 minutes 
 
         private readonly ISecretReader _internalReader;
         private readonly ConcurrentDictionary<string, CachedSecret> _cache;
-        private readonly TimeSpan _refreshInterval;
+        private readonly TimeSpan _refreshInterval;        
+        private readonly TimeSpan _refreshIntervalBeforeExpiry;
 
-        public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec)
+        public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec, int refreshIntervalBeforeExpirySec = DefaultRefreshIntervalBeforeExpirySec)
```
skofman1: please add a comment explaining the parameters

skofman1:
Is the expiration time for SAS tokens something KeyVault determines or is it configurable?

src/NuGet.Services.KeyVault/ISecretReader.cs
farehar: updated

src/NuGet.Services.KeyVault/CachingSecretReader.cs
farehar: added.

farehar:
@skofman1: Validity period is specified in SAS Definitions. We can have multiple sas definitions for same storage account. We call key vault with a combination of storage account name and sas definition. So, we can get sas tokens with different lifetimes for the same storage account.

Here's a example in powershell to create a sas definition.(I believe we can do similiar via C#).

$validityPeriod = [System.Timespan]::FromDays(1)
$readSasName = "readBlobSas"

Set-AzureKeyVaultManagedStorageSasDefinition -Service Blob -ResourceType Container,Service -VaultName $keyVaultName -AccountName $accountName -Name $readSasName -Protocol HttpsOnly -ValidityPeriod $validityPeriod -Permission Read,List

src/NuGet.Services.KeyVault/EmptySecretReader.cs
```diff
@@ -10,6 +11,28 @@ public class EmptySecretReader : ISecretReader
...
+        {            
+            return Task.FromResult((ISecret)new EmptySecret(secretName));
+        }
+
+        private class EmptySecret : ISecret
+        {
+            public EmptySecret(string name)
+            {
+                Name = name;
+                Value = null;
```
agr: `Value` should return the value of `Name`.  Otherwise, it is inconsistent with `GetSecretAsync` method above.

src/NuGet.Services.KeyVault/ISecretReader.cs
```diff
@@ -1,12 +1,20 @@
...
 using System.Threading.Tasks;
 
 namespace NuGet.Services.KeyVault
 {
     public interface ISecretReader
     {
         Task<string> GetSecretAsync(string secretName);
+        Task<ISecret> GetSecretObjectAsync(string secretName);
+    }
+    public interface ISecret
```
agr: Please, move it to its own file.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -34,41 +42,51 @@ public async Task<string> GetSecretAsync(string secretName)
...
+            public CachedSecret(ISecret secret)
             {
-                Value = value;
+                Name = secret.Name;
+                Value = secret.Value;                
+                Expiration = secret.Expiration;
                 CacheTime = DateTimeOffset.UtcNow;
             }
-
+            public CachedSecret(string name, string value, DateTime? expiryDate)
```
skofman1: is this ctor used?

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -34,41 +42,51 @@ public async Task<string> GetSecretAsync(string secretName)
...
         }
 
         /// <summary>
         /// A cached secret.
         /// </summary>
-        private class CachedSecret
+        private class CachedSecret : ISecret
         {
-            public CachedSecret(string value)
+            public CachedSecret(ISecret secret)
```
skofman1: you could store the ISecret and instead of copying the internal values

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
```diff
@@ -45,6 +45,7 @@
     <Compile Include="CachingSecretReader.cs" />
     <Compile Include="CachingSecretReaderFactory.cs" />
     <Compile Include="CertificateUtility.cs" />
+    <Compile Include="ISecret.cs" />
```
skofman1: I don't see the file in this PR

src/NuGet.Services.KeyVault/CachingSecretReader.cs
farehar: No. Removed.

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
farehar: missed it. adding it now.

skofman1:
@farehar , thanks for the explanation on the creating of SAS definitions. How will the "get" part from KeyVault look like? You mention 

> We call key vault with a combination of storage account name and sas definition.

The current KeyVaultReader calls KeyVault with "secret" name. Will this need to change?

src/NuGet.Services.KeyVault/CachingSecretReader.cs
farehar: We want CachedSecret to implement ISecret interface as we're returning it in GetSecretObjectAsync method.

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -56,5 +62,24 @@ private async Task<string> GetTokenAsync(string authority, string resource, stri
 
             return result.AccessToken;
         }
+
+        private class KeyVaultSecret : ISecret
```
chenriksson: Now that a public API returns this, I would make it public and putting in a separate file. If so, we should also add arg validation.

farehar:
@skofman1 : We use the same api. Secret name will be in this format: "storageaccountname"-"sasdefinition"

If storageaccountname is "psgallerystorageaccount" and sasdefintion is "readsas". We will call  KeyVault.GetSecretAsync("psgallerystorageaccount"-"sasdefinition"). It will return the SAS Token.

farehar:
In Powershell

$readSasToken = (Get-AzureKeyVaultSecret -VaultName $keyVaultName -SecretName "$accountName-$readSasName").SecretValueText

src/NuGet.Services.KeyVault/EmptySecretReader.cs
```diff
@@ -10,6 +11,28 @@ public class EmptySecretReader : ISecretReader
...
         {
             return Task.FromResult(secretName);
+        }        
+
+        public Task<ISecret> GetSecretObjectAsync(string secretName)
+        {            
+            return Task.FromResult((ISecret)new EmptySecret(secretName));
+        }
+
+        private class EmptySecret : ISecret
```
chenriksson: This class seems unnecessary - can we just reuse `KeyVaultSecret`? e.g., `return new KeyVaultSecret(secretName, secretName, null);`

src/NuGet.Services.KeyVault/CachingSecretReader.cs
chenriksson: Agree - ideally CachedSecret is wrapper class that also adds cache time.

skofman1:
> using System;

please add header

---
Refers to: src/NuGet.Services.KeyVault/ISecret.cs:1 in d1f596b. [](commit_id = d1f596b8e5a69d987a105ce87d13fa5498dca0bf, deletion_comment = False)

farehar:
I've added the headers. Please merge.

farehar:
@chenriksson : I've added the tests. Please review and merge.

tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
```diff
@@ -70,5 +70,53 @@ public async Task WhenGetSecretIsCalledCacheIsRefreshedIfPastInterval()
...
+
+            // Assert
+            mockSecretReader.Verify(x => x.GetSecretObjectAsync(It.IsAny<string>()), Times.Once);
+            Assert.Equal(secretObject1.Value, firstSecretValue);
+            Assert.Equal(secretObject1.Expiration, firstSecretExpiration);
+            Assert.Equal(secretObject2.Value, firstSecretValue);
+            Assert.Equal(secretObject2.Expiration, firstSecretExpiration);
+
+            // Arrange 2
+            await Task.Delay(TimeSpan.FromSeconds(3));
```
agr: Instead of sleeping for 3 seconds here, we can just make `secret1` expire immediately (it would be easy to do if `secret1` was mock of `ISecret` instead of `KeyVaultSecret`). Then we'll not have a test that takes seconds to run.

agr: Also having expiration set for 3 seconds and sleeping for 3 seconds begs for a race condition, which might cause test flakiness.

src/NuGet.Services.KeyVault/ISecret.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
+    public interface ISecret
+    {
+        string Name { get; }
+        string Value { get; }
+        DateTime? Expiration { get; }
```
agr: Can we please have `DateTimeOffset` everywhere?

tests/NuGet.Services.KeyVault.Tests/CachingSecretReaderFacts.cs
farehar: Updated so that the secret expires immediately.

PR #214, opened 2018-10-26
Fix email messaging artifacts to include all dependencies.

cristinamanum:
The change is similar with https://github.com/NuGet/NuGet.Jobs/commit/293675041b30b6c81574e7d31fb430f2d4d05d2d#diff-f0e02e42e53ec40f5d3b8bed031ad09e

PR #215, opened 2018-10-27
All packages to PackageReference

agr:
Current master branch builds produce broken packages: for example, `NuGet.Services.Messaging.Email.nupkg` includes `NuGet.Services.Entities` in the package instead of referenced as a dependency. This effectively prevents entity updates in Gallery (`Add-Migration` produces empty migrations), if one only updates `NuGet.Services.Entities` reference without updating the `NuGet.Services.Messaging.Email` reference.
This happens due to some issues between PC, project.json and PR projects.
This PR 
* makes all projects to use PackageReference;
* Updates build script to use `New-ProjectPackage` to produce .nupkgs;
* Drops the NupkgWrench call from the build script.


joelverhagen:
Does this supersede https://github.com/NuGet/ServerCommon/pull/214/files?

agr:
@joelverhagen, yes

chenriksson:
@agr @joelverhagen Since PackageRef is a big change affecting dependencies, can we also apply other standardizations at the same time?
- Target Net462
- Tools version "15"
- Autogenerate binding redirects
- Dependency cleanup (remove transitive dependencies, which should no longer be necessary)

Cleanup of dependencies would help us get to the point of flipping the WarnAsError switch.

agr:
There is only one project that generates executable: `AzureSqlConnectionTest` and for it auto generate binding redirects is already enabled. As far as I understand, this option makes no sense for the libraries.

Regarding the transitive dependencies: most of the projects updated (all except one) were project.json projects, for which transitive dependencies were not supposed to be listed in the first place.

Tools version and target framework were updated.

agr:
I looked through the nuspecs and contents of the packages produced to make sure no extra files are included and proper dependencies are listed. But ultimately, the indication of correctness is that services that use it does not break, so we'll have to update and deploy.

src/NuGet.Services.Build/NuGet.Services.Build.csproj
```diff
@@ -52,20 +56,23 @@
...
+      <Version>0.3.0</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop">
+      <Version>1.8.24</Version>
+    </PackageReference>
+    <PackageReference Include="NuGet.Build.Tasks.Pack">
+      <Version>4.8.0</Version>
+      <PrivateAssets>all</PrivateAssets>
```
joelverhagen: nit, this should have
```xml
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
```

4.8.0 is not marked as a development dependency, but it should be treated as one. Therefore, it should have the same asset rules as `MicroBuild.Core`.

agr: This was taken from [NuGetGallery.Core.csproj](https://github.com/NuGet/NuGetGallery/blob/9070951769c2a57f7d56746bdec48938df80bf94/src/NuGetGallery.Core/NuGetGallery.Core.csproj#L292-L295). Do we need to update it there, too?

PR #217, opened 2018-10-29
Add SendX5c field to Configuration

rebro-msft:
This was causing problems when trying to call this in Validation.Orchestrator, because we don't use the default false behavior. Now it can be configured.


rebro-msft:
Can someone please merge this PR for me?  I do not have permissions to perform that.

PR #219, opened 2018-10-31
Include the Service Bus message ID in the log scope

joelverhagen:
This will help debugging of message duplication.


src/NuGet.Services.ServiceBus/BrokeredMessageWrapper.cs
```diff
@@ -26,6 +26,11 @@ public BrokeredMessageWrapper(BrokeredMessage brokeredMessage)
         public int DeliveryCount => BrokeredMessage.DeliveryCount;
         public IDictionary<string, object> Properties => BrokeredMessage.Properties;
         public DateTimeOffset EnqueuedTimeUtc => new DateTimeOffset(BrokeredMessage.EnqueuedTimeUtc);
+        public string MessageId
+        {
+            get => BrokeredMessage.MessageId;
+            set => BrokeredMessage.MessageId = value;
```
loic-sharma: What's the setter for? Do we need this today?

joelverhagen: I was just blindly wrapping the API exposed by the SDK. I can remove the setter if you feel strongly since I have no plans to call it. I put it there to just reduce future friction in expanding this wrapper.

PR #222, opened 2018-11-07
Don't attempt to send emails if they have no recipients

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/6632

`EmailRecipients.None` is being compared to by reference, which always fails because it is unused outside of this check.


src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
```diff
@@ -35,16 +43,16 @@ public AsynchronousEmailMessageService(IEmailMessageEnqueuer emailMessageEnqueue
...
             bool copySender = false,
             bool discloseSenderAddress = false)
         {
             var recipients = emailBuilder.GetRecipients();
 
-            if (recipients == EmailRecipients.None)
+            if (!recipients.To.Any())
             {
-                // Optimization: no need to construct message body when no recipients.
+                _logger.LogInformation("Message has no recipients. Cannot send.");
```
loic-sharma: >Cannot send [](start = 67, length = 11)

`Cannot send.` could be confused as an error. What do you think of a message like `Skipping message as it has no recipients`.

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
```diff
@@ -27,6 +32,9 @@ public AsynchronousEmailMessageService(IEmailMessageEnqueuer emailMessageEnqueue
                 throw new ArgumentNullException(nameof(emailBuilder));
             }
 
+            _logger.LogInformation("Sending message with CopySender {CopySender} and DiscloseSenderAddress {DiscloseSenderAddress}",
+                copySender, discloseSenderAddress);
+
             var message = CreateMessage(
```
loic-sharma: >CreateMessage [](start = 26, length = 13)

It's not clear here that `CreateMessage` may return `null`, which `EnqueueMessageAsync` handles by doing no work. What do you think of renaming `CreateMessage` to `CreateMessageOrNull`, and not calling `EnqueueMessageAsync` if `CreateMessageOrNull` returned a `null` message?

scottbommarito: I don't want to change the structure of the code here that much. I agree it's confusing however.

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
scottbommarito: Split it into two log messages.

`Cannot create message to send as it has no recipients.` in this function
`Skipping enqueueing message because it is null or has no recipients.` in `EnqueueMessageAsync`

PR #223, opened 2018-11-08
Divided symbol error message

cristinamanum:
On error was used for both cases: when the pdb is not portable or when a pdb did not have a corresponding dll.  Split the error is two.

### DIFF
--- a/src/NuGet.Services.Validation.Issues/ValidationIssue.cs
+++ b/src/NuGet.Services.Validation.Issues/ValidationIssue.cs
@@ -46,8 +46,8 @@ namespace NuGet.Services.Validation.Issues
             ValidationIssueCode.OnlySignatureFormatVersion1Supported,
             ValidationIssueCode.AuthorCounterSignaturesNotSupported,
             ValidationIssueCode.PackageIsNotSigned,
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.Validation.Issues/ValidationIssue.cs
```diff
@@ -45,8 +46,9 @@ public abstract class ValidationIssue : IValidationIssue
             ValidationIssueCode.OnlySignatureFormatVersion1Supported,
             ValidationIssueCode.AuthorCounterSignaturesNotSupported,
             ValidationIssueCode.PackageIsNotSigned,
-            ValidationIssueCode.SymbolErrorCode_MatchingPortablePDBNotFound,
-            ValidationIssueCode.SymbolErrorCode_ChecksumDoesNotMatch
+            ValidationIssueCode.SymbolErrorCode_MatchingAssemblyNotFound,
+            ValidationIssueCode.SymbolErrorCode_ChecksumDoesNotMatch,
+            ValidationIssueCode.SymbolErrorCode_PdbIsNotPortable
```
joelverhagen: nit: trailing common on all items to reduce diff noise next time.

PR #224, opened 2018-11-08
Set default TTL for EmailMessageData to 2 days

xavierdecoster:


joelverhagen:
How will we know when a message is dropped due to this TTL?

xavierdecoster:
@joelverhagen we won't, unless we use a strategy to move expired messages to DLQ, which would defeat the purpose of trying to ensure we don't hold on to these messages for longer than 2 days.

See https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-expiration#entity-level-expiration

PR #225, opened 2018-11-09
Improve signing time of ServerCommon

scottbommarito:
https://github.com/NuGet/Engineering/issues/1887

Now: 50 minutes
After: 7 minutes

How did I do it?
- Build and sign solution in parallel (`/m`)
- Turn off signing when packing (`/p:SignType:none`)
- Sign intermediate assemblies in `sign.microbuild.targets` (necessary because otherwise `New-ProjectPackage` will replace the signed output assemblies with the unsigned intermediate assemblies)

I expect we can get similar wins by doing this in our other repositories as well.

Additionally, I changed our `verbosity` from `minimal` back to the default. I did this because I think it will make it easier for us to debug build issues in the future.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -194,6 +194,7 @@ Function Build-Solution {
     # Build the solution
     $opts = , $SolutionPath
     $opts += "/p:Configuration=$Configuration;BuildNumber=$(Format-BuildNumber $BuildNumber)"
+    $opts += "/m"
     
     if ($TargetProfile) {
         $opts += "/p:TargetProfile=$TargetProfile"
@@ -201,10 +202,6 @@ Function Build-Solution {
     
     if ($Target) {
         $opts += "/t:$Target"
-    }
-    
-    if (-not $VerbosePreference) {
-        $opts += '/verbosity:minimal'
     }
     
     if ($MSBuildProperties) {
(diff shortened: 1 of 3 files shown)


build/common.ps1
```diff
@@ -194,6 +194,7 @@ Function Build-Solution {
     # Build the solution
     $opts = , $SolutionPath
     $opts += "/p:Configuration=$Configuration;BuildNumber=$(Format-BuildNumber $BuildNumber)"
+    $opts += "/m"
```
chenriksson: Would be useful to add a comment on this, or link to the following... so we don't accidentally remove it.
https://docs.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2017#-maxcpucount-switch

build/common.ps1
```diff
@@ -203,10 +204,6 @@ Function Build-Solution {
         $opts += "/t:$Target"
     }
     
-    if (-not $VerbosePreference) {
-        $opts += '/verbosity:minimal'
-    }
-
```
chenriksson: Looks like this removes our powershell verbosity support altogether. Instead, change the verbosity at the command line or change the default?

scottbommarito: We don't have any mechanism on VSTS to increase/decrease verbosity of specific builds, and I was completely unaware that we had this functionality for local builds.

The current verbosity is way too quiet, especially for debugging on VSTS.  We should probably make `-verbose` make it more verbose instead of silencing the build if `-verbose` is not provided.

I think we should design this functionality at a later point. I made these changes because without more verbose logs, it was very difficult to debug our builds on VSTS, and I considered that was a high enough priority to fix as it will be much more difficult to improve our build/signing times in  other repositories without improving the debuggability.

build/common.ps1
scottbommarito: yup, done!

PR #229, opened 2018-11-21
Update Service Bus SDK dependency

loic-sharma:
Also, this fixes some build warnings. Preview build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2225776&view=logs

See https://github.com/NuGet/NuGetGallery/issues/6692

### DIFF
--- a/tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
+++ b/tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
@@ -96,7 +96,7 @@ namespace NuGet.Services.Configuration.Tests
             Assert.Equal(default(NoConversionFromStringToThisClass), unsupportedFromDictionary);
             Assert.Equal(defaultNoConversion, unsupportedFromDictionaryWithDefault);
             // Safety check to prevent the test from passing if defaultNoConversion is equal to default(NoConversionFromStringToThisClass)
-            Assert.NotEqual(defaultNoConversion, default(NoConversionFromStringToThisClass));
+            Assert.NotEqual(default(NoConversionFromStringToThisClass), defaultNoConversion);
         }
 
         [Theory]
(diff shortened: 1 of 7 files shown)


tests/NuGet.Services.Validation.Issues.Tests/ValidationIssuesFacts.cs
```diff
@@ -92,7 +92,7 @@ public class TheDeserializeMethod
         {
             [Theory]
             [MemberData(nameof(DeserializationOfIssuesWithNoPropertiesData))]
-            public void DeserializationOfIssuesWithNoProperties(ValidationIssueCode code, string data)
```
loic-sharma: The `data` parameter was completely unused

tests/NuGet.Services.Configuration.Tests/DictionaryExtensionsFacts.cs
```diff
@@ -106,7 +106,10 @@ public void GetOrThrowThrowsIfKeyNotFound<T>(T value)
             // Arrange
             const string notKey = "notKey";
 
-            IDictionary<string, string> dictionary = new Dictionary<string, string>();
+            IDictionary<string, string> dictionary = new Dictionary<string, string>()
+            {
+                { "otherKey", value.ToString() }
```
loic-sharma: There was a build warning because `value` was unused. However, the `value` is used as a generic parameter for the `GetOrThrow` on line 115.

skofman1:
What are the benefits of the new SDK version?

loic-sharma:
Talked to @skofman1 offline. The hope was that we may see less deadlettering if we updated our Service Bus SDK. After sending this PR, I did some more investigations and I believe I found the source of our deadlettering. See: https://github.com/NuGet/NuGetGallery/issues/6624

I will fix this issue first. If we need to continue investing in reducing our deadlettering noise after that change, I will reopen this PR.

PR #230, opened 2018-11-21
Improve SubscriptionProcessor's telemetry for messages exceeding expected duration

loic-sharma:
Today, we have to parse Application Insights logs to figure out how a message handler's duration. This makes it impossible to dashboard and monitor. This metric will let us detect messages that caused deadlettering by exceeding their expected duration.

I also fixed a bunch of build warnings as part of this change.

Part of: https://github.com/NuGet/NuGetGallery/issues/6624
Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2228537&view=logs

### DIFF
--- a/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
+++ b/src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
@@ -91,8 +91,11 @@ namespace NuGet.Services.ServiceBus
 
             TrackMessageLags(brokeredMessage);
 
+            var callGuid = Guid.NewGuid();
+            var stopwatch = Stopwatch.StartNew();
+
             using (var scope = _logger.BeginScope($"{nameof(SubscriptionProcessor<TMessage>)}.{nameof(OnMessageAsync)} {{CallGuid}} {{CallStartTimestamp}} {{MessageId}}",
-                Guid.NewGuid(),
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.ServiceBus/SubscriptionProcessor.cs
```diff
@@ -104,18 +107,33 @@ private async Task OnMessageAsync(IBrokeredMessage brokeredMessage)
...
+                            "Handler did not finish processing message after {DurationSeconds} seconds, requeueing message to be reprocessed",
+                            stopwatch.Elapsed.TotalSeconds);
+
+                        _telemetryService.TrackMessageHandlerDuration<TMessage>(stopwatch.Elapsed, callGuid, handled: false);
                     }
                 }
                 catch (Exception e)
                 {
-                    _logger.LogError(Event.SubscriptionMessageHandlerException, e, "Requeueing message as it was unsuccessfully processed due to exception");
+                    _logger.LogError(
```
dtivel: Should `callGuid` be logged in the error?  It only matters if you want to associate an error with the duration metric.

loic-sharma: It is logged in the error because of the scope we created on line 97 :)

PR #231, opened 2018-11-27
Ensure separate email copy is sent to sender when asked

xavierdecoster:


src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
```diff
@@ -40,7 +44,14 @@ public class AsynchronousEmailMessageService : IMessageService
                 copySender,
                 discloseSenderAddress);
 
-            return EnqueueMessageAsync(message);
+            await EnqueueMessageAsync(message);
+
+            // If we're asked to copy the sender but cannot disclose the sender address,
+            // we should send a separate email to the sender.
+            if (copySender && !discloseSenderAddress)
```
cristinamanum: CreateMessage below seem to handle arguments copySender and discloseSenderAddress. However does not handle the case copySender == true and discloseSenderAddress == false. Can a comment be added at the CreateMessage level to explain the intended behavior of the method?

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
```diff
@@ -40,7 +44,14 @@ public class AsynchronousEmailMessageService : IMessageService
                 copySender,
                 discloseSenderAddress);
 
-            return EnqueueMessageAsync(message);
+            await EnqueueMessageAsync(message);
```
cristinamanum: It this email sent to the emailBuiulder.GetRecipients().ReplyTo ? If yes will these people receive double emails on copysender== true and discloseSenderAddress == false?

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
```diff
@@ -40,7 +44,14 @@ public class AsynchronousEmailMessageService : IMessageService
...
+
+            // If we're asked to copy the sender but cannot disclose the sender address,
+            // we should send a separate email to the sender.
+            if (copySender && !discloseSenderAddress)
+            {
+                await EnqueueMessageToSenderAsync(emailBuilder);
```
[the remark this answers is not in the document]
xavierdecoster: That would work too.

I did it like this mainly for consistency: this logic is also part of the `MarkdownMessageService`, which implements the same interface. If we make it a caller responsibility to construct a separate message, we'll be duplicating code that decides whether or not to send a separate message or CC the sender, while there's no difference in implementation: the message body is always modified in the same way, so is the subject, and the message's recipients. It also means we can no longer test this logic as part of the message service, but need to test this for every caller.

As this logic already was part of the other `IMessageService` implementation, and this was the least effort, I decided to stick to this approach for now.

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
xavierdecoster: Added a comment https://github.com/NuGet/ServerCommon/pull/231/commits/80f313122cc55dde7a452ffcb11e22d2b5797216

src/NuGet.Services.Messaging.Email/AsynchronousEmailMessageService.cs
xavierdecoster: The `ReplyTo` address is the address that email replies will be sent to, by the recipient of these messages. 

Most emails we send are from `noreply@`, but occasionally they're not. In the case of `ContactOwnersMessage` for example (which is the scenario that we're fixing here), the `FromAddress` (of the user contacting the owner) is the `ReplyTo` address.

xavierdecoster:
I verified a local Gallery using the locally built Servercommon packages, behavior seems OK. Will merge this so we can release Servercommon for consumption in gallery.

PR #234, opened 2018-12-21
Added NuGet.Services.Licenses

agr:
Progress on https://github.com/NuGet/NuGetGallery/issues/6721

Code mostly comes from https://github.com/NuGet/NuGet.Licenses.
Added higher level wrapper around `ILicenseExpressionSegmentator` that enables to do the whole transformation in single call.
`ILicenseExpressionSegmentator.GetLicenseExpressionSegments` now accepts `NuGetLicenseExpression` as an input instead of `LicenseOperator` enabling it to "traverse" expression trees consisting of a single license.


build.ps1
```diff
@@ -117,7 +118,8 @@ Invoke-BuildStep 'Creating artifacts' { `
             "src\NuGet.Services.Status.Table\NuGet.Services.Status.Table.csproj",
             "src\NuGet.Services.Messaging\NuGet.Services.Messaging.csproj",
             "src\NuGet.Services.Messaging.Email\NuGet.Services.Messaging.Email.csproj",
-            "src\NuGet.Services.Entities\NuGet.Services.Entities.csproj"
+            "src\NuGet.Services.Entities\NuGet.Services.Entities.csproj",
+            "src\NuGet.Services.Licenses\NuGet.Services.Licenses.csproj"
```
loic-sharma: Does this work? Shouldn't there be backticks? It seems that we stopped adding backticks after line 118

src/NuGet.Services.Licenses/ILicenseExpressionSegmentator.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// The complicated case is:
+    /// 
+    /// (((MIT OR   (ISC))))
+    /// 
+    /// i.e. we cannot just restore the whole sequence from the tree alone since there is some extra
+    /// parentheses that have no representation in the expression tree. That's what 
+    /// <see cref="SplitFullExpression(string, IReadOnlyCollection{CompositeLicenseExpressionSegment})"/>
+    /// does.
+    /// </remarks>
+    public interface ILicenseExpressionSegmentator
```
loic-sharma: >ILicenseExpressionSegmentator [](start = 21, length = 29)

Will we use this class directly, or, will everyone use `LicenseExpressionSplitter`? If the latter, could we make these private methods on `LicenseExpressionSplitter`?

src/NuGet.Services.Licenses/LicenseExpressionSplitter.cs
```diff
@@ -0,0 +1,44 @@
...
+                throw new ArgumentNullException(nameof(licenseExpression));
+            }
+
+            var expressionRoot = _parser.Parse(licenseExpression);
+
+            var meaningfulSegments = _segmentator.GetLicenseExpressionSegments(expressionRoot);
+            return _segmentator.SplitFullExpression(licenseExpression, meaningfulSegments);
+        }
+
+        public static List<CompositeLicenseExpressionSegment> Split(string licenseExpression)
```
joelverhagen: >Split [](start = 62, length = 5)

Why do we need this static?

It seems like there are two "approaches" implemented here:

1. Use an instance of this thing, via interface (perhaps to facilitate unit testing)
2. Use a static method.

We should pick one.

joelverhagen:
> <?xml version="1.0" encoding="utf-8"?>

test.ps1 should mention this project

---
Refers to: tests/NuGet.Services.Licenses.Tests/NuGet.Services.Licenses.Tests.csproj:1 in be00e1c. [](commit_id = be00e1ca64c259d70bbea11652ec3ae034e41cb3, deletion_comment = False)

src/NuGet.Services.Licenses/LicenseExpressionSegmentator.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                fullSegmentList.Add(
+                    new CompositeLicenseExpressionSegment(licenseExpression.Substring(startIndex),
+                    CompositeLicenseExpressionSegmentType.Other));
+            }
+
+            return fullSegmentList;
+        }
+
+        private static void TraverseExpressionTreeInOrder(NuGetLicenseExpression root, List<CompositeLicenseExpressionSegment> segmentList)
```
joelverhagen: Will this stack overflow with malicious input from gallery?

joelverhagen: Or will there be a more "friendly" error message for the user?

build.ps1
agr: Backticks are only needed if by the end of a line it looks like a complete Powershell "statement" (not sure what's the correct term). Since we have a comma, Powershell parser expects something after that comma and will check the next line without a need to put a backtick there.

src/NuGet.Services.Licenses/ILicenseExpressionSegmentator.cs
agr: NuGet.Licenses uses segmentator alone. It's a pain to test if we leave a single entry point to the whole operation, so I kept a low level implementation that exposes each underlying operation separately, and a high level "glue" that provides a nicer interface.

src/NuGet.Services.Licenses/LicenseExpressionSegmentator.cs
agr: Gallery has a 500 character limit on a license expression length, so the worst case would be around 100 levels deep tree (` OR ` with a single character license id (we don't have those) totals 5 characters, 500 / 5 = 100), which is not enough to have a stack overflow.

joelverhagen: Sounds good. Consider throwing here if the input is too long so that we don't have this hidden dependency.

src/NuGet.Services.Licenses/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Reflection;
+using System.Runtime.InteropServices;
```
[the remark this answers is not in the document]
agr: We generally don't put copyright headers into `AssemblyInfo.cs` files.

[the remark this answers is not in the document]
joelverhagen: Nearly all of them in gallery have it 🤷

src/NuGet.Services.Licenses/LicenseExpressionSplitter.cs
agr: Removed the method.

src/NuGet.Services.Licenses/ILicenseExpressionSegmentator.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// series of "segments" that allow to identify elements (license ids, exception ids, operators, etc.)
+    /// of license expression inside that string individually.
+    /// </summary>
+    /// <remarks>
+    /// The goal is to be able to "pretty print" the license expression on a web page: e.g given the
+    /// (MIT OR ISC OR GPL WITH Classpath-exception)
+    /// license expression we should be able to display it while linking individual elements to their
+    /// respective license or exception URLs (so 'MIT' when displayed on a web page is a link to
+    /// MIT license page, while 'OR', 'WITH', whitespace and parentheses are not).
+    /// So we split the string into the series of "segments" each representing some element of the expression.
```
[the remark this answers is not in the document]
agr: Added examples for each method.

src/NuGet.Services.Licenses/ILicenseExpressionSplitter.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Licenses
+{
+    /// <summary>
+    /// Higher level wrapper around <see cref="ILicenseExpressionSegmentator"/>
+    /// </summary>
+    public interface ILicenseExpressionSplitter
```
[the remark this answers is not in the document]
agr: This would complicate testing the `SplitExpression` method, as we won't be able to abstract away the library call.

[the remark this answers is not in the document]
scottbommarito: Maybe use `InternalsVisibleTo`?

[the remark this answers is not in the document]
scottbommarito: Nevermind--you need to still expose it so that it can be used in the constructor of the classes that use it.

PR #271, opened 2019-01-04
Introduce package deprecation entities

xavierdecoster:


src/NuGet.Services.Entities/Package.cs
```diff
@@ -233,6 +233,11 @@ public bool HasReadMe
         /// </summary>
         public PackageStatus PackageStatusKey { get; set; }
 
+        /// <summary>
+        /// The package deprecation status key, referring to the <see cref="PackageDeprecationStatus"/> enum.
+        /// </summary>
+        public PackageDeprecationStatus PackageDeprecationStatusKey { get; set; }
```
joelverhagen: Can we wait on adding this until we have measure performance gains? This adds complexity and potential data inconsistencies.

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Gets or sets the primary key for the entity.
+        /// </summary>
+        public int Key { get; set; }
+
+        /// <summary>
+        /// Gets or sets the foreign key of the deprecated package entity.
+        /// </summary>
+        [Index("IX_PackageDeprecations_DeprecatedPackageKey", IsUnique = true)]
+        public int DeprecatedPackageKey { get; set; }
```
joelverhagen: We typically do indexes in the `DbContext` right? `OnModelCreating`?

https://github.com/NuGet/NuGetGallery/blob/995709555c8cbcba05925756b27eefbcff0c8242/src/NuGetGallery.Core/Entities/EntitiesContext.cs#L318-L331

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <summary>
+        /// Gets or sets the alternate package entity.
+        /// </summary>
+        public virtual Package AlternatePackage { get; set; }
+
+        /// <summary>
+        /// Gets or sets the foreign key of the user that executed the package deprecation.
+        /// </summary>
+        public int DeprecatedByKey { get; set; }
```
joelverhagen: This should be nullable to allow for user deletion, right?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <summary>
+        /// Gets or sets the alternate package registration entity.
+        /// </summary>
+        public virtual PackageRegistration AlternatePackageRegistration { get; set; }
+
+        /// <summary>
+        /// Gets or sets the foreign key of the altnerate package entity.
+        /// </summary>
+        public int AlternatePackageKey { get; set; }
```
joelverhagen: nullable?

src/NuGet.Services.Entities/PackageDeprecation.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Gets or sets the deprecated package entity.
+        /// </summary>
+        [Required]
+        public virtual Package DeprecatedPackage { get; set; }
+
+        /// <summary>
+        /// Gets or sets the foreign key of the altnerate package registration entity.
+        /// </summary>
+        public int AlternatePackageRegistrationKey { get; set; }
```
joelverhagen: nullable?

src/NuGet.Services.Entities/PackageDeprecation.cs
joelverhagen: Oh maybe we are not supporting that for now since user delete is soft right now?

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the primary key for the entity.
+        /// </summary>
+        public int Key { get; set; }
+
+        /// <summary>
+        /// Gets or sets the CVSS rating for this package vulnerability.
+        /// </summary>
+        public int CVSSRating { get; set; }
```
joelverhagen: Nullable? IIRC, the UI allows any three of these to be picked.

src/NuGet.Services.Entities/PackageVulnerability.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Gets or sets the JSON-serialized array of CVE ID's related to this package vulnerability.
+        /// </summary>
+        public string CVEIds { get; set; }
```
joelverhagen: nit: JSON object with a `version` property please

skofman1:
Not directly related to this PR, but I think making schema changes would be much much easier if we move entities back to the Gallery solution and use a project reference instead of a package reference. I checked and there are no projects depending on the entities package besides Gallery. For simplicity we can leave namespaces as is and even move the project as is.

agr:
Regarding moving the Entities back: the Powershell team would benefit from it greatly. They maintain their own fork of Gallery with slightly different schema, so having entities in NuGetGallery would mean they have to maintain less forks if they want to keep up with the upstream (that's what they are trying to do now: get up to date with what we have).

scottbommarito:
@skofman1 @agr The entities have been moved back. I will be moving this PR to NuGetGallery.

PR #272, opened 2019-01-08
Enable KeyVault injection into environment variables

joelverhagen:
This will enable configuration for Web Apps.

This is the ARM resource that sets configuration on an Azure slot.
```json
{
    "apiVersion": "2018-02-01",
    "name": "appsettings",
    "type": "config",
    "dependsOn": [
        "[resourceId('Microsoft.Web/sites/slots', parameters('websiteName'), variables('slotName'))]"
    ],
    "properties": "[parameters('appSettings')]"
}
```

This is a trimmed example of the parameters file:
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "appSettings": {
            "value": {
                "SearchService:SearchServiceName": "jver-search"
            }
        }
    }
}
```

The environment variables come out like this at runtime.

**Key**: `APPSETTING_SearchService:SearchServiceName`
**Value**: `jver-search`

The benefit of this approach over pure JSON configuration on disk (like jobs) is that we can modify app settings from the portal.

image

### DIFF
--- a/src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
@@ -9,11 +9,26 @@ namespace NuGet.Services.Configuration
 {
     public static class ConfigurationBuilderExtensions
     {
-        public static IConfigurationBuilder AddInjectedJsonFile(this IConfigurationBuilder configurationBuilder, string path, ISecretInjector secretInjector)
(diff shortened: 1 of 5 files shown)


joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6793

PR #273, opened 2019-01-09
[Feature Flags] Add feature flag/flighting package

loic-sharma:
Adds the client libraries for detecting the state of feature flags. Usage:

```csharp
IFeatureFlagClient features = ...;
IFeatureFlagUser currentUser = ...;

if (features.IsEnabled("NuGetGallery.TyposquattingDetection", defaultValue: false))
{
    ...
}

if (features.IsEnabled("NuGetGallery.TyposquattingDetection", currentUser, defaultValue: false))
{
  ...
}
```

Part of https://github.com/NuGet/NuGetGallery/issues/6788
Gallery integration: https://github.com/NuGet/NuGetGallery/pull/6806

### DIFF
--- a/src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
+++ b/src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
@@ -1,0 +1,102 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.Services.FeatureFlags
+{
+    public class FeatureFlagCacheService : IFeatureFlagCacheService
+    {
+        private readonly IFeatureFlagStorageService _storage;
+        private readonly FeatureFlagOptions _options;
+        private readonly IFeatureFlagTelemetryService _telemetryServiceOrNull;
+        private readonly ILogger<FeatureFlagCacheService> _logger;
+
+        private FeatureFlagsAndRefreshTime _latestFlags;
+
+        public FeatureFlagCacheService(
+            IFeatureFlagStorageService storage,
+            FeatureFlagOptions options,
+            ILogger<FeatureFlagCacheService> logger)
+          : this(storage, options, telemetryService: null, logger: logger)
+        {
+        }
+
+        public FeatureFlagCacheService(
+            IFeatureFlagStorageService storage,
+            FeatureFlagOptions options,
+            IFeatureFlagTelemetryService telemetryService,
+            ILogger<FeatureFlagCacheService> logger)
+        {
+            _storage = storage ?? throw new ArgumentNullException(nameof(storage));
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _telemetryServiceOrNull = telemetryService;
+            _latestFlags = null;
+        }
+
+        public FeatureFlags GetLatestFlagsOrNull()
+        {
+            return _latestFlags?.FeatureFlags;
+        }
+
+        public async Task RefreshAsync()
+        {
+            var refreshTime = DateTimeOffset.UtcNow;
+            var latestFlags = await _storage.GetAsync();
+
+            _latestFlags = new FeatureFlagsAndRefreshTime(latestFlags, refreshTime);
+        }
+
+        public async Task RunAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    _logger.LogInformation("Refreshing the feature flags...");
+
+                    await RefreshAsync();
+
+                    _logger.LogInformation("Refreshed the feature flags");
+                }
+                catch (Exception e)
+                {
+                    _logger.LogError(0, e, "Unable to refresh the feature flags due to exception");
+                }
+
+                // Report the feature flags' time time since the flags were last successfully refreshed, aka their "staleness".
+                var staleness = (_latestFlags == null)
+                    ? TimeSpan.MaxValue
+                    : DateTimeOffset.UtcNow - _latestFlags.RefreshStartTime;
+
+                _telemetryServiceOrNull?.TrackFeatureFlagStaleness(staleness);
+
+                _logger.LogInformation(
+                    "Feature flags were last refreshed {Staleness} ago. Sleeping for {SleepDuration} before next refresh...",
+                    staleness,
+                    _options.RefreshInterval);
+
+                await Task.Delay(_options.RefreshInterval);
+            }
+        }
+
+        private class FeatureFlagsAndRefreshTime
+        {
+            public FeatureFlagsAndRefreshTime(FeatureFlags featureFlags, DateTimeOffset refreshTime)
+            {
+                FeatureFlags = featureFlags;
+                RefreshStartTime = refreshTime;
+            }
+
+            public FeatureFlags FeatureFlags { get; }
+            public DateTimeOffset RefreshStartTime { get; }
+        }
+    }
+}
--- a/src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
+++ b/src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
@@ -1,0 +1,119 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Net.Mail;
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.Services.FeatureFlags
+{
+    public class FeatureFlagClient : IFeatureFlagClient
+    {
+        private readonly IFeatureFlagCacheService _cache;
+        private readonly ILogger<FeatureFlagClient> _logger;
+
+        public FeatureFlagClient(IFeatureFlagCacheService cache, ILogger<FeatureFlagClient> logger)
+        {
+            _cache = cache ?? throw new ArgumentNullException(nameof(cache));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public bool IsEnabled(string feature, bool defaultValue)
+        {
+            var latest = _cache.GetLatestFlagsOrNull();
+            if (latest == null)
+            {
+                _logger.LogError(
+                    "Couldn't determine status of feature {Feature} as the flags haven't been loaded",
+                    feature);
+
+                return defaultValue;
+            }
+
+            if (!latest.Features.TryGetValue(feature, out var featureStatus))
+            {
+                _logger.LogWarning(
+                    "Couldn't determine status of feature {Feature} as it isn't in the latest flags",
+                    feature);
+
+                return defaultValue;
+            }
+
+            switch (featureStatus)
+            {
+                case FeatureStatus.Enabled:
+                    return true;
+
+                case FeatureStatus.Disabled:
+                    return false;
+
+                default:
+                    _logger.LogError(
+                        "Unknown feature status {FeatureStatus} for feature {Feature}",
+                        feature,
+                        featureStatus);
+
+                    return defaultValue;
+            }
+        }
+
+        public bool IsEnabled(string flightName, IFlightUser user, bool defaultValue)
+        {
+            var latest = _cache.GetLatestFlagsOrNull();
+            if (latest == null)
+            {
+                _logger.LogError(
+                    "Couldn't determine status of flight {Flight} as the flags haven't been loaded",
+                    flightName);
+
+                return defaultValue;
+            }
+
+            if (!latest.Flights.TryGetValue(flightName, out var flight))
+            {
+                _logger.LogWarning(
+                    "Couldn't determine status of flight {Flight} as it isn't in the latest feature flags",
+                    flightName);
+
+                return defaultValue;
+            }
+
+            if (flight.All)
+            {
+                return true;
+            }
+
+            if (flight.Accounts.Contains(user.Username, StringComparer.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            if (TryParseEmailDomain(user.EmailAddress, out var domain) && flight.Domains.Contains(domain, StringComparer.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            if (flight.SiteAdmins && user.IsSiteAdmin)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private bool TryParseEmailDomain(string email, out string domain)
+        {
+            try
+            {
+                domain = new MailAddress(email).Host;
+
+                return true;
+            }
+            catch (Exception) { }
+
+            domain = null;
+            return false;
+        }
+    }
+}
--- a/src/NuGet.Services.FeatureFlags/FeatureFlagOptions.cs
+++ b/src/NuGet.Services.FeatureFlags/FeatureFlagOptions.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.FeatureFlags
+{
+    public class FeatureFlagOptions
+    {
+        /// <summary>
+        /// How frequently the feature flags should be refreshed.
+        /// </summary>
+        public TimeSpan RefreshInterval { get; set; } = TimeSpan.FromMinutes(1);
+    }
+}
--- a/src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
+++ b/src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.FeatureFlags
+{
+    /// <summary>
+    /// The service that caches and periodically refreshes the feature flags' state.
+    /// </summary>
+    public interface IFeatureFlagCacheService
+    {
+        /// <summary>
+        /// Continuously refresh the feature flags' cache.
+        /// </summary>
+        /// <param name="cancellationToken">Cancelling this token will complete the returned task.</param>
+        /// <returns>A task that completes once the cancellation token is cancelled.</returns>
+        Task RunAsync(CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Refresh the feature flags cache once.
+        /// </summary>
+        /// <remarks>This should be called at app startup to guarantee feature flags have been loaded.</remarks>
+        /// <returns>A task that completes once the cached feature flags have been refreshed.</returns>
+        Task RefreshAsync();
+
+        /// <summary>
+        /// Fetch the latest cached flags. This should be called after either <see cref="RunAsync(CancellationToken)"/>
+        /// or <see cref="RefreshAsync(CancellationToken)"/>.
+        /// </summary>
+        /// <returns>The latest cached flags, or null if the flags have never been loaded successfully.</returns>
+        FeatureFlags GetLatestFlagsOrNull();
+    }
+}
--- a/src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
+++ b/src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.FeatureFlags
+{
+    public interface IFeatureFlagClient
+    {
+        /// <summary>
+        /// Get whether a feature is enabled. This method does not throw.
+        /// </summary>
+        /// <param name="feature">The unique identifier for this feature. This is case insensitive.</param>
+        /// <param name="defaultValue">The value to return if the status of the feature is unknown.</param>
+        /// <returns>Whether the feature is enabled.</returns>
+        bool IsEnabled(string feature, bool defaultValue);
+
+        /// <summary>
+        /// Get whether a flight is enabled for a user. This method does not throw.
+        /// </summary>
+        /// <param name="flight">The unique identifier for this flight. This is case insensitive.</param>
+        /// <param name="user">The user whose status should be determined.</param>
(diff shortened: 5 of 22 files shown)


src/NuGet.Services.FeatureFlags/IFeatureFlagStorageService.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGet.Services.FeatureFlags
+{
+    public interface IFeatureFlagStorageService
```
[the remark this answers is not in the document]
joelverhagen: Good move.

src/NuGet.Services.FeatureFlags/LatestFlagsResult.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace NuGet.Services.FeatureFlags
+{
+    /// <summary>
+    /// The latest known feature flags' state.
+    /// </summary>
+    public class LatestFlagsResult
+    {
+        public static readonly LatestFlagsResult Uninitialized = new LatestFlagsResult(LatestFlagsStatus.Uninitialized, flags: null);
+
+        private LatestFlagsResult(LatestFlagsStatus status, FeatureFlagsState flags)
```
zhhyu: >LatestFlagsResult [](start = 16, length = 17)

public properties should be above public functions?

src/NuGet.Services.FeatureFlags/LatestFlagsResult.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            if (flags == null) throw new ArgumentNullException(nameof(flags));
+
+            return new LatestFlagsResult(LatestFlagsStatus.Ok, flags);
+        }
+
+        /// <summary>
+        /// The status of the feature flags.
+        /// </summary>
+        public LatestFlagsStatus Status { get; }
```
zhhyu: public properties should be above public functions?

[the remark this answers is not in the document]
loic-sharma: I believe the style guide is to place static members before instance members. I'd prefer to keep it as is, unless you feel strongly about it.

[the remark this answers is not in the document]
zhhyu: Ok, I see.

src/NuGet.Services.FeatureFlags/FeatureFlagRefreshService.cs
```diff
@@ -0,0 +1,95 @@
...
+        private class FeatureFlagsAndCacheTime
+        {
+            public FeatureFlagsAndCacheTime(FeatureFlagsState featureFlags, DateTimeOffset cacheTime)
+            {
+                FeatureFlags = featureFlags ?? throw new ArgumentNullException(nameof(featureFlags));
+                CacheTime = cacheTime;
+            }
+
+            public FeatureFlagsState FeatureFlags { get; }
+            public DateTimeOffset CacheTime { get; }
```
shishirx34: nit: move the fields above the constructor, that is the general guideline we've been following in the code base.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.Services.FeatureFlags
+{
+    public class FeatureFlagClient : IFeatureFlagClient
+    {
+        private readonly IFeatureFlagRefreshService _flags;
```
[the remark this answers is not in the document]
zhhyu: I agree.

src/NuGet.Services.FeatureFlags/FeatureFlagRefreshService.cs
shishirx34: counter examples 😆 

- [ConfirmationViewModel](https://github.com/NuGet/NuGetGallery/blob/bf7e98f696e346b55cb1eba38866682ae936160f/src/NuGetGallery/ViewModels/ConfirmationViewModel.cs)
- [LogOnViewModel](https://github.com/NuGet/NuGetGallery/blob/bf7e98f696e346b55cb1eba38866682ae936160f/src/NuGetGallery/ViewModels/LogOnViewModel.cs)

In any case I don't feel too strongly.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,79 @@
...
+        private readonly ILogger<FeatureFlagClient> _logger;
+
+        public FeatureFlagClient(IFeatureFlagRefreshService flags, ILogger<FeatureFlagClient> logger)
+        {
+            _flags = flags ?? throw new ArgumentNullException(nameof(flags));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public bool IsEnabled(string feature, bool @default)
```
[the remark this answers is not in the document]
loic-sharma: Done

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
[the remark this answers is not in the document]
loic-sharma: `_featureFlagRefreshService` is a little long. I renamed the class to `FeatureFlagCacheService` and renamed the field to `_cache`. Let me know if that's still not clear.

src/NuGet.Services.FeatureFlags/FeatureFlagRefreshService.cs
```diff
@@ -0,0 +1,95 @@
...
+            _storage = storage ?? throw new ArgumentNullException(nameof(storage));
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _latestFlags = null;
+        }
+
+        public LatestFlagsResult GetLatestFlags()
+        {
+            var latestFlags = _latestFlags;
```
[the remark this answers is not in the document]
loic-sharma: Added

src/NuGet.Services.FeatureFlags/LatestFlagsResult.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static LatestFlagsResult Stale(FeatureFlagsState flags)
+        {
+            if (flags == null) throw new ArgumentNullException(nameof(flags));
+
+            return new LatestFlagsResult(LatestFlagsStatus.Stale, flags);
+        }
+
+        public static LatestFlagsResult Ok(FeatureFlagsState flags)
+        {
+            if (flags == null) throw new ArgumentNullException(nameof(flags));
```
joelverhagen: nit: curlies around the single line (this is more consistent with our existing style)

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
```diff
@@ -0,0 +1,97 @@
...
+                    await RefreshAsync(cancellationToken);
+
+                    _logger.LogInformation("Refreshed feature flags");
+                }
+                catch (Exception e)
+                {
+                    _logger.LogError(0, e, "Unable to refresh the feature flags due to exception");
+                }
+
+                await Task.Delay(_options.RefreshInterval);
```
joelverhagen: >Delay [](start = 27, length = 5)

log refresh interval

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
```diff
@@ -0,0 +1,97 @@
...
+            return LatestFlagsResult.Ok(latestFlags.FeatureFlags);
+        }
+
+        public async Task RefreshAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            _latestFlags = new FeatureFlagsAndCacheTime(
+                await _storage.GetAsync(),
+                DateTimeOffset.UtcNow);
```
joelverhagen: >UtcNow [](start = 31, length = 6)

it's a little unclear about what point in time "cache time" refers to. Is this when it was fetched? when it was requested? The different is basically before or after `GetAsync()` completes. 

Which should it be? A comment would be good.

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
```diff
@@ -0,0 +1,97 @@
...
+                    return LatestFlagsResult.Stale(latestFlags.FeatureFlags);
+                }
+            }
+
+            return LatestFlagsResult.Ok(latestFlags.FeatureFlags);
+        }
+
+        public async Task RefreshAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
```
joelverhagen: >cancellationToken [](start = 12, length = 17)

Why observe a cancellation token here? If it gets passed to the storage abstraction, I get it, but it's not doing too much here.

src/NuGet.Services.FeatureFlags/FeatureFlagOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        /// How frequently the feature flags should be refreshed.
+        /// </summary>
+        public TimeSpan RefreshInterval { get; set; } = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// The maximum refresh staleness allowed by the <see cref="IFeatureFlagCacheService"/>.
+        /// If the threshold is reached, the returned feature flags will be labeled as stale.
+        /// </summary>
+        public TimeSpan? MaximumStaleness { get; set; } = TimeSpan.FromHours(1);
```
joelverhagen: >MaximumStaleness [](start = 25, length = 16)

What's the use case for a null `MaximumStaleness`?

src/NuGet.Services.FeatureFlags/Models/FlightState.cs
```diff
@@ -0,0 +1,43 @@
...
+            Domains = domains ?? new List<string>();
+        }
+
+        /// <summary>
+        /// Whether this flight is enabled for all users. If true, all other properties are ignored.
+        /// </summary>
+        public bool All { get; }
+
+        /// <summary>
+        /// Whether this flight is enabled for NuGet.org administrators.
```
joelverhagen: >administrators [](start = 57, length = 14)

Nit: clarify how SiteAdmins = false + an admin username in Accounts behaves.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            switch (featureStatus)
+            {
+                case FeatureStatus.Enabled:
+                    return FeatureResult.Enabled;
+
+                case FeatureStatus.Disabled:
+                    return FeatureResult.Disabled;
+
+                default:
```
joelverhagen: >default [](start = 16, length = 7)

Why is `FeatureStatus` an enum?

This default case is weird. If it was a bool we wouldn't have to worry about it.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        private bool TryParseEmailDomain(string email, out string domain)
+        {
+            try
+            {
+                domain = new MailAddress(email).Host;
+
+                return true;
+            }
+            catch (ArgumentNullException) { }
```
joelverhagen: >ArgumentNullException [](start = 19, length = 21)

Why not just `catch {}` here? You have something similar in the cache service and these is called in a "non-throwing" code path.

src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="cancellationToken">Cancelling this token will complete the returned task.</param>
+        /// <returns>A task that completes once the cancellation token is cancelled.</returns>
+        Task RunAsync(CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Refresh the feature flags cache once.
+        /// </summary>
+        /// <param name="cancellationToken">Cancels the refresh.</param>
+        /// <returns>A task that completes once the feature flags have been refreshed.</returns>
+        Task RefreshAsync(CancellationToken cancellationToken = default(CancellationToken));
```
joelverhagen: >RefreshAsync [](start = 13, length = 12)

My guess is that this method is for app startup -- right? If not, what's the intended use? It would be worth documenting that maybe under remarks.

src/NuGet.Services.FeatureFlags/LatestFlagsResult.cs
loic-sharma: 😢 

Fixed

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
loic-sharma: Added

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
loic-sharma: Good catch. This was vestigial code (it was how the `RunAsync` method stopped).

I went ahead and removed the cancellation token from the API call as our storage abstractions don't all accept cancellation tokens.

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
loic-sharma: I tweaked the code a little to make this clearer. Do you think it's good enough now that it doesn't warrant a comment?

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+
+                return FeatureResult.Unknown;
+            }
+
+            if (flight.All)
+            {
+                return FeatureResult.Enabled;
+            }
+
+            if (flight.Accounts.Contains(user.Username))
```
joelverhagen: >Contains [](start = 32, length = 8)

Is this intended to be case sensitive?

NuGet.org usernames and emails are case insensitive due to SQL Server collation.

src/NuGet.Services.FeatureFlags/Models/FeatureFlagsState.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// The state of all features and flights.
+    /// </summary>
+    public class FeatureFlagsState
+    {
+        public FeatureFlagsState(
+            IReadOnlyDictionary<string, FeatureStatus> features,
+            IReadOnlyDictionary<string, FlightState> flights)
+        {
+            Features = features ?? throw new ArgumentNullException(nameof(features));
```
joelverhagen: >features [](start = 23, length = 8)

nit: Can't we can't null here similar to `FlightStatus` ctor? Seems like the same approach should be applied in both places.

src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public interface IFeatureFlagClient
+    {
+        /// <summary>
+        /// Get whether a feature is enabled. This method does not throw.
+        /// </summary>
+        /// <param name="feature">The unique identifier for this feature.</param>
+        /// <param name="defaultValue">The value to return if the status of the feature is unknown.</param>
+        /// <returns>Whether the feature is enabled.</returns>
+        bool IsEnabled(string feature, bool defaultValue);
```
joelverhagen: >string [](start = 23, length = 6)

Specify whether feature is case sensitive or not. Seems fine to be case sensitive but I don't feel strongly.

src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// Get whether a flight is enabled for a user. This method does not throw.
+        /// </summary>
+        /// <param name="flight">The unique identifier for this flight.</param>
+        /// <param name="user">The user whose status should be determined.</param>
+        /// <returns>Whether the flight is enabled for this user.</returns>
+        FeatureResult IsEnabled(string flight, User user);
+    }
+
+    public enum FeatureResult
```
joelverhagen: >FeatureResult [](start = 16, length = 13)

nit: enum in its own file for easy GitHub lookup

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
joelverhagen: Yup, that's much better.

src/NuGet.Services.FeatureFlags/FeatureFlagOptions.cs
loic-sharma: Good catch. There's no use case, I'll remove this.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
loic-sharma: It's an enum because there's 3 values: on, off, or "unknown". The latter happens if the flags haven't been loaded yet, or if the maximum staleness has been reached. Features have to decide how they should behave if the flag's value is unknown.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            return ToBoolean(IsEnabled(flightName, user), defaultValue);
+        }
+
+        public FeatureResult IsEnabled(string feature)
+        {
+            var latest = _cache.GetLatestFlags();
+            if (latest.Status != LatestFlagsStatus.Ok)
+            {
+                _logger.LogWarning(
```
skofman1: LogError. If we get into this state it's very bad.

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+            switch (featureStatus)
+            {
+                case FeatureStatus.Enabled:
+                    return FeatureResult.Enabled;
+
+                case FeatureStatus.Disabled:
+                    return FeatureResult.Disabled;
+
+                default:
+                    _logger.LogWarning(
```
skofman1: This indicates a bug/bad config. Should be LogError

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
```diff
@@ -0,0 +1,152 @@
...
+                    return FeatureResult.Unknown;
+            }
+        }
+
+        public FeatureResult IsEnabled(string flightName, User user)
+        {
+            var latest = _cache.GetLatestFlags();
+            if (latest.Status != LatestFlagsStatus.Ok)
+            {
+                _logger.LogWarning(
```
skofman1: same. Should be LogError

src/NuGet.Services.FeatureFlags/NuGet.Services.FeatureFlags.csproj
```diff
@@ -0,0 +1,92 @@
...
+      <Version>1.1.2</Version>
+    </PackageReference>
+    <PackageReference Include="NuGet.Build.Tasks.Pack">
+      <Version>4.8.0</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\NuGet.Services.Entities\NuGet.Services.Entities.csproj">
```
skofman1: adding dependency on NuGet.Services.Entities is problematic. We intend to remove this package from ServerCommon and place it back in Gallery repo as project ref. 
Also, by having this dependency every project that will use feature flags will have this dependency.
My recommendation would be to remove flighting for admins (it can be done by enabling specific admin users) and remove usage of User class from APIs

src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public async Task RunAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    _logger.LogInformation("Refreshing feature flags...");
```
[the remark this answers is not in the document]
loic-sharma: The feature flags are refreshed only once per minute. This seems acceptable to me...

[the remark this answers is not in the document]
skofman1: I think it will be noisy in our logs, hence suggesting to reduce Log severity and possibly number of logs..

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
loic-sharma: My bad, I got the types mixed up! I received feedback from the spec review that the following JSON was unclear:

```
"Features": {
  "FeatureA": true,
  "FeatureB": false
}
```

The feedback was that the following would be clearer:

```
"Features": {
  "FeatureA": "Enabled",
  "FeatureB": "Disabled"
}
```

joelverhagen: Okay, sounds good.

src/NuGet.Services.FeatureFlags/FeatureFlagOptions.cs
loic-sharma: Talked to Joel offline and we decided to remove this option. Instead, jobs & services will report a staleness metric that will be monitored

src/NuGet.Services.FeatureFlags/Models/FlightState.cs
loic-sharma: Added. Please let me know if you think this is still unclear

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
loic-sharma: Fixed

src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
loic-sharma: Added a remark

src/NuGet.Services.FeatureFlags/FeatureFlagClient.cs
loic-sharma: Good catch. Made usernames and emails case insensitive

src/NuGet.Services.FeatureFlags/Models/FeatureFlagsState.cs
loic-sharma: Fixed

loic-sharma:
@joelverhagen All of your feedback should be addressed, please take another look when you have the chance.

PR #275, opened 2019-02-16
[Feature Flags] Expose the last refresh start time

loic-sharma:
Used by https://github.com/NuGet/NuGetGallery/pull/6895

### DIFF
--- a/src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
+++ b/src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.FeatureFlags/IFeatureFlagCacheService.cs
```diff
@@ -31,5 +32,11 @@ public interface IFeatureFlagCacheService
         /// </summary>
         /// <returns>The latest cached flags, or null if the flags have never been loaded successfully.</returns>
         FeatureFlags GetLatestFlagsOrNull();
+
+        /// <summary>
+        /// Fetch the time at which the flags were last refreshed.
+        /// </summary>
+        /// <returns>The latest cached flags, or null if the flags have never been loaded successfully.</returns>
+        DateTimeOffset? GetRefreshStartTimeOrNull();
```
scottbommarito: I think it'd be a lot clearer if this was just called `GetLatestRefreshTimeOrNull`. From looking at the existing implementation, it seems like the reason you call it `RefreshStartTime` is that you capture the current time **before** you call `_storage.GetAsync`, but to be honest I think that's more of an implementation detail, and simplifying the naming to `RefreshTime` or `RefreshedAt` would improve the readability of the code.

loic-sharma: Good point, I'll rename this

PR #280, opened 2019-03-14
Add recursive to common.ps1 submodules cmdlets

scottbommarito:
This is needed for https://github.com/nuget/engineering/issues/2249 because `NuGet.Status` has a `NuGetGallery` submodule which becomes a nested submodule when `NuGet.Status` is a submodule inside `NuGetDeployment`.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -332,8 +332,7 @@ Function Update-Submodule {
         [string] $Name,
         [string] $Path,
         [string] $Branch,
-        [string] $RemoteUrl,
-        [switch] $Recursive
+        [string] $RemoteUrl
     )
 
     Trace-Log "Configuring submodule $Name ($Path) to use branch $Branch."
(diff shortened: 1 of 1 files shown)


build/common.ps1
```diff
@@ -332,7 +332,8 @@ Function Update-Submodule {
         [string] $Name,
         [string] $Path,
         [string] $Branch,
-        [string] $RemoteUrl
+        [string] $RemoteUrl,
+        [switch] $Recursive
```
joelverhagen: Why is this optional? What is the impact if we just do this everywhere?

scottbommarito: Likely none--however, I thought that it's not clear whether or not somebody updating a submodule will always want to update that submodule and all its nested submodules simultaneously. On the other hand, for `Reset-Submodule`, it seems clear to me that if you're resetting a submodule, you always want to reset its nested submodules.

I thought about it more though, and we can change the code later when the need for complexity arises. Will remove optional parameter.

joelverhagen: Sounds good.

PR #290, opened 2019-03-26
[Feature Flags] Handle nulls as anonymous users

loic-sharma:
The gallery signifies a user that's not logged in as a `null` value, which wasn't handled properly.

Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2530168

### DIFF
--- a/src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
+++ b/src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
@@ -16,6 +16,10 @@ namespace NuGet.Services.FeatureFlags
         /// <summary>
         /// Get whether a flight is enabled for a user. This method does not throw.
         /// </summary>
+        /// <remarks>
+        /// Enabling a flight for all users will include anonymous users. If the flight should never be
+        /// enabled for logged out users, perform an "is logged in" check before calling this API.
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.FeatureFlags/IFeatureFlagClient.cs
```diff
@@ -17,7 +17,7 @@ public interface IFeatureFlagClient
         /// Get whether a flight is enabled for a user. This method does not throw.
```
joelverhagen: Please add a note or remark that if you want logged out users to never be on the flight, perform an "is logged in" check before calling the API. We are choosing one of a couple possible reasonable behaviors for when the `user` is null. I think we are choosing the best (others include: return `false`, return `defaultValue`) but we should provide clarity on how to achieve the other possible behaviors.

loic-sharma: I added a `remark` section. It's a tad wordy, please let me know if you have any suggestions on how to make the remark succinct.

joelverhagen: It's okay for remarks to be wordy 😄. looks good

PR #293, opened 2019-05-08
Add pending migration test

joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/4934
Similar to https://github.com/NuGet/NuGetGallery/pull/7143


tests/NuGet.Services.Validation.Tests/PendingMigrationsTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        public Task InitializeAsync()
+        {
+            return Task.CompletedTask;
+        }
+
+        public async Task DisposeAsync()
+        {
+            if (_dbName == null)
```
loic-sharma: Nit: what's this null check for?

joelverhagen: If we add another test here that does not set `_dbName` we shouldn't run this dispose code.

PR #294, opened 2019-05-14
Attempting to fix building everything when VS 2019 is the only one installed

agr:
Extracted part of the logic of `Get-MsBuildExe` function into `Get-MsBuildExeInternal` so I could try a bunch of things and `return` as soon as we find something. Adjusted the fallback in the original function accordingly.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -107,9 +107,11 @@ Function Get-MSBuildExe {
         }
         
         $installations = @([NuGet.Services.Build.VisualStudioSetupConfigurationHelper]::GetInstancePaths() | ForEach-Object {
-            $MSBuildRoot = Join-Path "$_\MSBuild" ([string]$MSBuildVersion + ".0")
-            Join-Path $MSBuildRoot $MSBuildExeRelPath
-        } | Where-Object { Test-Path $_ })
+            Join-Path "$_\MSBuild" ([string]$MSBuildVersion + ".0")
+            Join-Path $_ "MSBuild\Current"
+        } | ForEach-Object {
+            Join-Path $_ $MSBuildExeRelPath
+        } | Where-Object { Test-Path $_ } | Where-Object { (Get-Item $_).VersionInfo.ProductMajorPart -eq $MSBuildVersion })
         
         if ($installations.Count -ge 1) {
             $MSBuildPath = $installations[0]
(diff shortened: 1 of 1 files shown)


build/common.ps1
```diff
@@ -77,17 +77,51 @@ Function Clear-Artifacts {
...
+{
+    if ($MSBuildVersion -lt 15) {
+        $MSBuildExe = Join-Path $MSBuildRoot ([string]$MSBuildVersion + ".0")
+        $MSBuildPath = Join-Path $MSBuildExe $MSBuildExeRelPath
+        return $MSBuildPath;
+    } elseif ($MSBuildVersion -eq 16) {
+        $pathsToCheck = @(
+            "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe",
+            "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Professional\MSBuild\Current\Bin\MSBuild.exe",
+            "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\MSBuild.exe",
```
agr: Need to verify that top 3 paths are actually found in the wild.
Bottom 2 definitely are.

joelverhagen: At this point, do you think something link `vswhere` is a better approach?
https://github.com/microsoft/vswhere

agr: Sounds reasonable

agr: On the other hand, after talking to @scottbommarito, I made much simpler and more correct update, which should be OK.

build/common.ps1
```diff
@@ -107,9 +107,11 @@ Function Get-MSBuildExe {
         }
         
         $installations = @([NuGet.Services.Build.VisualStudioSetupConfigurationHelper]::GetInstancePaths() | ForEach-Object {
-            $MSBuildRoot = Join-Path "$_\MSBuild" ([string]$MSBuildVersion + ".0")
-            Join-Path $MSBuildRoot $MSBuildExeRelPath
-        } | Where-Object { Test-Path $_ })
+            Join-Path "$_\MSBuild" ([string]$MSBuildVersion + ".0")
```
loic-sharma: Use `[IO.Path]::Combine(...)` when there's more than two path fragments

PR #295, opened 2019-05-17
Add support to any AAD authority when authenticating with Azure Management

skofman1:


src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
```diff
@@ -39,6 +41,7 @@ public AzureManagementAPIWrapper(IAzureManagementAPIWrapperConfiguration configu
             }
 
             _clientCredential = new ClientCredential(configuration.ClientId, configuration.ClientSecret);
+            _authority = string.Format(CultureInfo.InvariantCulture, AuthorityTemplate, configuration.AadTenant);
```
joelverhagen: Throw if null? Or default to `microsoft.onmicrosoft.com`?

src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
```diff
@@ -12,11 +13,12 @@ namespace NuGet.Services.AzureManagement
 {
     public class AzureManagementAPIWrapper : IAzureManagementAPIWrapper
     {
-        private const string Authority = "https://login.microsoftonline.com/microsoft.onmicrosoft.com";
+        private const string AuthorityTemplate = "https://login.microsoftonline.com/{0}";
```
joelverhagen: Are there UTs?

joelverhagen: Would it be more flexible to make the hole Authority URL config? Or will anyone hitting `management.core.windows.net` need to hit `https://login.microsoftonline.com/`?

src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
skofman1: added

src/NuGet.Services.AzureManagement/AzureManagementAPIWrapper.cs
skofman1: this class doesn't have much logic. It's all about connecting to Azure. There are UTs around parsing the output.

skofman1: login.microsoftonline.com is a constant URL for AAD auth. It's used for SQL AAD as well: https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Sql/AzureSqlConnectionStringBuilder.cs#L24

PR #296, opened 2019-05-27
Add package validation set status so that validation sets can be marked as completed

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7185.


src/NuGet.Services.Validation/Migrations/201905272210233_AddValidationSetStatus.cs
```diff
@@ -0,0 +1,18 @@
+namespace NuGet.Services.Validation
```
agr: copyright headers.

agr: Do we put copyright headers to migrations?

src/NuGet.Services.Validation/Migrations/201905272210233_AddValidationSetStatus.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddValidationSetStatus : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.PackageValidationSets", "ValidationSetStatus", c => c.Int(nullable: false, defaultValue: 0));
```
agr: What's the migration story? Specifically, when this migration is applied, all the validation sets in the system (including the ones that are currently in progress) will get the DB entry that says `Unknown`. What's the plan for the service to handle that?

src/NuGet.Services.Validation/Migrations/201905272210233_AddValidationSetStatus.cs
joelverhagen: Nope, I think we try to leave these auto-generated files as-is.

src/NuGet.Services.Validation/Migrations/201905272210233_AddValidationSetStatus.cs
joelverhagen: They will ignore it at first (on SELECT and INSERT thanks to default).

Then we roll out a code change to set the status to in progress and completed at the beginning and end of validation, respectively.

joelverhagen: Oh Unknown may transition to completed. That's fine.

Unknown may suffer from message duplication and therefore deadletter. This can happen today but has no user impact and is relatively rare (6 in 90 days).

joelverhagen: Oh Unknown may transition to completed. That's fine.

Unknown may suffer from message duplication and therefore deadletter. This can happen today but has no user impact and is relatively rare (6 in 90 days).

PR #297, opened 2019-05-29
[WIP] Use SDK style projects

loic-sharma:
# Use SDK style projects

This updates our project files to use MSBuild SDK. This is useful as:

1. This cleans up our project files
1. This lets us use the dotnet CLI, which reduces build times from ~35 seconds to ~8 seconds on my machine
1. This is a required step to target .NET Standard or .NET Core

### Out-of-scope changes

The following changes are out-of-scope and should be done in future pull requests:

1. Target .NET Standard 2.0
1. Generate .snupkg symbol files
1. Update our build scripts to use the dotnet CLI
1. Align dependencies' versions using central version management

### Validation

* [x] Ensure `dotnet restore`, `dotnet build`, `dotnet test`, and `dotnet pack` pass
* [ ] Build packages: [#3274509](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3274509)
* [ ] Deploy to DEV
  * [ ] NuGetGallery
  * [ ] Orchestrator
  * [ ] Run validation migrations
  * [ ] TODO: Status service + job
  * [ ] TODO: Something that uses AzureManagement?
  * [ ] TODO: Something that uses NuGet.Services.Build
  * [ ] TODO: Something that uses NuGet.Services.Cursor/HttpReadCursor.cs 

### Differences before and after

❌ This analysis needs to be redone after the .NET Framework 4.7.2 upgrade.

* Package metadata
   * ⚠️ **Copyright has changed!**
  * ⚠️ "Framework Assembly References" changed dramatically. Need to investigate
  * Packages now have a "title" field set to package id. Consider removing
  * Package description used to be "Package description". This is fixed now
* Assembly metadata
  * ⚠️ **Strong name version has changed!** Strong name version used to be "0.0.0.0" 
  * Need to investigate all added assembly attributes:
    * `AssemblyFileVersion`
    * `AssemblyInformationalVersion`
    * `AssemblyMetadata`
  * `File version` assembly property changed from `0.0.0.0` to `1.0.0.0`
  * `Product version` assembly property changed from `0.0.0.0` to `2.54.0-branch`
* The `Guid` assembly attribute has been removed. Need to investigate

### _HandlePackageFileConflicts Output

✔ This issue has been resolved by updating to .NET Framework v4.7.2

<details>
  <summary>_HandlePackageFileConflicts Output</summary>
  
⚠️ When building, you will see the following build output:

> _HandlePackageFileConflicts:
> message NETSDK1041: Encountered conflict between 'Reference:System.abc.dll' and 'Platform:System.abc.dll'.  NETSDK1033: Choosing 'Reference:System.abc.dll' because AssemblyVersion '4.1.2.0' is greater than '4.0.0.0'.

This warning appears to be caused by our projects and dependencies overriding assemblies that are implicitly added by the SDK using `PackageReference`:

1. Our projects override these assemblies as the implicit assemblies are missing APIs. We may be able to drop our overrides by upgrading our target framework, but that's out of scope for this change.
2. Our dependencies override these assemblies as they target .NET Standard 1.X. We will need to upgrade our assemblies to use .NET Standard 2.0, but doing so would require us to upgrade our target framework as well.

TLDR: Silencing the conflicts requires updating our dependencies and target framework, which is out-of-scope for this change.
</details>


Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MicroBuild.Core">
+      <Version>0.3.0</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="NuGet.Build.Tasks.Pack">
```
loic-sharma: This is no longer necessary in SDK style projects and should be removed

src/NuGet.Services.Cursor/NuGet.Services.Cursor.csproj
```diff
@@ -8,10 +8,15 @@
...
     </PackageReference>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\NuGet.Services.Storage\NuGet.Services.Storage.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Net.Http.WebRequest" />
```
loic-sharma: Ahhhhh that's how you do it! Very nice find

Directory.Build.props
```diff
@@ -0,0 +1,42 @@
...
+    <PackageProjectUrl>https://github.com/NuGet/ServerCommon</PackageProjectUrl>
+    <Copyright>© .NET Foundation. All rights reserved.</Copyright>
+    <IncludeSymbols Condition="'$(IncludeSymbols)' == ''">true</IncludeSymbols>
+    <NoPackageAnalysis Condition="'$(NoPackageAnalysis)' == ''">true</NoPackageAnalysis>
+
+    <!--
+      Our builds generate a file containing the desired assembly attributes.
+      To prevent duplicate attribute compilation errors, disable the SDK's attributes.
+      See: https://github.com/dotnet/sdk/blob/2df7f13c8c25e36e3620b62a420bc6b733e5bbb4/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L65-L99
+      TODO: Remove this
```
loic-sharma: TODO

agr:
Builds still produce a ton of warnings, need to look at those.

joelverhagen:
I think we should align to System.Net.Http 4.3.4 package instead of the assembly reference. Entry points will be generating this binding redirect in almost all cases anyway since most of our ServerCommon packages pull in System.Net.Http 4.3.4 anyways.

Also, we should be able to eliminate the `System.Net.Http.WebRequest` reference since pipelining in that context is nonsense... AFAIK pipeline will only occur on that handler and it is only used for a single request.

The changes in `frameworkAssemblies` are benign. They all come from the GAC anyway so whether the consuming project references them or not is only sugar for cases where some types are in the library's API (not just an implementation detail) and the consuming library needs the reference too.

Assembly file version doesn't matter. I think ideally it would match the assembly version.

AFAIK `Guid` attribute is just for COM. Shouldn't matter for us.

Also, I don't see the "_HandlePackageFileConflicts" warnings mentioned. Is this still occurring? Perhaps you are using a prerelease .NET Core SDK?

Overall this looks great. What next needs to be done to get this in?

Directory.Build.props
```diff
@@ -0,0 +1,33 @@
+<Project>
+  <PropertyGroup>
+    <Authors>.NET Foundation</Authors>
+    <PackageLicenseUrl>https://github.com/NuGet/ServerCommon/blob/master/License.md</PackageLicenseUrl>
+    <PackageProjectUrl>https://github.com/NuGet/ServerCommon</PackageProjectUrl>
+    <Copyright>© .NET Foundation. All rights reserved.</Copyright>
+    <Company>.NET Foundation</Company>
+    <IncludeSymbols Condition="'$(IncludeSymbols)' == ''">true</IncludeSymbols>
+    <NoPackageAnalysis Condition="'$(NoPackageAnalysis)' == ''">true</NoPackageAnalysis>
```
joelverhagen: We should be able to enable package analysis once we switch to license expressions:

```
<PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
```

src/NuGet.Services.Build/NuGet.Services.Build.csproj
```diff
@@ -1,79 +1,16 @@
...
-    <RootNamespace>NuGet.Services.Build</RootNamespace>
-    <AssemblyName>NuGet.Services.Build</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <NuGetPackageImportStamp>
-    </NuGetPackageImportStamp>
+    <TargetFramework>net462</TargetFramework>
+    <Title>NuGet.Services.Build</Title>
+    <Description>Shared component to assist NuGet build scripts.</Description>
+    <Product>NuGet.Services.Build</Product>
```
joelverhagen: `<Product>` should default to assembly name IIRC. What is `<Title>` for?

joelverhagen:
I pushed my suggested changes. Feel free to revert if you want.

src/NuGet.Services.Build/NuGet.Services.Build.csproj
joelverhagen: Removed them. All the happened what `<title>` was removed from .nuspec which is a good thing IMO.

joelverhagen:
> Also, I don't see the "_HandlePackageFileConflicts" warnings mentioned. Is this still occurring? Perhaps you are using a prerelease .NET Core SDK?

I see them now. They are not appearing as warnings for me, just "grey text". Unless we have runtime issues I don't think we should worry about it. Reading around some .NET Core SDK issues I can't make heads or tails of it.

Directory.Build.props
joelverhagen: Pushed this.

joelverhagen:
File version will default to assembly version with `<GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>`, which I pushed.

joelverhagen:
Validation build: [#2913486](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2913486)

loic-sharma:
Thanks for the help @joelverhagen! All your changes look good to me. As for the `Title` and `Product` attributes, I was trying to minimize the changes to the resulting build artifacts. I'm fine with removing these attributes right away. I'd like to test these changes on a few projects before merging this in:

* Gallery validation build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2913675
* Orchestrator: TODO

loic-sharma:
I rebased on top of the latest `master` branch. The previous changes are saved as the `loshar-sdk-backup` branch.

tests/NuGet.Services.Build.Tests/NuGet.Services.Build.Tests.csproj
```diff
@@ -50,15 +13,13 @@
     </PackageReference>
     <PackageReference Include="xunit.runner.visualstudio">
       <Version>2.4.1</Version>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
```
loic-sharma: TODO?

loic-sharma:
The `_HandlePackageFileConflicts` warnings have been resolved now that we've updated to .NET Framework 4.7.2.

tests/NuGet.Services.Logging.Tests/NuGet.Services.Logging.Tests.csproj
```diff
@@ -1,47 +1,15 @@
...
-    <None Include="App.config" />
-  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="Moq">
       <Version>4.10.0</Version>
     </PackageReference>
     <PackageReference Include="SerilogTraceListener">
-      <Version>2.0.10027</Version>
+      <Version>3.2.0</Version>
```
loic-sharma: `NuGet.Services.Logging.csproj` uses v3.2.0, this was causing a package downgrade warning.

joelverhagen:
You have served your purpose, brave PR!

PR #298, opened 2019-06-06
Sort of builds in VSTS, tests fail

agr:
Merged with recent master
Converted all projects to SDK
Fixed the assembly metadata build issues.

Tests pass. By default SDK style projects embedded resources with under wrong names which caused tests to break.


src/NuGet.Services.AzureManagement/NuGet.Services.AzureManagement.csproj
```diff
@@ -2,6 +2,11 @@
 
   <PropertyGroup>
     <TargetFramework>net462</TargetFramework>
+    <Description>Azure Management tools for NuGet services</Description>
+    <Title>NuGet.Services.AzureManagement</Title>
+    <Product>NuGet.Services.AzureManagement</Product>
+    <Copyright>Copyright © .NET Foundation 2015</Copyright>
```
loic-sharma: `Copyright` and `Company` can be pulled up to the `Directory.Build.props` file

agr: It does not seem that it gets picked up from that file, but removed it anyway.

PR #303, opened 2019-07-17
Add notion of "check validator" message for orchestrator

joelverhagen:
Progress on https://github.com/nuget/nugetgallery/issues/7354.

As spec'd in the Queue Back design this adds a message type for orchestrator that just contains the validation ID so that downstream validations can notify orchestrator that they are done.

### DIFF
--- a/src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
+++ b/src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
@@ -14,9 +14,6 @@ namespace NuGet.Services.ServiceBus
     /// <typeparam name="TMessage">A type decorated with a <see cref="SchemaAttribute"/>.</typeparam>
     public class BrokeredMessageSerializer<TMessage> : IBrokeredMessageSerializer<TMessage>
     {
-        private const string SchemaNameKey = "SchemaName";
-        private const string SchemaVersionKey = "SchemaVersion";
-
         private static readonly string SchemaName;
         private static readonly int SchemaVersion;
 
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.Validation/PackageValidationMessageData.cs
```diff
@@ -2,58 +2,83 @@
...
+                    break;
+                default:
+                    throw new NotSupportedException($"The package validation message type '{type}' is not supported.");
+            }
+
+            var notNullCount = new object[]
+            {
+                processValidationSet,
+                checkValidator,
+            }.Count(x => x != null);
```
loic-sharma: This seems a little over kill :P

src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
```diff
@@ -66,17 +69,17 @@ private static void AssertTypeAndSchemaVersion(IBrokeredMessage message, string
             }
         }
 
-        private static int GetSchemaVersion(IBrokeredMessage message)
+        public static int GetSchemaVersion(IBrokeredMessage message)
```
loic-sharma: Thoughts on making these extension methods?

src/NuGet.Services.Validation/PackageValidationMessageData.cs
joelverhagen: Wonder how that slipped in.

loic-sharma: Reverse code golf should be a thing :)

agr:
What is the migration plan? How are we going to deal with the messages that are in ServiceBus when the change is deployed?

joelverhagen:
@agr, the migration plan was documented in the design doc. But this is it:

1. ~~Apply Validation DB migration.~~
1. Deploy symbol orchestrator that handles the new message type.
1. Deploy symbol validator and symbol ingester.
1. Turn on the feature flag.
1. E2E tests and manual testing of symbol validation.
1. Deploy package orchestrator that handles the new message type.
1. Deploy revalidate certificate job.
1. Deploy package validators and ESRP submitter.

src/NuGet.Services.ServiceBus/BrokeredMessageSerializer.cs
joelverhagen: Yeah, I like the idea.

PR #304, opened 2019-07-25
Add RefreshableSecretReader to allow background thread secret refresh

joelverhagen:
1. Add `RefreshSecretReader`. This returns a secret from cache if it exists. The cache lifetime from this class's perspective is forever. If the secret does not exist in the cache, it will be fetched from an underlying secret reader.

1. Add `RefreshSecretReaderFactory`. This holds the reference to the shared secret cache and has a method `RefreshAsync` which is on `IRefreshSecreReaderFactory` that can be called from a "safe" context, e.g. background thread in a web app.

I added a thing called `BlockUncachedReads` which makes is so non-`RefreshAsync` paths cannot read secrets from KeyVault. In other words it means the secret must be in the cache. This is added so that if we have a bug and a request context tries to read a secret, we throw an exception to fail fast instead of deadlocking.

Progress on https://github.com/NuGet/NuGetGallery/issues/7337.

### DIFF
--- a/src/NuGet.Services.KeyVault/EmptySecretReader.cs
+++ b/src/NuGet.Services.KeyVault/EmptySecretReader.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading.Tasks;
 
 namespace NuGet.Services.KeyVault
--- a/src/NuGet.Services.KeyVault/IRefreshableSecretReaderFactory.cs
+++ b/src/NuGet.Services.KeyVault/IRefreshableSecretReaderFactory.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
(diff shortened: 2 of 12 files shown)


src/NuGet.Services.KeyVault/IRefreshSecretReaderFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
+    /// <summary>
+    /// An interface that allows caching and refreshing the secrets fetched by secret readers.
+    /// </summary>
+    public interface IRefreshSecretReaderFactory : ISecretReaderFactory
```
scottbommarito: Maybe consider naming the classes `Refreshable` instead of just `Refresh`? I think it makes it a little clearer that the user is supposed to refresh the secret reader, and that it doesn't happen automatically.

joelverhagen:
> For the Azure Search Service, is the intention that:

Yes, you got it. PR https://github.com/NuGet/NuGet.Services.Metadata/pull/610 illustrates that.

> I think we should try to build in an easy solution to this initialization problem into the library. If it's too much work though, this is fine.

I would rather share the pieces that I know are commonly usable first and move code up later when it becomes more clear.

src/NuGet.Services.KeyVault/IRefreshSecretReaderFactory.cs
joelverhagen: Yeah, I like `Refreshable`.

PR #309, opened 2019-08-28
Rename StartValidationAsync to SendMessageAsync

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7490.

This is a breaking change so I will actively update ServerCommon everywhere to react to this.

### DIFF
--- a/src/NuGet.Services.Validation/IPackageValidationEnqueuer.cs
+++ b/src/NuGet.Services.Validation/IPackageValidationEnqueuer.cs
@@ -9,16 +9,16 @@ namespace NuGet.Services.Validation
     public interface IPackageValidationEnqueuer
     {
         /// <summary>
-        /// Enqueues validation of the specified package to start ASAP.
+        /// Enqueues a package validation message to be consumed as soon as possible.
         /// </summary>
-        /// <param name="message">Package information</param>
-        Task StartValidationAsync(PackageValidationMessageData message);
-        
+        /// <param name="message">Package validation information</param>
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.Validation/IPackageValidationEnqueuer.cs
```diff
@@ -9,16 +9,16 @@ namespace NuGet.Services.Validation
     public interface IPackageValidationEnqueuer
     {
         /// <summary>
-        /// Enqueues validation of the specified package to start ASAP.
+        /// Enqueues a package validation message of the specified package to start ASAP.
```
agr: It probably needs a complete rewording. There is no package directly specified in the method arguments.

agr: > Enqueues the specified validation message to get delivered ASAP.
?

joelverhagen: The type is `PackageValidationMessageData` which is still accurate... if you look at it a certain way I guess. Any type of message that goes into this method is about the validation of a single package. The ways this validation is specified varies (package ID+version, package validation) but it's still about a package validation.

joelverhagen: Perhaps I am missing what you mean.

joelverhagen: Spoke offline. Fixed up now.

PR #310, opened 2019-09-12
Update server common telemetrywrapper to adapt to new interface of IDiagnosticsSource.

zhhyu:
Recently in Gallery.Core, we update the interface of "IDiagnosticsSource" on "TraceEvent", and the new implementation depends on the new member "TraceTrace" in "ITelemetryClient".
Gallery has its own version of "ITelemetryClient" and implementation "TelemetryClientWrapper", while server common has another individual version, which is used by NuGet.Jobs or others.
NuGet.Jobs needs to adapt to the new interface of "IDiagnosticsSource" and implement the updated method "TraceEvent", otherwise the telemetry will be broken, which finally leads to the update of the server common's "ITelemetryClient" and "TelemetryClientWrapper".
The current solution looks a headache for the maintenance work, and we need to have a unified telemetry and diagnostic services.
The Gallery's change is here, and basically the change of this PR follows the Gallery's change:
https://github.com/NuGet/NuGetGallery/pull/7513

### DIFF
--- a/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
+++ b/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
@@ -3,11 +3,14 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.Logging
 {
     public interface ITelemetryClient
     {
+        void TrackTrace(string message, LogLevel logLevel, EventId eventId);
+
         void TrackMetric(
             string metricName,
             double value,
--- a/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
+++ b/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
@@ -7,6 +7,10 @@ using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.Logging
 {
+    /// <summary>
+    /// Interface for the Application Insights TelemetryClient class.
+    /// Notice that NuGetGallery.Services has another interface and implementation for the same functionality.
+    /// </summary>
     public interface ITelemetryClient
     {
         void TrackTrace(string message, LogLevel logLevel, EventId eventId);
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
```diff
@@ -3,11 +3,14 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.Logging
 {
     public interface ITelemetryClient
```
loic-sharma: Consider adding a comment that this mirrors the type in NuGetGallery so that we're less likely to forget these updates in the future.

zhhyu:
The "TrackDependency" is used in Gallery. But the constructor between Gallery and Server common is different, which needs time to figure out the best way to unify the two pieces. I decide to make the smallest change to finish the SDL asap.  So remove the track dependency at this time.

src/NuGet.Services.Contracts/NuGet.Services.Contracts.csproj
```diff
@@ -70,6 +70,9 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Logging">
+      <Version>1.0.0</Version>
```
joelverhagen: Please verify that this is the same version we use in NuGet.Services.Logging.

joelverhagen: (I've seen 1.1.2 in the graph elsewhere before)

zhhyu:
Drop it now.
For NuGet.Jobs, move to use the minimal effort to fix the telemetry problem:
https://github.com/NuGet/NuGet.Jobs/pull/815

PR #312, opened 2019-09-16
Add batch signing for the output assemblies

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1821.


build/sign-binaries.proj
```diff
@@ -0,0 +1,47 @@
...
+
+  <PropertyGroup>
+    <RepositoryRootDirectory>$([System.IO.Path]::GetDirectoryName($(MSBuildProjectDirectory)))\</RepositoryRootDirectory>
+    <IntermediateOutputPath>$(RepositoryRootDirectory)artifacts\sign\obj\</IntermediateOutputPath>
+    <OutDir>$(RepositoryRootDirectory)</OutDir>
+    <SignTargetsDependOn>BatchSign</SignTargetsDependOn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <SolutionProjects Include="$(RepositoryRootDirectory)src\**\*.csproj" />
```
loic-sharma: This is the extension point other repositories should use, right? Might be worth adding a quick comment.

build/FindDuplicateFiles.cs
```diff
@@ -0,0 +1,311 @@
...
+
+                infos.Add(info);
+            }
+
+            return infos;
+        }
+
+        private Dictionary<string, List<TaskItemInfo>> FindDuplicates(List<TaskItemInfo> infos)
+        {
+            var fileSizeToInfo = GetFileSizeToInfo(infos);
```
loic-sharma: Super nit: pluralize infos to match the `infos` parameter?

```suggestion
            var fileSizeToInfos = GetFileSizeToInfos(infos);
```

build/FindDuplicateFiles.cs
```diff
@@ -0,0 +1,311 @@
...
+                else
+                {
+                    infosWithSameFileSize.Add(info);
+                }
+            }
+
+            return fileSizeToInfo;
+        }
+
+        private void BreakTiesWithLeadingHash(
```
loic-sharma: Does the leading hash optimization noticeably help performance? I'd imagine we don't have that many files with the exact same size, maybe falling back to full hash is fine?

build/sign-binaries.proj
joelverhagen: I moved the extension point to the PowerShell method `Sign-Binaries` so it much more clear now.

build/FindDuplicateFiles.cs
joelverhagen: I copied the logic of the SO link mentioned in the code. I didn't want to perf test so I blindly copied an Internet Person's prescription. You may be totally right here but my intuition was that maybe two binaries built at different times might have the same file size but different hashes and this leading hash could maybe catch that. It's a wild guess but since this is build infra I didn't want to invest too much time perf testing and just took a popular SO idea.

build/sign-binaries.proj
loic-sharma: Yup looks great

PR #316, opened 2019-10-03
Add traces to emit telemetry initialization and heartbeat config

xavierdecoster:
Facilitates debugging potential AI config issues. If `HeartbeatManager` is not configured/found, this will clearly tell us.

### DIFF
--- a/src/NuGet.Services.Logging/ApplicationInsights.cs
+++ b/src/NuGet.Services.Logging/ApplicationInsights.cs
@@ -40,7 +40,13 @@ namespace NuGet.Services.Logging
                     if (heartbeatManager != null)
                     {
                         heartbeatManager.HeartbeatInterval = heartbeatInterval.Value;
+
+                        Trace.TraceInformation($"Telemetry initialized using configured heartbeat interval: {heartbeatInterval.Value}.");
                     }
+                }
+                else
+                {
+                    Trace.TraceInformation($"Telemetry initialized using default heartbeat interval.");
                 }
 
                 Initialized = true;


src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -40,8 +40,14 @@ private static void InitializeTelemetryConfiguration(string instrumentationKey,
                     if (heartbeatManager != null)
                     {
                         heartbeatManager.HeartbeatInterval = heartbeatInterval.Value;
+
+                        Trace.TraceInformation($"Telemetry initialized using configured heartbeat interval: {heartbeatInterval.Value}.");
                     }
                 }
+                else
+                {
+                    Trace.TraceInformation($"Telemetry initialized using default heartbeat interval.");
```
skofman1: Can you print the default?

xavierdecoster: I wanted to but it's an _internal_ implementation detail of the AppInsights SDK (which may be subject to change?): https://github.com/microsoft/ApplicationInsights-dotnet/blob/f181abbcdf6a6a2a4aea821ce1d72e98b0c68fbe/src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatProvider.cs#L22

Would rather not hard-code a value in the trace message here to avoid it becoming out of sync if the value does change in the SDK.

src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -40,8 +40,14 @@ private static void InitializeTelemetryConfiguration(string instrumentationKey,
                     if (heartbeatManager != null)
                     {
                         heartbeatManager.HeartbeatInterval = heartbeatInterval.Value;
+
+                        Trace.TraceInformation($"Telemetry initialized using configured heartbeat interval: {heartbeatInterval.Value}.");
```
joelverhagen: We we have to use `Trace.*` here? why not log to AI?

xavierdecoster: We're initializing the AI telemetry configuration here and don't have access to a logger. Unless we already create a `TelemetryClient` in this code, but note that additional AI initialization may happen after this method was called.

joelverhagen: So how will we see these logs?

xavierdecoster: I believe TraceListener will pick them up and send them to AI :)
Perhaps not a bad idea to use TelemetryClient directly then to avoid another dependency to TraceListener?

xavierdecoster: I'm leaning towards keeping this as-is in this class. We always revisit.

PR #318, opened 2019-10-15
Update Microsoft.Extensions packages

ryuyu:
Update references to Microsoft.Extensions packages from 1.x -> 2.x
Also retargets net472

### DIFF
--- a/src/NuGet.Services.Configuration/NonCachingOptionsSnapshot.cs
+++ b/src/NuGet.Services.Configuration/NonCachingOptionsSnapshot.cs
@@ -36,7 +36,7 @@ namespace NuGet.Services.Configuration
 
         public TOptions Get(string name)
         {
-            return _value;
+            throw new NotImplementedException();
         }
     }
 }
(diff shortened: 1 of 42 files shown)


src/NuGet.Services.Configuration/NonCachingOptionsSnapshot.cs
```diff
@@ -33,5 +33,10 @@ public NonCachingOptionsSnapshot(IEnumerable<IConfigureOptions<TOptions>> setups
         }
 
         public TOptions Value => _value;
+
+        public TOptions Get(string name)
+        {
+            return _value;
```
agr: It should throw `NotImplementedException`.

joelverhagen: I agree with @agr, unless we fully understand the contract we should throw `NotImplementedException`. This implementation seems fishy since we ignore `name`.

ryuyu: Changed.

PR #321, opened 2019-11-19
Work-around heartbeat interval initialization bug in AI

xavierdecoster:
Verified changes locally against the solution that reproduced the AppInsights bug, and this work-around works.

Also need to verify our usage of `TelemetryConfiguration.Active` throughout the code consuming this library, as that API is being deprecated, and `TelemetryConfiguration.Active` won't pick up this custom initialized instance.

### DIFF
--- a/src/NuGet.Services.Logging/ApplicationInsights.cs
+++ b/src/NuGet.Services.Logging/ApplicationInsights.cs
@@ -9,69 +9,85 @@ using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
 namespace NuGet.Services.Logging
 {
+    /// <summary>
+    /// Utility class to initialize an <see cref="ApplicationInsightsConfiguration"/> instance
+    /// using provided instrumentation key, optional heartbeat interval, 
+    /// and, if detected, taking into account an optional ApplicationInsights.config file.
+    /// </summary>
+    /// <remarks>
+    /// Calling <see cref="Initialize(string)"/> or <see cref="Initialize(string, TimeSpan)"/> returns the
+    /// initialized <see cref="ApplicationInsightsConfiguration"/> object; 
+    /// it does not set the obsolete <see cref="TelemetryClient.Active"/> component.
+    /// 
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -16,37 +16,54 @@ public static class ApplicationInsights
 
         public static bool Initialized { get; private set; }
 
-        public static void Initialize(string instrumentationKey)
+        public static TelemetryConfiguration Initialize(string instrumentationKey)
```
xavierdecoster: Updated API to return the `TelemetryConfiguration` instance, as `TelemetryConfiguration.Active` is being deprecated in a future version of AppInsights.

See https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152

xavierdecoster:
Updated the PR with:
* more API docs and comments
* removed static state (call-site responsibility to handle the returned state)
* updated API to return new `ApplicationInsightsConfiguration` object holding handles to the initialized `TelemetryConfiguration` and `DiagnosticsTelemetryModule`

src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -5,98 +5,89 @@
...
+    /// and, if detected, taking into account an optional ApplicationInsights.config file.
+    /// </summary>
+    /// <remarks>
+    /// Calling <see cref="Initialize(string)"/> or <see cref="Initialize(string, TimeSpan)"/> returns the
+    /// initialized <see cref="ApplicationInsightsConfiguration"/> object; 
+    /// it does not set the obsolete <see cref="TelemetryClient.Active"/> component.
+    /// 
+    /// It is the caller's responsibility to ensure the returned configuration is used 
+    /// when creating new <see cref="TelemetryClient"/> instances.
+    /// </remarks>
```
xavierdecoster: Source: https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152#issuecomment-518742922

PR #323, opened 2019-12-04
Make ApplicationInsightsConfiguration implement IDisposable

xavierdecoster:
See https://github.com/NuGet/NuGet.Jobs/pull/837#discussion_r353380789

### DIFF
--- a/src/NuGet.Services.Logging/ApplicationInsightsConfiguration.cs
+++ b/src/NuGet.Services.Logging/ApplicationInsightsConfiguration.cs
@@ -8,6 +8,7 @@ using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 namespace NuGet.Services.Logging
 {
     public sealed class ApplicationInsightsConfiguration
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.Logging/ApplicationInsightsConfiguration.cs
```diff
@@ -32,5 +33,11 @@ public sealed class ApplicationInsightsConfiguration
         /// Allows tweaking Application Insights heartbeat telemetry.
         /// </summary>
         public DiagnosticsTelemetryModule DiagnosticsTelemetryModule { get; }
+
+        public void Dispose()
```
joelverhagen: Who's responsibility is it to call this dispose?

xavierdecoster: The caller of `ApplicationInsights.Initialize`, which typically is the application host that will configure DI and handle lifetime of those objects.

PR #325, opened 2019-12-18
Ensure all TelemetryModules are initialized from applicationinsights.config

xavierdecoster:
Related to findings in https://github.com/microsoft/ApplicationInsights-dotnet/issues/1503.


src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -65,6 +66,13 @@ public static ApplicationInsightsConfiguration Initialize(string instrumentation
 
             telemetryConfiguration.TelemetryInitializers.Add(new TelemetryContextInitializer());
 
+            // Hook-up the TelemetryModules configured in applicationinsights.config into our own
+            // TelemetryConfiguration instance, as this doesn't happen automatically...
+            foreach (var telemetryModule in TelemetryModules.Instance.Modules)
+            {
+                telemetryModule.Initialize(telemetryConfiguration);
```
joelverhagen: Is `Initialized` idempotent? I'm wondering if a change is made in the future and the initialization starts happening automatically will this become a weird bug?

xavierdecoster: That's a question the AppInsights SDK folks should be able to answer. There's not much docs on this API: https://docs.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.extensibility.itelemetrymodule.initialize?view=azure-dotnet

xavierdecoster: If the loading is happening automatically in the future, we should indeed remove this particular code... Though I don't think it would be an issue if the modules get initialized twice using the same configuration.

joelverhagen:
Could we take this change to gallery and a couple of jobs in DEV with a private build to make sure everything is looking okay?

xavierdecoster:
Confirmed it does fix dependency and request telemetry for gallery.

PR #327, opened 2019-12-20
Upgrade Serilog and ensure logging is formatted as Traces

xavierdecoster:


src/NuGet.Services.Logging/LoggingSetup.cs
```diff
@@ -40,19 +40,13 @@ public static LoggerConfiguration CreateDefaultLoggerConfiguration(bool withCons
...
-                // Even though this method call is marked [Obsolete],
-                // there's currently no other way to pass in the active TelemetryConfiguration as configured in DI.
-                // These SeriLog APIs are very likely to change to support passing in the TelemetryConfiguration again.
-                // See also https://github.com/serilog/serilog-sinks-applicationinsights/issues/121.
-
-#pragma warning disable CS0618 // Type or member is obsolete
                 loggerConfiguration = loggerConfiguration.WriteTo.ApplicationInsights(
-                    telemetryConfiguration, 
+                    telemetryConfiguration,
+                    TelemetryConverter.Traces,
```
xavierdecoster: @agr this may be the cause of logs to be sent as `customEvents` instead of `traces`. Tracked it down to `ILogger` messages only being subject to that weirdness, which are set up using Serilog here.

PR #329, opened 2020-01-08
Move and consolidate TelemetryInitializers from NuGet.Jobs

xavierdecoster:
This moves the `JobNameTelemetryInitializer`, `InstanceNameTelemetryInitializer`, and `JobPropertiesTelemetryInitializer` from `NuGet.Jobs` to the `NuGet.Services.Logging` package, and consolidates them into a single `ITelemetryInitializer`.

This will help us standardize job telemetry enrichment by sharing the telemetry initializer through this package instead of having it only available within `ng.exe`, and we can enrich telemetry within a single initializer instead of 3 separate ones (given they're used together).

When consumed in `NuGet.Jobs`, we can clean up the code and tests in that repo.


loic-sharma:
Should this be targeting the `dev` branch?

PR #330, opened 2020-01-09
Port MachineNameTelemetryInitializer from NuGet.ApplicationInsights.Owin

xavierdecoster:
This type belongs in `NuGet.Services.Logging` as it can be generally reused and is not specific to `Owin`.

In addition, there's no need to retrieve the machine name from `Environment.MachineName` on every telemetry initialization, which may happen multiple times for the same telemetry item.

### DIFF
--- a/src/NuGet.Services.Logging/TelemetryInitializers/MachineNameTelemetryInitializer.cs
+++ b/src/NuGet.Services.Logging/TelemetryInitializers/MachineNameTelemetryInitializer.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.Logging
+{
+    public class MachineNameTelemetryInitializer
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.Logging/TelemetryInitializers/MachineNameTelemetryInitializer.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class MachineNameTelemetryInitializer
+        : SupportPropertiesTelemetryInitializer
+    {
+        public MachineNameTelemetryInitializer()
+            : base("MachineName", TryGetMachineName())
+        {
+        }
+
+        private static string TryGetMachineName()
```
xavierdecoster: Current version lives here: Current version lives here: https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/NuGet.ApplicationInsights.Owin/MachineNameTelemetryInitializer.cs

Note that we now inherit from `SupportPropertiesTelemetryInitializer` which uses the proper AppInsights API to set properties. The current version in `NuGet.ApplicationInsights.Owin` still uses the obsolete API via `TelemetryContext`.

PR #334, opened 2020-02-05
Update KeyVault to use Managed Identities

shishirx34:
This is needed to be consumed in the code and in to be used in Jobs.Common

### DIFF
--- a/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
@@ -22,6 +22,7 @@ namespace NuGet.Services.Configuration
         public ConfigurationRootSecretReaderFactory(IConfigurationRoot config)
         {
             _vaultName = config[Constants.KeyVaultVaultNameKey];
+
             string useManagedIdentity = config[Constants.KeyVaultUseManagedIdentity];
             if (!string.IsNullOrEmpty(useManagedIdentity))
             {
@@ -53,17 +54,30 @@ namespace NuGet.Services.Configuration
                 return new EmptySecretReader();
             }
 
-            var certificateConfig = new CertificateConfiguration(_storeName,
-                _storeLocation,
+            KeyVaultConfiguration keyVaultConfiguration;
+
+            if (_useManagedIdentity)
+            {
+                keyVaultConfiguration = new KeyVaultConfiguration(_vaultName);
+            }
+            else
+            {
+                var certificate = CertificateUtility.FindCertificateByThumbprint(
+                    !string.IsNullOrEmpty(_storeName)
+                    ? (StoreName)Enum.Parse(typeof(StoreName), _storeName)
+                    : StoreName.My,
+                !string.IsNullOrEmpty(_storeLocation)
+                    ? (StoreLocation)Enum.Parse(typeof(StoreLocation), _storeLocation)
+                    : StoreLocation.LocalMachine,
                 _certificateThumbprint,
                 _validateCertificate);
 
-            var keyVaultConfiguration = new KeyVaultConfiguration(
-                _vaultName,
-                _useManagedIdentity,
-                _clientId,
-                certificateConfig,
-                _sendX5c);
+                keyVaultConfiguration = new KeyVaultConfiguration(
+                    _vaultName,                
+                    _clientId,
+                    certificate,
+                    _sendX5c);
+            }
 
             return new KeyVaultReader(keyVaultConfiguration);
         }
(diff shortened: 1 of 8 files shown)


joelverhagen:
UTs?

shishirx34:
I will be adding the tests here.

src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -9,22 +9,60 @@ namespace NuGet.Services.KeyVault
...
-        public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2 certificate, bool sendX5c=false)
+        public bool UseManagedIdentity { get; }
+        public string ClientId { get; private set; }
+        public X509Certificate2 Certificate { get; private set; }
+        public bool SendX5c { get; private set; }
+
+        /// <summary>
+        /// The constructor for keyvault configuration when using managed identities or certificate
+        /// </summary>
+        public KeyVaultConfiguration(string vaultName,
```
joelverhagen: FYI, it looks like this constructor will be used directly by gallery and status:

Status:
https://github.com/NuGet/NuGet.Services.Status/blob/7ef36169174d260abc6d023ed130322c3cfda170/src/NuGet.Status/Configuration/SecretReaderFactory.cs#L57-L61

Gallery:
https://github.com/NuGet/NuGetGallery/blob/08758f6582a1de098cfe725d5c0944f86252486b/src/NuGetGallery.Services/Configuration/SecretReader/SecretReaderFactory.cs#L55-L57

I'll be doing this change and it looks like it should work, but I wanted to call out these ones don't go the `IConfigurationRoot` route you implemented above for jobs. Azure Search service works like jobs in this way so that's good.

src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -9,22 +9,60 @@ namespace NuGet.Services.KeyVault
...
+            bool useManagedIdentity,
```
joelverhagen: Not sure if I understand this overload. When would you specify `useManagedIdentity = true` and also send in non-null `clientId` or `certConfig`? In other words this overload seems to have some non-sense combinations. Couldn't this overload just have `vaultName` and internally it sets `UseManagedIdentity` to true.

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
```diff
@@ -21,6 +22,12 @@ public class ConfigurationRootSecretReaderFactory : ISecretReaderFactory
         public ConfigurationRootSecretReaderFactory(IConfigurationRoot config)
         {
             _vaultName = config[Constants.KeyVaultVaultNameKey];
+            string useManagedIdentity = config[Constants.KeyVaultUseManagedIdentity];
+            if (!string.IsNullOrEmpty(useManagedIdentity))
+            {
+                _useManagedIdentity = bool.Parse(useManagedIdentity);
+            }
+
             _clientId = config[Constants.KeyVaultClientIdKey];
```
skofman1: The config should be either Managed Identity or certificate. If we allow both it would be confusing. So if _useManagedIdentity = true, we should error if certificate is provided.

src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
skofman1: Agree with Joel. Consider separate ctors for certificate case and for Managed Identity case.

---
In reply to: [375568212](https://github.com/NuGet/ServerCommon/pull/334#discussion_r375568212) [](ancestors = 375568212)

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -4,13 +4,14 @@
...
 using Microsoft.Azure.KeyVault;
+using Microsoft.Azure.Services.AppAuthentication;
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 
 namespace NuGet.Services.KeyVault
 {
     /// <summary>
     /// Reads secretes from KeyVault.
-    /// Authentication with KeyVault is done using a certificate in location:LocalMachine store name:My 
+    /// Authentication with KeyVault is done using either a managed identity or a certificate in location:LocalMachine store name:My
```
skofman1: > certificate in location:LocalMachine store name:My  [](start = 81, length = 52)

This comment is no longer correct.. we pass cert location as a param :)

src/NuGet.Services.KeyVault/CertificateUtility.cs
```diff
@@ -6,8 +6,33 @@
 
 namespace NuGet.Services.KeyVault
 {
+    public class CertificateConfiguration
```
skofman1: Great refactoring!

src/NuGet.Services.KeyVault/KeyVaultReader.cs
shishirx34: I have reverted the ctors to previous ones.. so now its correct again :D

src/NuGet.Services.KeyVault/CertificateUtility.cs
shishirx34: umm.. 😕 I've removed this, since its not really useful as we have separate ctors for each scenario. I like the refactoring but it will be work to propagate this to the client libraries unlikely it will happen.

PR #335, opened 2020-03-03
[ECR] Populate logger to have better telemetry

zhhyu:
Currently we don't have the logs around the caching control of secret reader.
In order to understand the on-line behavior of certificates (SQL), try to populate the logger to the corresponding parts.

### DIFF
--- a/src/NuGet.Services.KeyVault/CachingSecretReader.cs
+++ b/src/NuGet.Services.KeyVault/CachingSecretReader.cs
@@ -1,16 +1,17 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Concurrent;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.KeyVault
 {
     public class CachingSecretReader : ISecretReader
     {
         public const int DefaultRefreshIntervalSec = 60 * 60 * 24; // 1 day
-        public const int DefaultRefreshIntervalBeforeExpirySec = 60 * 30 ; // 30 minutes 
+        public const int DefaultRefreshIntervalBeforeExpirySec = 60 * 30 ; // 30 minutes
 
         private readonly ISecretReader _internalReader;
         private readonly ConcurrentDictionary<string, CachedSecret> _cache;
@@ -18,7 +19,9 @@ namespace NuGet.Services.KeyVault
         //_refreshIntervalBeforeExpiry specifies the timespan before secret expiry to refresh the secret value.
         private readonly TimeSpan _refreshIntervalBeforeExpiry;
 
-        public CachingSecretReader(ISecretReader secretReader, int refreshIntervalSec = DefaultRefreshIntervalSec, int refreshIntervalBeforeExpirySec = DefaultRefreshIntervalBeforeExpirySec)
+        public CachingSecretReader(ISecretReader secretReader,
+            int refreshIntervalSec = DefaultRefreshIntervalSec,
+            int refreshIntervalBeforeExpirySec = DefaultRefreshIntervalBeforeExpirySec)
         {
             _internalReader = secretReader ?? throw new ArgumentNullException(nameof(secretReader));
             _cache = new ConcurrentDictionary<string, CachedSecret>();
@@ -26,12 +29,22 @@ namespace NuGet.Services.KeyVault
             _refreshIntervalBeforeExpiry = TimeSpan.FromSeconds(refreshIntervalBeforeExpirySec);
         }
 
-        public async Task<string> GetSecretAsync(string secretName)
+        public Task<string> GetSecretAsync(string secretName)
         {
-            return (await GetSecretObjectAsync(secretName)).Value;            
+            return GetSecretAsync(secretName, logger: null);
         }
 
-        public async Task<ISecret> GetSecretObjectAsync(string secretName)
+        public async Task<string> GetSecretAsync(string secretName, ILogger logger)
+        {
+            return (await GetSecretObjectAsync(secretName, logger)).Value;
+        }
+
+        public Task<ISecret> GetSecretObjectAsync(string secretName)
+        {
+            return GetSecretObjectAsync(secretName, logger: null);
+        }
+
+        public async Task<ISecret> GetSecretObjectAsync(string secretName, ILogger logger)
         {
             if (string.IsNullOrEmpty(secretName))
             {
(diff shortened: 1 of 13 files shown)


src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -44,8 +57,16 @@ public async Task<ISecret> GetSecretObjectAsync(string secretName)
             {
                 return result.Secret;
             }
+
+            var start = DateTimeOffset.Now;
```
xavierdecoster: Shouldn't this always be UTC time?

```suggestion
            var start = DateTimeOffset.UtcNow;
```

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -44,8 +57,16 @@ public async Task<ISecret> GetSecretObjectAsync(string secretName)
...
             // The cache does not contain a fresh copy of the secret. Fetch and cache the secret.
             var updatedValue = new CachedSecret(await _internalReader.GetSecretObjectAsync(secretName));
+
+            logger?.LogInformation("Refresh secret {SecretName}{ExpirationTime} (Latency = {ElapsedMilliseconds}ms)",
+                updatedValue.Secret.Name,
+                updatedValue.Secret.Expiration == null ? "" : " which expires at " + ((DateTimeOffset) updatedValue.Secret.Expiration).DateTime,
+                (DateTimeOffset.Now - start).TotalMilliseconds.ToString("F2"));
```
xavierdecoster: Same here

zhhyu: Yeah, that's a good point!
The time difference here is to track the time of the network call to KeyVault server, in order for us to understand the online time cost and further performance improvement, so the functionality is basically the same here.
However, we want to track the expiration time of the secret, in order to identify the version of the secret, so I update the Expiration time to use "UtcDateTime".

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -44,8 +57,16 @@ public async Task<ISecret> GetSecretObjectAsync(string secretName)
             {
                 return result.Secret;
             }
+
+            var start = DateTimeOffset.UtcNow;
             // The cache does not contain a fresh copy of the secret. Fetch and cache the secret.
             var updatedValue = new CachedSecret(await _internalReader.GetSecretObjectAsync(secretName));
+
+            logger?.LogInformation("Refresh secret {SecretName}{ExpirationTime} (Latency = {ElapsedMilliseconds}ms)",
```
skofman1: nit: please phrase more clearly.
Refreshing secret {SecretName}. Expiring at:{ExpirationTime}. Took: {ElapsedMilliseconds} ms)

skofman1:
I thought you also wanted to log the certificate we use to authenticate with SQL. Is this going to be a separate PR?

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -1,37 +1,50 @@
...
             _cache = new ConcurrentDictionary<string, CachedSecret>();
             _refreshInterval = TimeSpan.FromSeconds(refreshIntervalSec);
             _refreshIntervalBeforeExpiry = TimeSpan.FromSeconds(refreshIntervalBeforeExpirySec);
         }
 
-        public async Task<string> GetSecretAsync(string secretName)
+        public Task<string> GetSecretAsync(string secretName)
         {
-            return (await GetSecretObjectAsync(secretName)).Value;            
+            return GetSecretAsync(secretName, logger: null);
```
loic-sharma: Please make this method `async` and `await` the result of `GetSecretAsync`. For reasons why you should await, see: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

loic-sharma: Same comment for `GetSecretObjectAsync` on line 44

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -44,8 +57,16 @@ public async Task<ISecret> GetSecretObjectAsync(string secretName)
...
             }
+
+            var start = DateTimeOffset.UtcNow;
             // The cache does not contain a fresh copy of the secret. Fetch and cache the secret.
             var updatedValue = new CachedSecret(await _internalReader.GetSecretObjectAsync(secretName));
+
+            logger?.LogInformation("Refresh secret {SecretName}{ExpirationTime} (Latency = {ElapsedMilliseconds}ms)",
+                updatedValue.Secret.Name,
+                updatedValue.Secret.Expiration == null ? "" : " which expires at " + ((DateTimeOffset) updatedValue.Secret.Expiration).UtcDateTime,
+                (DateTimeOffset.UtcNow - start).TotalMilliseconds.ToString("F2"));
```
loic-sharma: We can search and analyze the log values using App Insights. One may want to do some number crunching on the `ExpirationTime` parameter, but this is difficult to do if the values are prefixed with `which expired at `. Maybe something like this instead:

```suggestion
            logger?.LogInformation("Refreshed secret {SecretName} in {ElapsedMilliseconds}ms, expires at {ExpirationTime}",
                updatedValue.Secret.Name,
                (DateTimeOffset.UtcNow - start).TotalMilliseconds.ToString("F2"),
                updatedValue.Secret.Expiration == null
                    ? DateTimeOffset.MaxValue
                    : ((DateTimeOffset) updatedValue.Secret.Expiration).UtcDateTime);
```

src/NuGet.Services.KeyVault/CachingSecretReader.cs
loic-sharma: I wouldn't worry about this for the `EmptySecretReader` though

src/NuGet.Services.KeyVault/CachingSecretReader.cs
```diff
@@ -1,37 +1,50 @@
...
 
-        public async Task<string> GetSecretAsync(string secretName)
+        public Task<string> GetSecretAsync(string secretName)
         {
-            return (await GetSecretObjectAsync(secretName)).Value;            
+            return GetSecretAsync(secretName, logger: null);
         }
 
-        public async Task<ISecret> GetSecretObjectAsync(string secretName)
+        public async Task<string> GetSecretAsync(string secretName, ILogger logger)
```
loic-sharma: Can we pass the `ILogger` through a new constructor that accepts an `ILogger` argument? I understand that your approach is easier today, but I'd like to understand why you chose your approach a little better given that it adds tech debt.

zhhyu:
> I thought you also wanted to log the certificate we use to authenticate with SQL. Is this going to be a separate PR?

KeyVault gives us the plaintext of the certificate, and I prefer not to log the certificate, because it's a secret. The "SecretName" will match the exact name in the KeyVault, and the expiration time will denote the version of the certificate. So the combination of these two will be enough for us to identify the SQL certificate.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
zhhyu: It's a simple passthrough, so that's why I don't use "await" at the first time. (https://blog.stephencleary.com/2016/12/eliding-async-await.html)
But anyway, I have updated to use "await" as there may not be an important performance gain there.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
zhhyu: Yeah, I have updated the log info, and I may prefer to log "null" rather than DateTimeOffset.MaxValue for null expiration time.

src/NuGet.Services.KeyVault/CachingSecretReader.cs
zhhyu: Yeah, that's a good point. And at the first time, I also feel that it's better to pass the logger at the constructor. But then I find that it will lead to a bigger code refactoring. We have a high cohesion around this and directly build the instance of the secret reader at many places. 
Like Gallery DI initialization, now we build the secret reader/injector before the logger because the initialization of the logger needs the injector.
I prefer to keep this now and if it hurts in the future, we can consider refactoring around this.

loic-sharma: Makes sense, thanks for the info!

PR #336, opened 2020-03-11
[ECR-SQL] Enable the jobs to achieve SQL certificates update and provide better telemetry

zhhyu:
Use the very similar way like Gallery and Monitoring runner to enable the jobs to achieve the active SQL certificates update and provide us the logs about the on-line status of the SQL certificates.
After this one is merged and flowed to the job repository, it will fix most of the jobs which depend on JobBase, JsonConfigurationJob, SubcriptionProcessorJob, such as validation jobs, Gallery jobs, Azure search jobs and etc.

Verify locally under:
Validation.Orchestrator,
CredentialExpiration.

Will send another PR to fix the V3 jobs (Db2Catalog, Db2Monitoring, Monitoring Processor) because of using another job infrastructure.


src/NuGet.Services.Configuration/NotInjectedConfigurations.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.Services.Configuration
+{
+    public static class NotInjectedConfigurations
+    {
+        public static readonly HashSet<string> Keys = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
```
joelverhagen: This can be in a private `static readonly` in `KeyVaultInjectingConfigurationProvider` -- or do you plan on using it elsewhere?

zhhyu: Yeah, I have moved to one place. Previously, I feel that this could be reused at other places. But when I try to fix the Ng jobs, I find we use a different mechanism. And once we read json configuration file, we will use that extension method to inject the secret. In addition, the key is different, especially for Json configuration.

PR #337, opened 2020-03-12
[ECR-SQL] Update "SecretDictionary" not to inject the connection string for NgJobs. 

zhhyu:
NgJobs use the "SecretDictionary" to inject the connection string from the arguments. We should not inject the connection string until we pass it to the SQL connection factory like other fixes.
Update the "SecretDictionary" not to inject the connection string.

### DIFF
--- a/src/NuGet.Services.Configuration/SecretDictionary.cs
+++ b/src/NuGet.Services.Configuration/SecretDictionary.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using NuGet.Services.KeyVault;
 
 namespace NuGet.Services.Configuration
@@ -13,7 +14,7 @@ namespace NuGet.Services.Configuration
     {
         private readonly ISecretInjector _secretInjector;
         private readonly IDictionary<string, string> _unprocessedArguments;
-        private readonly HashSet<string> _notInjectedKeys = new HashSet<string>();
+        private readonly HashSet<string> _notInjectedKeys;
 
         public SecretDictionary(ISecretInjector secretInjector, IDictionary<string, string> unprocessedArguments,
             HashSet<string> notInjectedKeys) : this(secretInjector, unprocessedArguments)
@@ -25,6 +26,7 @@ namespace NuGet.Services.Configuration
         {
             _secretInjector = secretInjector;
             _unprocessedArguments = unprocessedArguments;
+            _notInjectedKeys = new HashSet<string>();
         }
 
         public string this[string key]
@@ -43,7 +45,7 @@ namespace NuGet.Services.Configuration
 
         public ICollection<string> Values => _unprocessedArguments.Select(p => InjectOrSkip(p.Key, p.Value)).ToList();
 
-        public class SecretEnumerator : IEnumerator<KeyValuePair<string, string>>
+        private class SecretEnumerator : IEnumerator<KeyValuePair<string, string>>
         {
             private readonly Func<string, string, string> _injectOrSkipFunc;
             private readonly IList<KeyValuePair<string, string>> _unprocessedPairs;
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -12,42 +13,54 @@ public class SecretDictionary : IDictionary<string, string>
     {
         private readonly ISecretInjector _secretInjector;
         private readonly IDictionary<string, string> _unprocessedArguments;
+        private static readonly HashSet<string> _notInjectedKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
```
joelverhagen: Can this be a parameter to the `SecretInjector` instead of hardcoded?

src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -12,42 +13,54 @@ public class SecretDictionary : IDictionary<string, string>
...
+            };
 
         public SecretDictionary(ISecretInjector secretInjector, IDictionary<string, string> unprocessedArguments)
         {
             _secretInjector = secretInjector;
             _unprocessedArguments = unprocessedArguments;
         }
 
-        private string Inject(string key)
+        private string Inject(string value)
```
skofman1: private methods should be placed under public methods in the class

src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -12,42 +13,54 @@ public class SecretDictionary : IDictionary<string, string>
...
         {
-            return _secretInjector.InjectAsync(key).Result;
+            return _secretInjector.InjectAsync(value).Result;
+        }
+
+        private string Inject(string key, string value)
```
skofman1: >Inject [](start = 23, length = 6)

the name of the method implies that injection always happens. But it's not the case. Consider renaming to InjectOrSkip

src/NuGet.Services.Configuration/SecretDictionary.cs
skofman1: an optional param to ctor will be great

---
In reply to: [391710346](https://github.com/NuGet/ServerCommon/pull/337#discussion_r391710346) [](ancestors = 391710346)

src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -12,44 +13,46 @@ public class SecretDictionary : IDictionary<string, string>
     {
         private readonly ISecretInjector _secretInjector;
         private readonly IDictionary<string, string> _unprocessedArguments;
+        private readonly HashSet<string> _notInjectedKeys = new HashSet<string>();
```
joelverhagen: nit: initialize this either only in the constructor or only with a field initializer as you have here. When the constructor with the `notInjectedKeys` parameter is called, we unnecessarily initialize a new hash set then immediately replace it with the provided one.

src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -12,44 +13,46 @@ public class SecretDictionary : IDictionary<string, string>
...
             var isFound = _unprocessedArguments.TryGetValue(key, out unprocessedValue);
-            value = isFound ? Inject(unprocessedValue) : null;
+            value = isFound ? InjectOrSkip(key, unprocessedValue) : null;
             return isFound;
         }
 
-        public ICollection<string> Values => _unprocessedArguments.Values.Select(Inject).ToList();
+        public ICollection<string> Values => _unprocessedArguments.Select(p => InjectOrSkip(p.Key, p.Value)).ToList();
 
         public class SecretEnumerator : IEnumerator<KeyValuePair<string, string>>
```
joelverhagen: This can be `private` right? In general public types with `Func` or `Action` parameters are hard to understand.

src/NuGet.Services.Configuration/SecretDictionary.cs
```diff
@@ -113,5 +116,18 @@ public void CopyTo(KeyValuePair<string, string>[] array, int arrayIndex)
...
+        private string InjectOrSkip(string key, string value)
+        {
+            if (!_notInjectedKeys.Contains(key))
+            {
+                return Inject(value);
+            }
+            return value;
+        }
+
+        private string Inject(string value)
```
joelverhagen: nit: do `.Result` (or `Wait()`) at the highest level possible

PR #338, opened 2020-03-12
Add StringArrayConverter to handle string arrays in Gallery-style config

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/3020

We have some internal configuration that we want to pull into the private build of gallery that has an array configuration variable. The way we do configuration in gallery is via `TypeConverter`/component model stuff and our config comes from string to string pairs. Therefore if the configuration model has a string array in it, this is how we handle it.

Note that we already have the implementation here in gallery:
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Services/Helpers/StringArrayConverter.cs

However I'd like to move it to a common place so that an internal project that gallery will depend on with build magic can use it too.

### DIFF
--- a/tests/NuGet.Services.Configuration.Tests/StringArrayConverterFacts.cs
+++ b/tests/NuGet.Services.Configuration.Tests/StringArrayConverterFacts.cs
@@ -1,0 +1,56 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using Xunit;
+
+namespace NuGet.Services.Configuration.Tests
+{
+    public class StringArrayConverterFacts
+    {
+        [Fact]
+        public void CanConvertFromString()
+        {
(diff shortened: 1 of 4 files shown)


tests/NuGet.Services.Configuration.Tests/StringArrayConverterFacts.cs
```diff
@@ -0,0 +1,56 @@
...
+            var target = new StringArrayConverter();
+
+            var output = target.ConvertFrom(input);
+
+            var array = Assert.IsType<string[]>(output);
+            Assert.Equal(new[] { input }, array);
+        }
+
+        [Fact]
+        public void ReturnsEmptyArrayWithEmpty()
```
loic-sharma: Should we also have a test case for `null` values?

joelverhagen: `null` should never happen per the type converter contract. The `CanConvertFrom` will return `null` and the method will never be called.

PR #344, opened 2020-04-08
Move KnownOperationNameEnricher to a common place

joelverhagen:
This is moved from [NuGetGallery](https://github.com/NuGet/NuGetGallery/blob/2aaa5ce7f9be3b59ba2abff7f8295e4153d1e828/src/NuGetGallery/Telemetry/KnownOperationNameEnricher.cs) and made more general purpose.

This will be used in the search service.

### DIFF
--- a/src/NuGet.Services.Logging/TelemetryInitializers/KnownOperationNameEnricher.cs
+++ b/src/NuGet.Services.Logging/TelemetryInitializers/KnownOperationNameEnricher.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights.Channel;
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.Logging/TelemetryInitializers/KnownOperationNameEnricher.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var operationName = telemetry.Context?.Operation?.Name;
+            if (operationName == null)
+            {
+                return;
+            }
+
+            if (_knownOperations.Contains(operationName))
+            {
+                itemTelemetry.Properties[KnownOperation] = operationName;
```
loic-sharma, line 56: Should this have some quick unit tests?

joelverhagen: Yes, I sort of maybe forgot a little bit

joelverhagen: Fixed.

PR #347, opened 2020-10-05
Added commandlet to try finding path to preinstalled dotnet.exe

agr:
Need `dotnet` to run tests in some repos, added discovery.

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -543,6 +543,22 @@ Function Install-DotnetCLI {
     & $DotNetExe --info
 }
 
+Function Get-ExistingDotnet {
+    $p = Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\ASP.NET Core\Shared Framework"
+    if ($p.InstallDir) {
+        $path = Join-Path $p.InstallDir "dotnet.exe"
+        if (Test-Path $path) {
+            return $path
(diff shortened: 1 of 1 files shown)


build/common.ps1
```diff
@@ -543,6 +543,22 @@ Function Install-DotnetCLI {
     & $DotNetExe --info
 }
 
+Function Get-ExistingDotnet {
+    $p = Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\ASP.NET Core\Shared Framework"
```
joelverhagen, line 547: Is this the recommended approach from .NET CLI? The "ASP.NET Core" part looks a little suspicious/specific given the context

zhhyu (approved):

build/common.ps1
agr: I didn't find an "official prescription" for discovery. That's what I found to work on my and build machines we use. I will ask dotnet folks about it.

joelverhagen: Okay, if it works for us, don't sink too much time of the "proper". This can be easily fixed later if need be.

joelverhagen (approved):

agr:
Will imply that it is available through the `PATH` environment variable for now.

PR #350, opened 2020-11-10
Migrate to SDK-based projects

joelverhagen:
This is based off of great effort from @loic-sharma and @agr. 

This is in support of https://github.com/NuGet/Engineering/issues/3487, moving the Azure Search Service to .NET Core.

The purpose of this effort is to move all projects away from legacy .csproj to SDK-based .csproj. The next step will be adding .NET Standard support to as many projects as possible. No runtime behavior will be affected by this PR. 

I have validated this PR by builds and deployments of the jobs consuming ServerCommon libraries.


zhhyu (approved):

src/NuGet.Services.Owin/NuGet.Services.Owin.csproj
```diff
@@ -1,78 +1,18 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{5AF3CE0B-9C47-4FD0-83F6-D6C1E0A50E85}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>NuGet.Services.Owin</RootNamespace>
-    <AssemblyName>NuGet.Services.Owin</AssemblyName>
-    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
+    <TargetFramework>net472</TargetFramework>
+    <Description>Shared Owin code for the NuGet projects</Description>
```
loic-sharma, line 5: Super nit:

```suggestion
    <Description>Shared OWIN code for the NuGet projects</Description>
```

src/NuGet.Services.Validation/NuGet.Services.Validation.csproj
```diff
@@ -1,187 +1,12 @@
...
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>NuGet.Services.Validation</RootNamespace>
-    <AssemblyName>NuGet.Services.Validation</AssemblyName>
-    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
+    <TargetFramework>net472</TargetFramework>
+    <Description>Logic shared between the front-end and back-end concerning asynchronous validation</Description>
+
+    <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
```
loic-sharma, line 7: Consider adding a quick comment about this attribute

loic-sharma (approved):

nkolev92 (commented):
So beautiful! 🏆 🏆 🏆

src/NuGet.Services.AzureManagement/NuGet.Services.AzureManagement.csproj
```diff
@@ -1,85 +1,17 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{66CCFC9E-DA43-4DAB-9746-92B51AAC5F29}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>NuGet.Services.AzureManagement</RootNamespace>
-    <AssemblyName>NuGet.Services.AzureManagement</AssemblyName>
-    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <TargetFrameworkProfile />
+    <TargetFramework>net472</TargetFramework>
```
nkolev92, line 4: You will always target the same target framework, correct?

2 ideas to minimize your NETCore footprint later on. 

* Declare `TargetFramework` in a common import. 
* Use a variable for targetframework like client does: https://github.com/NuGet/NuGet.Client/blob/5a0d881d8cbb144de7bc0a03f40a6a633662d7d4/build/common.project.props#L14-L26 and https://github.com/NuGet/NuGet.Client/blob/5a0d881d8cbb144de7bc0a03f40a6a633662d7d4/src/NuGet.Core/NuGet.Common/NuGet.Common.csproj#L7

build.ps1
```diff
@@ -85,6 +85,12 @@ Invoke-BuildStep 'Set version metadata in AssemblyInfo.cs' { `
             "$PSScriptRoot\src\NuGet.Services.Validation\Properties\AssemblyInfo.g.cs"
             
         $versionMetadata | ForEach-Object {
+            # Ensure the directory exists before generating the version info file.
+            $directory = Split-Path $_
+            if (-not (Test-Path $directory)) {
```
agr, line 90: I think we can run `New-Item` unconditionally. It will not fail if the directory already exist.

agr (approved):

src/NuGet.Services.Validation/NuGet.Services.Validation.csproj
joelverhagen: There is a better option now. Switched to that.

src/NuGet.Services.AzureManagement/NuGet.Services.AzureManagement.csproj
joelverhagen: I don't think we're ready to align at the solution level yet. For this specific step, you're right. `net472` is everywhere. However my next step is moving things to cross-target `net472` and `netstandard`. Unfortunately, not all projects will be able to be support `netstandard` (due to dependency) and others will need `netstandard2.1` instead of `netstandard2.0` (EF 6 supports netstandard2.1). Therefore, I'd rather keep it at the project level for now. Hopefully we can move to your suggested model after we've fixed up dependency issues.

build.ps1
joelverhagen: Nice catch. Yes that works.

skofman1 (approved):
:shipit:

PR #351, opened 2020-11-10
Add .NET Standard support to some projects

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/3487.

The approach taken here was to add `netstandard2.0` support to as many projects as possible. In places where the dependency was not needed for running the Azure Search Service, I only added `netstandard2.0` support if it was very easy (i.e. no code or dependency changes required). I only made code or dependency changes if the library was really needed at runtime for the Azure Search Service.

I chose to keep the `net472` target because I wanted to reduce the potential for restore graph or runtime changes for entry points still running on .NET Framework (everything, at this point). In the future when we have a more wholistic move to .NET Core, we can consider dropping the .NET Framework target.

Cross-targeting also has the benefit that you can conditionally compile some source files based on target framework. This is a great tool that allows you to only ports classes that really need to move and leave behind ones that aren't use by your entry point that's moving to .NET Core. For example, `ExceptionTelemetryProcessor` in NuGet.Services.Logging is not used by the Azure Search Service so it can be excluded from the `netstandard2.0` target but it can't be entirely deleted since it is used by the gallery.

The full status, per project is:

- NuGet.Services.AzureManagement
  - ✔️ added `netstandard2.0`
- NuGet.Services.Build
  - ❌ no change since the MSBuild APIs it uses require .NET Framework 
- NuGet.Services.Configuration
  - ✔️ added `netstandard2.0`
- NuGet.Services.Contracts
  - ✔️ added `netstandard2.0`
- NuGet.Services.Cursor
  - ❌ no change due to incompatible dependency on NuGet.Services.Storage
- NuGet.Services.FeatureFlags 
  - ✔️ added `netstandard2.0`
- NuGet.Services.Incidents
  - ✔️ added `netstandard2.0`
- NuGet.Services.KeyVault
  - ✔️ added `netstandard2.0`
  - ⚠️ updated Microsoft.Azure.KeyVault which supports .NET Framework and .NET Core
- NuGet.Services.Licenses
  - ✔️ added `netstandard2.0`
- NuGet.Services.Logging
  - ✔️ added `netstandard2.0`
  - ⚠️ excluded `ExceptionTelemetryProcessor` from the `netstandard2.0` target since it requires .NET Framework
- NuGet.Services.Messaging
  - ❌ no change due to incompatible dependency on NuGet.Services.ServiceBus
- NuGet.Services.Messaging.Email
  - ❌ no change due to incompatible dependency on NuGet.Services.ServiceBus and AnglicanGeek.MarkdownMailer
- NuGet.Services.Owin
  - ❌ no change due to incompatible dependency on Microsoft.Owin
- NuGet.Services.ServiceBus
  - ❌ this one cannot move to `netstandard2.0` due to the WindowsAzure.ServiceBus dependency
- NuGet.Services.Sql
  - ✔️ added `netstandard2.0`
  - ⚠️ added System.Data.SqlClient on the `netstandard2.0` target since it's not included by default in .NET Core
- NuGet.Services.Status.Table
  - ✔️ added `netstandard2.0`
- NuGet.Services.Status
  - ✔️ added `netstandard2.0`
- NuGet.Services.Storage
  - ❌ no change due to unavailable synchronous APIs in WindowsAzure.Storage used in constructors
- NuGet.Services.Validation
  - ❌ no change due to incompatible dependency on NuGet.Services.ServiceBus
- NuGet.Services.Validation.Issues
  - ✔️ added `netstandard2.0`

Other minor changes:
- Removed unnecessary explicit MicroBuild reference (it's pulled in by Build.Directory.props)
- Removed noisy warning from the batch signing task
- Added a very basic "outer build" support to the batch signing task by enumerating files to sign for each of the 3 target frameworks that will be used during this migration process.
- Add an option to the web app creation to disable signing and to not automatically ZIP
  - This enables a yucky batch signing approach needed for `dotnet publish` of a .NET Core web app (best option I found)
- Updated to Newtonsoft.Json 10.0.3 since this is needed for the updated KeyVault dependency


src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
```diff
@@ -1,13 +1,13 @@
...
   <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFrameworks>net472;netstandard2.0</TargetFrameworks>
     <Description>Key vault access for NuGet services</Description>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.KeyVault">
-      <Version>1.0.0</Version>
+      <Version>3.0.5</Version>
```
skofman1, line 10: >. [](start = 16, length = 1)

that's a big version bump.. are we good?

skofman1 (approved):
:shipit:

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
joelverhagen: I have validated on gallery and the new .NET Core Azure Search service during an intermediate stage but I am revalidating this PR again and in more places. See https://github.com/NuGet/Engineering/issues/3487 for my test plan.

loic-sharma (approved):

PR #352, opened 2020-11-13
Allow jobs and services to flush telemetry 

loic-sharma:
Application Insights SDK buffers telemetry for up to 30 seconds (see [this](https://devblogs.microsoft.com/premier-developer/application-insights-use-case-for-telemetryclient-flush-calls/)). In https://github.com/NuGet/Engineering/issues/3270, I updated the symbol ingester job to forcefully restart if the job reached a non-recoverable state. Today, the job doesn't flush upon restarting, which causes up to 30 seconds of telemetry to be lost.

### DIFF
--- a/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
+++ b/src/NuGet.Services.Contracts/Logging/ITelemetryClient.cs
@@ -23,5 +23,7 @@ namespace NuGet.Services.Logging
             Exception exception,
             IDictionary<string, string> properties = null,
             IDictionary<string, double> metrics = null);
+
+        void Flush();
     }
 }
--- a/src/NuGet.Services.Logging/TelemetryClientWrapper.cs
+++ b/src/NuGet.Services.Logging/TelemetryClientWrapper.cs
@@ -77,5 +77,7 @@ namespace NuGet.Services.Logging
                 // logging failed, don't allow exception to escape
             }
         }
+
+        public void Flush() => _telemetryClient.Flush();
     }
 }


joelverhagen:
Have you verified that this has the desired effect? I vaguely remember some gotchas with AI flushing...

loic-sharma:
> Have you verified that this has the desired effect? I vaguely remember some gotchas with AI flushing...

Was the gotcha that you need to sleep for the flush to complete? See: https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#flushing-data

skofman1:
Regarding the gotcha, this is how it's implemented for ng.exe: https://github.com/NuGet/NuGet.Jobs/blob/master/src/Ng/Program.cs#L128

I don't know   _telemetryClient.Flush() does the same thing, but just FYI.

loic-sharma:
> I don't know _telemetryClient.Flush() does the same thing, but just FYI.

`TelemetryClient.Flush` flushes more than just the channel, it also flushes pre-aggregated metrics (see [this](https://github.com/microsoft/ApplicationInsights-dotnet/blob/41f29de01da3deb235f7d42975b0b50c062e366a/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L625-L642). Ignore the weird branch, it is the closest I could find to v2.10.0 which we use ).

> Regarding the gotcha, this is how it's implemented for ng.exe: https://github.com/NuGet/NuGet.Jobs/blob/master/src/Ng/Program.cs#L128

This is likely wrong. We should consider moving to `TelemetryClient.Flush` and adding a sleep as per the docs: https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#flushing-data

loic-sharma:
Ok I tested this with console applications. Flushing worked as expect, with and without sleep. Regardless, I will follow Application Insights recommendation of sleeping for 5 seconds when I update the symbol ingester job as I suspect a job with non-trivial amounts of buffered telemetry is more susceptible to data loss.

<details>
<summary>Test scenarios...</summary>

I tested three variants. I ran each variant 5 times and looked for data loss.

```csharp
config = TelemetryConfiguration.CreateDefault();
config.InstrumentationKey = "...";
client = new TelemetryClient(config);

client.TrackTrace($"{DateTime.UtcNow}... No flush no sleep!");
```

```csharp
config = TelemetryConfiguration.CreateDefault();
config.InstrumentationKey = "...";
client = new TelemetryClient(config);

client.TrackTrace($"{DateTime.UtcNow}... Flush, no sleep!");
client.Flush();
```

```
config = TelemetryConfiguration.CreateDefault();
config.InstrumentationKey = "...";
client = new TelemetryClient(config);

client.TrackTrace($"{DateTime.UtcNow}... Flush and sleep!");
client.Flush();
Thread.Sleep(TimeSpan.FromSeconds(5));
</details>

joelverhagen (approved):

zhhyu:
This issue may help: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407
Sleep is needed: 
https://github.com/NuGet/NuGetGallery/blob/14e370ad90860be6e19c8daf8094eef9ba03d078/src/DatabaseMigrationTools/Program.cs#L21

zhhyu (approved):

PR #355, opened 2021-01-25
Update Orchestrator message serialization libraries

loic-sharma:
Create the libraries to serialize and deserialize the new [`StartValidation`](https://github.com/NuGet/Engineering/blob/master/Server.Specs/ValidationProtocol.md#start-validation-message) and [`CheckValidationSet`](https://github.com/NuGet/Engineering/blob/master/Server.Specs/OrchestratorUnifiedValidations.md#checkvalidationset) messages.

Part of https://github.com/NuGet/Engineering/issues/3612

<details>
<summary>Testing...</summary>

Build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4406068&view=results
</details>

/cc @nikhgup

### DIFF
--- a/src/NuGet.Services.Validation/PackageValidationMessageType.cs
+++ b/src/NuGet.Services.Validation/PackageValidationMessageType.cs
@@ -5,7 +5,9 @@ namespace NuGet.Services.Validation
 {
     public enum PackageValidationMessageType
     {
+        StartValidation,
         ProcessValidationSet,
+        CheckValidationSet,
         CheckValidator,
     }
 }
--- a/src/NuGet.Services.Validation/CheckValidationSetData.cs
+++ b/src/NuGet.Services.Validation/CheckValidationSetData.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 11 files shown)


src/NuGet.Services.Validation/PackageValidationMessageType.cs
```diff
@@ -5,7 +5,9 @@ namespace NuGet.Services.Validation
 {
     public enum PackageValidationMessageType
     {
+        StartValidation,
         ProcessValidationSet,
```
joelverhagen, line 9: It's safest to only add enum values at the end since numeric values are defined from top to bottom. This is fine if we're 100% sure that we're not using integer serialization anywhere and only ever the string serialization.

joelverhagen (approved):

src/NuGet.Services.Validation/PackageValidationMessageType.cs
loic-sharma: Yup, we're not serializing the `PackageValidationMessageType` type at all. It is only used in memory when determining which property to read off the [`PackageValidationMessageData`](https://github.com/NuGet/ServerCommon/blob/dad3f6daca3bfc8106780ae7d83b2eabf22f5b86/src/NuGet.Services.Validation/PackageValidationMessageData.cs#L76-L78).

PR #360, opened 2021-02-25
adding sync methods for servicebus to be used for marketplace

nikhgup:
Marketplace job agents are synchronous. Adding async support would be a large undertaking, so instead, we will use synchronous methods to send Service Bus messages.

### DIFF
--- a/src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
+++ b/src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
@@ -39,41 +39,13 @@ namespace NuGet.Services.ServiceBus
 
         public Task SendAsync(IBrokeredMessage message)
         {
-            // For now, assume the only implementation is the wrapper type. We could clone over all properties
-            // that the interface supports, but this is not necessary right now.
-            var wrapper = message as BrokeredMessageWrapper;
-            BrokeredMessage innerMessage;
-            if (message != null)
-            {
-                innerMessage = wrapper.BrokeredMessage;
-            }
-            else
-            {
-                throw new ArgumentException(
-                    $"The message must be of type {typeof(BrokeredMessageWrapper).FullName}.",
-                    nameof(message));
-            }
-
+            BrokeredMessage innerMessage = GetBrokeredMessage(message);
             return _client.SendAsync(innerMessage);
         }
 
         public void Send(IBrokeredMessage message)
         {
-            // For now, assume the only implementation is the wrapper type. We could clone over all properties
-            // that the interface supports, but this is not necessary right now.
-            var wrapper = message as BrokeredMessageWrapper;
-            BrokeredMessage innerMessage;
-            if (message != null)
-            {
-                innerMessage = wrapper.BrokeredMessage;
-            }
-            else
-            {
-                throw new ArgumentException(
-                    $"The message must be of type {typeof(BrokeredMessageWrapper).FullName}.",
-                    nameof(message));
-            }
-
+            BrokeredMessage innerMessage = GetBrokeredMessage(message);
             _client.Send(innerMessage);
         }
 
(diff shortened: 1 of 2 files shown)


nikhgup:
@loic-sharma please review

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
```diff
@@ -53,13 +75,7 @@ public Task SendAsync(IBrokeredMessage message)
...
                     nameof(message));
             }
-
-            return _client.SendAsync(innerMessage);
-        }
-
-        public Task Close()
-        {
-            return _client.CloseAsync();
+            return innerMessage;
```
loic-sharma, line 78: Please add an extra line:
 
```suggestion

            return innerMessage;
```

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
```diff
@@ -38,6 +38,28 @@ public TopicClientWrapper(string clientId, string clientSecret, string tenantId,
...
+
+        public void CloseSync()
+        {
+            _client.Close();
+        }
+
+        public Task Close()
+        {
+            return _client.CloseAsync();
+        }
```
loic-sharma, line 60: Oops this was a mistake by us. This method should've had an `Async` suffix and `await` the task. See: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

Would you mind fixing that as part of your change?

```suggestion
        public async Task CloseAsync()
        {
            await _client.CloseAsync();
        }
```

loic-sharma: Also, this well you rename `CloseSync` to `Close` to match the underlying SDK :)

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
```diff
@@ -38,6 +38,28 @@ public TopicClientWrapper(string clientId, string clientSecret, string tenantId,
         }
 
         public Task SendAsync(IBrokeredMessage message)
+        {
+            BrokeredMessage innerMessage = GetBrokeredMessage(message);
```
loic-sharma, line 42: Nit: NuGet style is to use `var` wherever possible:

```suggestion
            var innerMessage = GetBrokeredMessage(message);
```

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
```diff
@@ -38,6 +38,28 @@ public TopicClientWrapper(string clientId, string clientSecret, string tenantId,
...
+            return _client.SendAsync(innerMessage);
+        }
+
+        public void Send(IBrokeredMessage message)
+        {
+            BrokeredMessage innerMessage = GetBrokeredMessage(message);
```
loic-sharma, line 48: Please use var:

```suggestion
            var innerMessage = GetBrokeredMessage(message);
```

loic-sharma:
Thanks for opening this! As part of this change, please add the `Send`, `Close`, and `CloseAsync` methods to the [`ITopicClient` interface](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Contracts/ServiceBus/ITopicClient.cs). Other than that, your changes look good. I'll merge this in once you've addressed the feedback.

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
```diff
@@ -38,6 +38,28 @@ public TopicClientWrapper(string clientId, string clientSecret, string tenantId,
...
+
+        public void Send(IBrokeredMessage message)
+        {
+            var innerMessage = GetBrokeredMessage(message);
+            _client.Send(innerMessage);
+        }
+
+        public void Close()
+        {
+            _client.Close();
```
loic-sharma, line 54: @joelverhagen Any concerns about this? This will swap out Gallery DI with synchronous implementation here, but this seems like a good thing to me: https://github.com/NuGet/NuGetGallery/blob/67084b9f518e299dc6712c34a8ca10506aee4bc4/src/NuGetGallery/App_Start/DefaultDependenciesModule.cs#L754-L758

loic-sharma (approved):

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
nikhgup: @loic I am not sure about Nuget source code, but will it not give build error when nugetgallery code will use new version of servicebus complaining about return type of this method(for sync it is void and for async it is Task)?

joelverhagen: This is a good change. Just make sure is doesn't cause any weird exceptions thrown on shutdown. You can use a debugger to verify this in Gallery codebase (maybe @loic-sharma will do this? donno)

joelverhagen (approved):

src/NuGet.Services.ServiceBus/TopicClientWrapper.cs
loic-sharma: > but will it not give build error ... ?

Good question. The `Release` method accepts an `Action`, so the returned `Task` is ignored.

Yup I'll take care of the Gallery verification + upgrade for this change.

zhhyu (approved):

PR #361, opened 2021-03-04
Adding build cmdlet arguments to make msbuild produce binlog

agr:
Sometimes you need it, and making msbuild to produce binlog in local copy is a bit of a pain.

This adds appropriate arguments to all cmdlets that end up calling msbuild.


loic-sharma (approved):

joelverhagen (approved):
💯

PR #363, opened 2021-03-18
Sync secret reader with background refresh task.

agr:
Progress on https://github.com/NuGet/Engineering/issues/3522

Added sync secret reader and secret injector.

Secret reader receives the list of secret names to refresh and does all refreshes strictly in background. Read operation always reads from cache (regardless of staleness), with some emergency updates to the secret list if unknown secrets are requested.

Secret injector is pretty much a copy of old secret injector, but using the new reader interface and not doing async operations.

`GetSecretNames` function was moved to a separate class to be reused between secret injector implementations.


src/NuGet.Services.FeatureFlags/FeatureFlagCacheService.cs
```diff
@@ -10,28 +10,28 @@ namespace NuGet.Services.FeatureFlags
 {
     public class FeatureFlagCacheService : IFeatureFlagCacheService
     {
-        private readonly IFeatureFlagStorageService _storage;
```
agr, line 13: Will revert the whole file and send a separate PR.
Need both changes for testing.

agr: https://github.com/NuGet/ServerCommon/pull/364

tests/NuGet.Services.KeyVault.Tests/CachingBackgroundRefreshingSecretReaderFacts.cs
```diff
@@ -0,0 +1,212 @@
...
+            _secretReaderMock
+                .Setup(sr => sr.GetSecretObjectAsync(secretName, It.IsAny<ILogger>()))
+                .ReturnsAsync(new KeyVaultSecret(secretName, secretValue, null))
+                .Callback(() => mre.Set());
+
+            Record.Exception(() => target.GetSecret(secretName));
+            Assert.True(mre.Wait(1000));
+            // There is a race here between the current thread calling `GetSecret` and the background thread
+            // refreshing the cache. We'll sleep extra second to let background thread win that race.
+            Thread.Sleep(1000);
```
agr, line 73: Any suggestions here?

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            if (secretName == null)
+            {
+                throw new ArgumentNullException(nameof(secretName));
+            }
+            if (string.IsNullOrWhiteSpace(secretName))
+            {
+                throw new ArgumentException($"{nameof(secretName)} must not be empty", nameof(secretName));
+            }
+            // Given that the user of the object called WaitForTheFirstRefhresh() on their own
```
joelverhagen, line 128:
```suggestion
            // Given that the user of the object called WaitForTheFirstRefresh() on their own
```

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
```diff
@@ -0,0 +1,275 @@
...
+            // this call should never block
+            WaitForTheFirstRefresh();
+            if (_cachedSecrets.TryGetValue(secretName, out var cachedSecret))
+            {
+                if (cachedSecret == null)
```
joelverhagen, line 133: I'm wondering if allowing the `null` state is worth it. Consider the Azure Search Service which does a similar synchronous wait for all secrets:
https://github.com/NuGet/NuGet.Jobs/blob/376ac06e6e07d4ee8d1e28f6b2346e3891487496/src/NuGet.Services.SearchService.Core/Startup.cs#L131-L168

It essentially throws an exception if any secret name is asked for that isn't cached:
https://github.com/NuGet/NuGet.Jobs/blob/376ac06e6e07d4ee8d1e28f6b2346e3891487496/src/NuGet.Services.SearchService.Core/Startup.cs#L161

Do you think that we could simplify the code in this similar way?

src/NuGet.Services.KeyVault/ISyncSecretReader.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace NuGet.Services.KeyVault
+{
+    /// <summary>
+    /// An alternative to <see cref="ISecretReader"/> that provides sync interface for
+    /// secret retrieval. It is assumed that an implementation would not block calls to the interface method(s) for
+    /// some kind of I/O, but it will retrieve them in appropriate context instead (for example background thread).
+    /// </summary>
+    public interface ISyncSecretReader
+    {
+        string GetSecret(string secretName);
```
joelverhagen, line 13: I'm wondering if we really need the sync version of this interface. Is it adding much? In my Azure Search approach I simply have an async implementation that, after startup and some flag is set, uncached reads throw an exception:
https://github.com/NuGet/ServerCommon/blob/554144e453653e1ae65c9234c9742ed811d50457/src/NuGet.Services.KeyVault/RefreshableSecretReader.cs#L98-L101

agr: Strictly speaking, we don't need sync version, but it helps getting rid of good chunk of `Task.Result` calls we make during DI resolution. Previously, configuration was constructed once, so those `Task.Result` calls happened at startup, now they would happen all the time during service operation. If there is a negative impact of using it, I wanted to not have it called at all.

tests/NuGet.Services.KeyVault.Tests/CachingBackgroundRefreshingSecretReaderFacts.cs
joelverhagen: Maybe a bit ugly but it couldn't be more simple 😄. I think it's fine and better than some complex usage of async primitives or callbacks.

joelverhagen:
In general the implementation looks very solid. However I wonder whether it's really necessary to have the number of safe-guards you have. In my mind the path for secret reading should be very simple and not require new interfaces or much forking in logic.

1. Sync blocking on async during startup to populate all of the secrets initially
2. Subsequent reads always come from cache (and throw if the secret name is not present) -- ideally we get step 1 good enough so there are no "surprise" secrets
3. Refresh in the background.

This is mostly what you already have but I've implemented it in what I feel is a simpler way in Search Service. That being said Search Service was built from the ground up with KeyVault integration in mind so it there were less surprising places where secrets were needed. 

I am fine with this implementation if you thing it is really necessary given the complexity of gallery itself. I just hope we keep the absolute simplest implementation possible by pushing as much smarts as possible into that step 1 where we know all the secrets and load the initial version synchronously.

joelverhagen:
It does make me a bit sad that Search Service and Gallery now have different code to implement a somewhat similar pattern.

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
agr: This code path will trigger if unexpected secret had been requested multiple times in short succession: for the first call `TryGetValue` will return `false` and it would get to a branch where it adds to a list of secrets to retrieve and throws, all subsequent calls will get to this branch. I mostly added to have a bit more clarity on what is happening if this happens.

The reason all this code exists in the first place is because I didn't find a reliable way to enumerate all the configuration items and instead go off the configuration objects used in the code. It may pointless right now, but might help in the future, especially with `ConfigurationService` exposing methods like `ResolveConfigObject` and `ReadSetting` for everyone (I checked all the usages for now, but future updates may change it). I am a bit paranoid that one of those can be called in some rarely used code, so it might not get noticed until it causes some outage in the middle of the night.

src/NuGet.Services.KeyVault/EmptySyncSecretReader.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace NuGet.Services.KeyVault
+{
+    public class EmptySyncSecretReader : ISyncSecretReader
+    {
+        public string GetSecret(string secretName)
+            => secretName;
+
+        public ISecret GetSecretObject(string secretName)
+            => new KeyVaultSecret(secretName, secretName, null);
```
loic-sharma, line 15: Super nit: please add the param name for the `null` argument

src/NuGet.Services.KeyVault/SecretUtilities.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
+    public static class SecretUtilities
```
loic-sharma, line 12: Should this be internal?

src/NuGet.Services.KeyVault/SecretUtilities.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        public static ICollection<string> GetSecretNames(string input, string frame)
```
loic-sharma, line 14: Should this have tests?

loic-sharma: Also it's not obvious what this method does, consider adding quick comments

src/NuGet.Services.KeyVault/ISyncSecretReader.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.KeyVault
+{
+    /// <summary>
+    /// An alternative to <see cref="ISecretReader"/> that provides sync interface for
+    /// secret retrieval. It is assumed that an implementation would not block calls to the interface method(s) for
+    /// some kind of I/O, but it will retrieve them in appropriate context instead (for example background thread).
+    /// </summary>
+    public interface ISyncSecretReader
```
loic-sharma, line 11: Do we need to split the `ISyncSecretReader` and `ISecretReader` interfaces? Why not have a single `ISecretReader` interface that exposes both sync and async methods?

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
```diff
@@ -0,0 +1,275 @@
...
+        private CancellationTokenSource _forceRefresh = new CancellationTokenSource();
+
+        /// <summary>
+        /// The object constructor.
+        /// </summary>
+        /// <param name="underlyingSecretReader">The actual secret reader that would provide secret values to a background task.</param>
+        /// <param name="backgroundTaskStarter">The hook into caller's infrastructure to run background tasks. A function that accepts a
+        /// function that runs a background secret refresh task and makes sure background task infrastructure executes it. See example below.</param>
+        /// <param name="secretNames">The list of secret names to maintain.</param>
+        /// <param name="refreshInterval">The time for which secret values are going to be EditableFeatureFlagFileStorageServicecached.</param>
```
loic-sharma, line 54: This comment looks funky

src/NuGet.Services.KeyVault/ISyncSecretReader.cs
agr: The base implementation ([`KeyVaultReader`](https://github.com/NuGet/ServerCommon/blob/main/src/NuGet.Services.KeyVault/KeyVaultReader.cs)) would have to either resort to `Task.Result` or similar to implement sync versions of the methods or not implement them at all (defeating the purpose of having them declared in the interface). I decided to split to not have to do either. This also clearly indicates capabilities: sync version does not do any shady async calls.

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
agr: accidental paste...

src/NuGet.Services.KeyVault/SecretUtilities.cs
agr: This method is needed by Gallery to enumerate used secrets, so, no.

src/NuGet.Services.KeyVault/CachingBackgroundRefreshingSecretReader.cs
```diff
@@ -0,0 +1,275 @@
...
+                        catch (Exception ex)
+                        {
+                            // we'll report the exception and ignore it, so we don't leak it outside the function
+                            _telemetryService?.TrackSecretRefreshFailure(keyValuePair.Key);
+                            _logger?.LogError(ex, "Exception while refreshing the secret {SecretName}", keyValuePair.Key);
+                            allRefreshesSucceeded = false;
+                        }
+                    }
+                }
+                if (allRefreshesSucceeded)
```
loic-sharma, line 223: What do you think of adding a logging if the refresh loop has never succeeded (if both `allRefreshesSucceeded` and `_firstRefreshComplete.IsSet` are false)?

agr: Considering that at the point of use it is expected that `WaitForTheFirstRefresh()` (which will block until `_firstRefreshComplete` is signaled), I think we will notice it even without additional logs/metrics. But, it made me consider moving the `_telemetryService?.TrackSecretRefreshIteration();` call to the bottom of the loop and passing `allRefreshesSucceeded` into it, so we could track if there are repeated failures.

src/NuGet.Services.KeyVault/SecretUtilities.cs
agr: Added comments. As for tests: this was previously an implementation detail of a `SecretInjector`, so it is indirectly tested through it. I'll see if I can extract anything from existing tests or add new.

tests/NuGet.Services.KeyVault.Tests/CachingBackgroundRefreshingSecretReaderFacts.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            const string secretName = "secretName";
+            var target = CreateTarget(new[] { secretName }, TimeSpan.FromMilliseconds(500), TimeSpan.FromMilliseconds(100));
+            target.WaitForTheFirstRefresh();
+
+            _secretReaderMock
+                .Setup(sr => sr.GetSecretObjectAsync(secretName, It.IsAny<ILogger>()))
+                .ReturnsAsync(new KeyVaultSecret(secretName, "someSecretValue", null));
+
+            await Task.Delay(1600);
```
drewgillies, line 89: Can we do this with a shorter delay, or do we come up against processing speeds here? Those seconds add up :)

drewgillies: If CI on Server.Common is short though, forget it.

zhhyu:
Looking now ...

PR #364, opened 2021-03-25
Storage factory in FeatureFlagCacheService

agr:
Progress on https://github.com/NuGet/Engineering/issues/3522

`FeatureFlagCacheService` needs to be a singleton in services to maintain cache, but it should not be holding on to an instance of a storage service: it prevents proper secret refresh. The class is modified instead to receive a storage service factory that it uses to produce instances of a storage service as needed.


skofman1 (approved):
:shipit:

loic-sharma (approved):

PR #368, opened 2021-05-11
Allow optional JSON config files

joelverhagen:
This will allow environments that expect config to come from environment variables (e.g. App Service app settings) to use the same code as local environments that more easily use JSON config files.

### DIFF
--- a/src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
@@ -15,9 +15,18 @@ namespace NuGet.Services.Configuration
             string path,
             ISecretInjector secretInjector)
         {
+            return configurationBuilder.AddInjectedJsonFile(path, secretInjector, optional: false);
+        }
+
+        public static IConfigurationBuilder AddInjectedJsonFile(
+            this IConfigurationBuilder configurationBuilder,
+            string path,
+            ISecretInjector secretInjector,
+            bool optional)
+        {
             configurationBuilder = configurationBuilder ?? throw new ArgumentNullException(nameof(configurationBuilder));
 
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -14,10 +14,19 @@ public static class ConfigurationBuilderExtensions
...
             ISecretInjector secretInjector)
+        {
+            return configurationBuilder.AddInjectedJsonFile(path, secretInjector, optional: false);
+        }
+
+        public static IConfigurationBuilder AddInjectedJsonFile(
+            this IConfigurationBuilder configurationBuilder,
+            string path,
+            ISecretInjector secretInjector,
+            bool optional)
```
drewgillies, line 25: Couldn't this just have been an optional parameter?

drewgillies (approved):
Wondering about overloading a method vs an optional parameter, but not a big issue. Looks good.

src/NuGet.Services.Configuration/ConfigurationBuilderExtensions.cs
drewgillies: I mean an optional parameter, not an optional parameter :-D

joelverhagen: Optional parameters can be a runtime breaking change. We would need to recompile all referencing code make it start "expecting" the new method signature.

joelverhagen: There's a bajillion gotchas that I don't know all of for breaking vs. non breaking changes. It's a deep subject. But an overload is more like to be compatible in this case from my knowledge.

zhhyu (approved):

PR #370, opened 2021-06-14
Add support for PAT for multiple private feeds using Credential Provider

dannyjdev:
This adds multiple private feeds to the env variable `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` that's used by the Credential Provider. This will remove the need to add private feeds one by one.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -8,7 +8,9 @@ param (
     [string]$SimpleVersion = '1.0.0',
     [string]$SemanticVersion = '1.0.0-zlocal',
     [string]$Branch,
-    [string]$CommitSHA
+    [string]$CommitSHA,
+    [string]$Username,
+    [string]$AccessToken
 )
 
 # For TeamCity - If any issue occurs, this script fails the build. - By default, TeamCity returns an exit code of 0 for all powershell scripts, even if they fail
@@ -51,9 +53,9 @@ Invoke-BuildStep 'Getting private build tools' { Install-PrivateBuildTools } `
 Invoke-BuildStep 'Cleaning test results' { Clean-Tests } `
     -ev +BuildErrors
 
-Invoke-BuildStep 'Installing NuGet.exe' { Install-NuGet } `
+Invoke-BuildStep 'Installing NuGet.exe' { Install-NuGet -Username $Username -AccessToken $AccessToken } `
     -ev +BuildErrors
-    
+
 Invoke-BuildStep 'Clearing package cache' { Clear-PackageCache } `
     -skip:(-not $CleanCache) `
     -ev +BuildErrors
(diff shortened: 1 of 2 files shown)


build/common.ps1
```diff
@@ -471,6 +473,32 @@ Function Install-NuGet {
...
+        $endpoints = @()
+        $endpointURIs = @()
+
+        Get-ChildItem "$NuGetClientRoot\nuget.config" -Recurse |% {
+            $nugetConfig = [xml](Get-Content -Path $_)
+
+            $nugetConfig.configuration.packageSources.add |? { ($_.value -match '^https://pkgs\.dev\.azure\.com/') -or ($_.value -match '^https://[\w\-]+\.pkgs\.visualstudio\.com/') } |% {
+                if ($endpointURIs -notcontains $_.Value) {
+                    $endpointURIs += $_.Value;
+                    $endpoint = New-Object -TypeName PSObject;
```
agr, line 486: You can do
```
$endpoint = New-Object PSObject -Property @{endpoint=$_.value; username='PAT'; password=$AccessToken}
```
instead of 4 lines of code;

build/common.ps1
```diff
@@ -471,6 +473,32 @@ Function Install-NuGet {
         throw "No file exists at the expected nuget.exe path: $NuGetExe"
     }
 
+    if ($AccessToken) {
+        $endpoints = @()
+        $endpointURIs = @()
+
+        Get-ChildItem "$NuGetClientRoot\nuget.config" -Recurse |% {
```
agr, line 480: Please, don't use cmdlet aliased in scripts. `%` should be `ForEach-Object` everywhere.

build.ps1
```diff
@@ -8,7 +8,8 @@ param (
     [string]$SimpleVersion = '1.0.0',
     [string]$SemanticVersion = '1.0.0-zlocal',
     [string]$Branch,
-    [string]$CommitSHA
+    [string]$CommitSHA,
+    [string]$AccessToken
```
agr, line 12: Will this break pipelines that don't already specify that argument? Should it have a default value?

build/common.ps1
```diff
@@ -402,7 +402,9 @@ Function Update-Submodule {
 }
 Function Install-NuGet {
     [CmdletBinding()]
-    param()
+    param(
+        [string]$AccessToken
```
agr, line 406: Same question as for `build.ps1` here: many scripts call it without argument. Will those break?

agr (approved):
Please, double check if default values for new parameters should be specified and handled, and don't use `%` and `?`.

dannyjdev:
Closing this issue since is not needed anymore; we use a task on our AzDo pipelines that do the same work.

PR #371, opened 2021-06-16
init.ps1 script now clones ServerCommon repository with a specific commit.

dannyjdev:
This change will help us to maintain a commit hash for specific reference when building a project.

- Since we cannot clone a repository using only a commit id, I'm adding branch name to help us have the flexibility to specify our own if necessary on `$BuildBranch` (default is `main`).
- `$BuildBranchCommit` specifies the exact commit the branch will move to. If no commit is specified, then it will use the latest from the cloned branch.

### DIFF
--- a/build/init.ps1
+++ b/build/init.ps1
@@ -1,6 +1,7 @@
 [CmdletBinding()]
 param(
-    [string]$BuildBranch = 'main'
+    [string]$BuildBranch = 'main',
+    [string]$BuildBranchCommit = $null
 )
 
 # This file is downloaded to "build/init.ps1" so use the parent folder as the root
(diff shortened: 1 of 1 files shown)


build/init.ps1
```diff
@@ -14,12 +15,22 @@ Function Get-BuildTools {
...
+    {
+        $BuildBranchCommit = & cmd /c "git rev-parse origin/$BuildBranch 2>&1"
+    }
+    else
+    {
+        & cmd /c "git fetch origin $BuildBranchCommit 2>&1"
+        & cmd /c "git reset --hard FETCH_HEAD 2>&1"
+    }
+
+    Write-Host "ServerCommon repository cloned with branch $BuildBranch and $BuildBranchCommit commit."
```
loic-sharma, line 33: If `$BuildBranchCommit` is `$null` this message will be weird. Consider moving this into the if/else statements above.

build/init.ps1
```diff
@@ -14,12 +15,22 @@ Function Get-BuildTools {
 
     if (-not (Test-Path $ServerCommonRoot))
     {
-        Write-Host "Clonning ServerCommon repository with $BuildBranch branch"
-        & cmd /c "git clone -b $BuildBranch https://github.com/NuGet/ServerCommon.git 2>&1"
+        Write-Host "Clonning ServerCommon repository with $BuildBranch branch."
```
loic-sharma, line 18:
```suggestion
        Write-Host "Cloning ServerCommon repository with $BuildBranch branch."
```

loic-sharma (approved):

build/init.ps1
agr: if `$BuildBranchCommit` then it is going to be set to the latest snapshot of the branch, right? So by the time `Write-Host` is called it will have a value.

build/init.ps1
```diff
@@ -14,12 +15,22 @@ Function Get-BuildTools {
...
     Set-Location $ServerCommonRoot
-    $BuildBranchCommit = & cmd /c "git rev-parse origin/$BuildBranch 2>&1"
-    Write-Host "Latest commit in branch $BuildBranch is $BuildBranchCommit"
+    if (!$BuildBranchCommit)
+    {
+        $BuildBranchCommit = & cmd /c "git rev-parse origin/$BuildBranch 2>&1"
+    }
+    else
+    {
+        & cmd /c "git fetch origin $BuildBranchCommit 2>&1"
```
agr, line 29: Will this fetch the specified snapshot alone? Can we run that command without `git clone` above? Would `git init && git remote add origin https://github.com/NuGet/ServerCommon.git` be enough instead?

agr (approved):

build/init.ps1
loic-sharma: Ah right, I can't read. Ignore this comment :)

build/init.ps1
dannyjdev: Done, I just removed the git clone and added that line, verified that any commit can be retrieved with that approach.

build/init.ps1
dannyjdev: Removed that line.

agr (approved):

PR #374, opened 2021-10-27
A sync interface for SQL connection.

agr:
Attempting to make Gallery's life easier with sync SQL connection creation.
Added `Caching` extensions of secret injector and reader interfaces. `AzureSqlConnectionFactory` was changed to require caching version of an injector.

### DIFF
--- a/src/NuGet.Services.KeyVault/RefreshableSecretReader.cs
+++ b/src/NuGet.Services.KeyVault/RefreshableSecretReader.cs
@@ -89,8 +89,24 @@ namespace NuGet.Services.KeyVault
             return false;
         }
 
-        public bool TryGetCachedSecretObject(string secretName, ILogger logger, out ISecret secretObject) 
-            => TryGetCachedSecretObject(secretName, out secretObject);
+        public bool TryGetCachedSecretObject(string secretName, ILogger logger, out ISecret secretObject)
+        {
+            if (_cache.TryGetValue(secretName, out secretObject))
+            {
+                return true;
+            }
+
+            if (_settings.BlockUncachedReads)
+            {
+                throw new InvalidOperationException($"The secret '{secretName}' is not cached.");
+            }
+
+            secretObject = null;
+            return false;
+        }
+
+        public bool TryGetCachedSecretObject(string secretName, out ISecret secretObject)
+            => TryGetCachedSecretObject(secretName, logger: null, secretObject: out secretObject);
 
         private async Task<string> UncachedGetSecretAsync(string secretName)
         {
@@ -103,22 +119,6 @@ namespace NuGet.Services.KeyVault
             var secretObject = await _secretReader.GetSecretObjectAsync(secretName);
             _cache.AddOrUpdate(secretName, secretObject, (_, __) => secretObject);
             return secretObject;
-        }
-
-        public bool TryGetCachedSecretObject(string secretName, out ISecret secret)
-        {
-            if (_cache.TryGetValue(secretName, out secret))
-            {
-                return true;
-            }
-
-            if (_settings.BlockUncachedReads)
-            {
-                throw new InvalidOperationException($"The secret '{secretName}' is not cached.");
-            }
-
-            secret = null;
-            return false;
         }
     }
 }


src/NuGet.Services.KeyVault/RefreshableSecretReader.cs
```diff
@@ -88,7 +105,7 @@ private async Task<ISecret> UncachedGetSecretObjectAsync(string secretName)
             return secretObject;
         }
 
-        private bool TryGetCachedSecretObject(string secretName, out ISecret secret)
+        public bool TryGetCachedSecretObject(string secretName, out ISecret secret)
```
zhhyu, line 108: Will it be better that the null logger method call the logger one with the implementation? I know this is because of the old implementation but it seems easy to fix.

zhhyu:
Still looking...

zhhyu:
Could we draft a PR in Gallery? I am interested in how it will be changed in Gallery.

loic-sharma:
@zhhyu See this: https://github.com/NuGet/NuGetGallery/issues/8858

```cs
private static DbConnection CreateDbConnection(ISqlConnectionFactory connectionFactory)
{
    // Attempt to create a connection using cached secrets and AAD token.
    // The vast majority of requests should be able to complete synchronously.
    if (connectionFactory.TryGet(out var connection)
    {
      return connection;
    }

    // Secrets or AAD token are near expiry or already expire.
    // Use sync-over-async to allow refreshing the token or secrets.
    // This blocks the calling thread and creates a new connection using
    // a threadpool thread. This may result in threadpool starvation.
    return Task.Run(() => connectionFactory.CreateAsync()).Result;
}
```

loic-sharma (approved):

agr:
@zhhyu the Gallery changes are in the following branch: https://github.com/NuGet/NuGetGallery/tree/agr-sync-dbconn

src/NuGet.Services.Sql/AzureSqlConnectionFactory.cs
```diff
@@ -97,5 +135,17 @@ protected virtual async Task<string> AcquireAccessTokenAsync(string clientCertif
 
             return authResult.AccessToken;
         }
+
+        protected virtual bool TryAcquireAccessToken(string clientCertificateData, out string accessToken)
```
zhhyu, line 139: Nit: Do we need this method? I think we can just simply call "AccessTokenCache.TryGetCached" in "TryCreate"?

src/NuGet.Services.KeyVault/RefreshableSecretReader.cs
```diff
@@ -16,7 +16,7 @@ namespace NuGet.Services.KeyVault
     /// cached with a previous invocation. The <see cref="RefreshAsync"/> method is used to refresh the values of
     /// secrets that have already been cached.
     /// </summary>
-    public class RefreshableSecretReader : ISecretReader
+    public class RefreshableSecretReader : ICachingSecretReader
```
zhhyu, line 19: Nit: Do we need to upgrade this class? I see in Gallery, we use "CachingSecretReader".

zhhyu:
Thanks! That makes sense.

I can see that the following cases pass:
1. If the certificate needs to be refreshed and will get expired, it will return false, and use async to retrieve the certificate again; It will use background refreshment to get the access token with the new certificate;
2. If the certificate is retrieved but the background refreshment to get the access token fails, the certificate is already updated in caching, so it will trigger the background refreshment again but the certificate has been updated and valid;
3. If the access token will get expired soon, it will trigger the background refreshment;
4. If the access token has expired, it will return false, and use async to update the access token.

Approved. (BTW, I didn't check tests.)

zhhyu:
Will it be helpful if we log the change of some steps?

zhhyu (approved):

src/NuGet.Services.KeyVault/RefreshableSecretReader.cs
agr: That particular one we didn't need to update, but for completeness sake, since it has caching already implemented I had the interface updated.

PR #375, opened 2021-11-11
Find by subject certificate helper method.

agr:
Need a common way to find certificate by subject for https://github.com/NuGet/Engineering/issues/4101 and similar items.

The function selects certificates with start date before now and end date after now and picks the most recent out of those.

### DIFF
--- a/src/NuGet.Services.KeyVault/CertificateUtility.cs
+++ b/src/NuGet.Services.KeyVault/CertificateUtility.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 
 namespace NuGet.Services.KeyVault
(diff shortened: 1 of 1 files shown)


loic-sharma:
Can we check the issuer too to follow SN+I?

agr:
We shouldn't be checking anything. This is a helper method to retrieve the cert to be passed down to AAD for authentication and AAD will check everything there is to check.

The method is added to provide a simple common way to identify certificate to use for authentication that is resilient to certificate rotations (so we don't need to redeploy): when rotation happens the subject is going to stay the same, so when new cert arrives it will get automatically picked up.

loic-sharma (approved):

zhhyu:
Will check it next Monday.

src/NuGet.Services.KeyVault/CertificateUtility.cs
```diff
@@ -28,5 +29,27 @@ public static X509Certificate2 FindCertificateByThumbprint(StoreName storeName,
...
+        public static X509Certificate2 FindLatestActiveCertificateBySubject(StoreName storeName, StoreLocation storeLocation, string subject, bool validationRequired)
+        {
+            using (var store = new X509Store(storeName, storeLocation))
+            {
+                store.Open(OpenFlags.ReadOnly);
+                var certificateCollection = store.Certificates.Find(X509FindType.FindBySubjectDistinguishedName, subject, validationRequired);
+                var now = DateTime.Now; // X509Certificate2 object NotBefore and NotAfter properties are in local time zone, so we'll compare with local time
+                var certificate = certificateCollection
+                    .OfType<X509Certificate2>()
+                    .Where(candidateCertificate => candidateCertificate.NotBefore <= now && now <= candidateCertificate.NotAfter)
```
zhhyu, line 42: I just wonder that whether we should use < rather than <= for NotAfter?

src/NuGet.Services.KeyVault/CertificateUtility.cs
```diff
@@ -28,5 +29,27 @@ public static X509Certificate2 FindCertificateByThumbprint(StoreName storeName,
...
+                    .OrderByDescending(candidateCertificate => candidateCertificate.NotAfter)
+                        .ThenByDescending(candidateCertificate => candidateCertificate.NotBefore)
```
zhhyu, line 44: Will it be better if we rank NotBefore by asc because of a safer range? We have already filtered the valid ones, so it's not a big deal.

zhhyu (approved):

dannyjdev (approved):

src/NuGet.Services.KeyVault/CertificateUtility.cs
agr: Certificate validity is inclusive on both ends of the range, hence the inclusion of equality checks. In either case if we are milliseconds close to expiration there would be race, but with our rotation policy we shouldn't be anywhere near to it.

src/NuGet.Services.KeyVault/CertificateUtility.cs
agr: It would be confusing when manual rotation happens and new cert is issued with the same expiration date as the old one (been there). With my approach, you issue new cert and immediately should see in logs that it is used, if we sort by `NotBefore` ascending it will cause older certificate to continue being used and it would be unclear if rotation happened.

zhhyu: I see. That makes sense.

PR #379, opened 2021-12-02
.net core 2.2 is out of support, trying to use 3.1

agr:
Previously used version of .net core installer does not seem to be available anymore (and causes build failures). 3.1 is LTS, so will use that one instead.


loic-sharma:
FYI 3.1 will go out of support next year: https://dotnet.microsoft.com/platform/support/policy

loic-sharma (approved):

PR #381, opened 2022-01-03
Update service bus version to update the breaking dependencies

shishirx34:
`WindowsAzure.ServiceBus` takes a dependency on the `System.IdentityModel.Tokens.Jwt` which is overridden by another package with a superior version of `Jwt` library which breaks the managed identity authentication due to a breaking change. Updating the service bus to up the underlying dependency.

### DIFF
--- a/tests/NuGet.Services.ServiceBus.Tests/NuGet.Services.ServiceBus.Tests.csproj
+++ b/tests/NuGet.Services.ServiceBus.Tests/NuGet.Services.ServiceBus.Tests.csproj
@@ -5,9 +5,6 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


tests/NuGet.Services.ServiceBus.Tests/NuGet.Services.ServiceBus.Tests.csproj
```diff
@@ -6,7 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="WindowsAzure.ServiceBus">
-      <Version>6.0.3</Version>
```
joelverhagen, line 9: Can this be implied by the project dependency and therefore removed entirely? Makes maintenance easier.

joelverhagen (approved):

PR #382, opened 2022-01-04
[SB] Fix assembly warnings caused by version mismatch

shishirx34:
Third time's the charm.

### DIFF
--- a/src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
+++ b/src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
@@ -10,7 +10,7 @@
       <Version>3.0.5</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication">
-      <Version>1.3.1</Version>
+      <Version>1.4.0</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions">
       <Version>2.2.0</Version>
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
```diff
@@ -10,7 +10,7 @@
       <Version>3.0.5</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication">
-      <Version>1.3.1</Version>
+      <Version>1.4.0</Version>
```
joelverhagen, line 13: Can we remove this one or is it needed in the KeyVault graph too?

joelverhagen (approved):

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
shishirx34: This was already present in the project. I am assuming it was added because it was needed (?)

joelverhagen: No clue! You could check the assets file for KeyVault.csproj and see if any other deps. need AppAuthentication

dannyjdev (approved):

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
shishirx34: Keyvault uses TokenProvider referenced from this package. So it is required.

PR #398, opened 2022-07-13
update newtonsoft.json dependencies

lyndaidaii:
WindowsAzure.Storage
9.3.3  has dependency of newtonsoft. Update newtonsoft.json to 13.0.1 avoid bring low version newtonsoft

Address: https://github.com/NuGet/Engineering/issues/4452

### DIFF
--- a/src/NuGet.Services.Configuration/NuGet.Services.Configuration.csproj
+++ b/src/NuGet.Services.Configuration/NuGet.Services.Configuration.csproj
@@ -21,9 +21,6 @@
     <PackageReference Include="Microsoft.Extensions.Options">
       <Version>2.2.0</Version>
     </PackageReference>
-    <PackageReference Include="Newtonsoft.Json">
-      <Version>13.0.1</Version>
-    </PackageReference>    
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


agr (approved):

src/NuGet.Services.Configuration/NuGet.Services.Configuration.csproj
```diff
@@ -21,6 +21,9 @@
     <PackageReference Include="Microsoft.Extensions.Options">
       <Version>2.2.0</Version>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json">
+      <Version>13.0.1</Version>
```
joelverhagen, line 25: This should not be necessary with the transitive dependency in KeyVault.csproj, right?

tests/NuGet.Services.Storage.Tests/NuGet.Services.Storage.Tests.csproj
```diff
@@ -8,6 +8,9 @@
     <PackageReference Include="Moq">
       <Version>4.10.0</Version>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json">
+      <Version>13.0.1</Version>
```
joelverhagen, line 12: This should not be necessary due to the transitive dependency in Storage.csproj right?

joelverhagen (changes requested):
⌚ we can prune the changes a bit I think.

lyndaidaii:
@joelverhagen, please take another look. Thanks.

joelverhagen (approved):

PR #399, opened 2022-07-14
Add netstandard2.1 target to NuGet.Services.Cursor/Storage

drewgillies:
This is to provide storage (and possibly cursor) services to this work: https://github.com/NuGet/Engineering/issues/4439

### DIFF
--- a/src/NuGet.Services.Cursor/NuGet.Services.Cursor.csproj
+++ b/src/NuGet.Services.Cursor/NuGet.Services.Cursor.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFrameworks>net472;netstandard2.1</TargetFrameworks>
     <Description>Shared cursor classes</Description>
   </PropertyGroup>
 
@@ -9,7 +9,7 @@
     <ProjectReference Include="..\NuGet.Services.Storage\NuGet.Services.Storage.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
     <Reference Include="System.Net.Http" />
   </ItemGroup>
 
--- a/src/NuGet.Services.Storage/AzureStorage.cs
+++ b/src/NuGet.Services.Storage/AzureStorage.cs
@@ -50,9 +50,17 @@ namespace NuGet.Services.Storage
             _directory = directory;
             _useServerSideCopy = useServerSideCopy;
 
+#if NETFRAMEWORK
             if (initializeContainer && _directory.Container.CreateIfNotExists())
+#else
+            if (initializeContainer && _directory.Container.CreateIfNotExistsAsync().Result) // the framework code blocks the thread similarly
+#endif
             {
+#if NETFRAMEWORK
                 _directory.Container.SetPermissions(new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Blob });
+#else
+                Task.Run(() => _directory.Container.SetPermissionsAsync(new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Blob })).Wait();  // the framework code blocks the thread similarly
+#endif
 
                 if (Verbose)
                 {
@@ -88,7 +96,11 @@ namespace NuGet.Services.Storage
 
             var blob = _directory.GetBlockBlobReference(blobName);
 
+#if NETFRAMEWORK
             if (blob.Exists())
+#else
+            if (blob.ExistsAsync().Result) // the framework code blocks the thread similarly
+#endif
             {
                 return true;
             }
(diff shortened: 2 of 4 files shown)


src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -268,10 +284,18 @@ protected override async Task<StorageContent> OnLoad(Uri resourceUri, Cancellati
 
             CloudBlockBlob blob = _directory.GetBlockBlobReference(name);
 
+#if NETFRAMEWORK
```
joelverhagen, line 287: Can we just make `NETFRAMEWORK` use the async flow here? Seems to me that the risk is low.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -311,7 +335,11 @@ protected override async Task OnDelete(Uri resourceUri, CancellationToken cancel
...
```
joelverhagen, line 338: Same as above, can this be async for both targets?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -50,9 +50,17 @@ public class AzureStorage : Storage
...
             if (initializeContainer && _directory.Container.CreateIfNotExists())
+#else
+            if (initializeContainer && _directory.Container.CreateIfNotExistsAsync().Result) // the framework code blocks the thread similarly
+#endif
             {
+#if NETFRAMEWORK
                 _directory.Container.SetPermissions(new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Blob });
+#else
+                Task.Run(() => _directory.Container.SetPermissionsAsync(new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Blob })).Wait();  // the framework code blocks the thread similarly
```
joelverhagen, line 62: Why `Task.Run` instead of just `Wait()` on the task returned by `SetPermissionsAsync`?

Also, for all of the places where there's a mix of sync and async, can we introduce a async, private, helper function that we just block on in the sync version?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -50,9 +50,17 @@ public class AzureStorage : Storage
             _directory = directory;
             _useServerSideCopy = useServerSideCopy;
 
+#if NETFRAMEWORK
             if (initializeContainer && _directory.Container.CreateIfNotExists())
+#else
+            if (initializeContainer && _directory.Container.CreateIfNotExistsAsync().Result) // the framework code blocks the thread similarly
```
joelverhagen, line 56: Can this logic be moved to a lazy method? e.g. a `Lazy<Task<...>` field that runs this initialization code exactly once upon the first method invocation? Also can we as async versions of all of the methods so we don't need to do "sync over async" (which has problems: https://stackoverflow.com/a/24298425)

This could move us in the right direction where we a) don't do hard work in the constructor and b) have async methods for all async dependencies.

It's possible we'll run into deadlock situations with sync over async depending on our runtime.

joelverhagen: If it's both too expensive to refactor like this and we don't run into any problems with sync over async, then I understand taking the shortcut, but it's really not ideal.

joelverhagen (changes requested):
I have left comments on a) refactoring to do async properly or b) at least moving all async invocations to a single helper method so we only have a single `.Wait()` or `.Result` per flow.

src/NuGet.Services.Cursor/NuGet.Services.Cursor.csproj
```diff
@@ -1,15 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFrameworks>net472;netstandard2.1</TargetFrameworks>
```
agr, line 4: Can we just target netstandard2.0 instead? What do we need from 2.1?

joelverhagen: Look at the dependencies of https://www.nuget.org/packages/EntityFramework/. This is why it was set as such on NuGetGallery repos.

src/NuGet.Services.Storage/AzureStorage.cs
drewgillies: Yes, this was a little excessive - Task.Run will put it on a thread pool thread but that's not necessary here.

src/NuGet.Services.Storage/AzureStorage.cs
drewgillies: Yes, good call.

src/NuGet.Services.Storage/AzureStorage.cs
drewgillies: We simply can't do this without a significant refactor across the board. For example, this class present `Exists()` and `ExistsAsync()`, overriding methods in the base class `Storage`. This change by necessity turns all flows into async, which is great, but it forces logic like the following into the synchronous method `Exists()`:

```
(await _directory.Value).GetBlockBlobReference(blobName);
```

This is fine/readable/maintainable, but neccesitates an async flow here--we can't access _directory if the caller has requested an initialization via the `initializeContainer` constructor argument, and for it to be anything other than asynchronous access (because initialization is asynchronous). This means `Exists()` needs to return a `Task<bool>` and we can't override the base method due to a different signature. We are essentially breaking an API.

Please let me know if I'm missing something @joelverhagen and @agr (I hope so). I like this approach but if we want to break the API, there will be cost and possibly a bug tail. Which we may also want to embrace.

drewgillies:
I'm going to close this one for now as I may have a workaround downstream.

PR #402, opened 2022-08-15
Add connection string processing to app insights client

drewgillies:
This will futureproof for this deprecation: https://docs.microsoft.com/en-us/azure/azure-monitor/app/sdk-connection-string?tabs=net, also https://docs.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings

Also consumed in [this work](https://github.com/NuGet/Engineering/issues/4376), which uses the new model (as Azure functions use connection string configs and this avoids config entry duplication).

### DIFF
--- a/src/NuGet.Services.Logging/ApplicationInsights.cs
+++ b/src/NuGet.Services.Logging/ApplicationInsights.cs
@@ -30,9 +30,9 @@ namespace NuGet.Services.Logging
         /// <paramref name="instrumentationKey"/>, taking into account the <c>ApplicationInsights.config</c> file if present.
         /// </summary>
         /// <param name="instrumentationKey">The instrumentation key to use.</param>
-        public static ApplicationInsightsConfiguration Initialize(string instrumentationKey)
+        public static ApplicationInsightsConfiguration Initialize(string instrumentationKey, bool isConnectionString = false)
         {
-            return InitializeApplicationInsightsConfiguration(instrumentationKey, heartbeatInterval: null);
+            return InitializeApplicationInsightsConfiguration(instrumentationKey, isConnectionString, heartbeatInterval: null);
         }
 
         /// <summary>
@@ -44,13 +44,15 @@ namespace NuGet.Services.Logging
         /// <param name="heartbeatInterval">The heartbeat interval to use.</param>
         public static ApplicationInsightsConfiguration Initialize(
             string instrumentationKey,
-            TimeSpan heartbeatInterval)
+            TimeSpan heartbeatInterval,
+            bool isConnectionString = false)
         {
-            return InitializeApplicationInsightsConfiguration(instrumentationKey, heartbeatInterval);
+            return InitializeApplicationInsightsConfiguration(instrumentationKey, isConnectionString, heartbeatInterval);
         }
 
         private static ApplicationInsightsConfiguration InitializeApplicationInsightsConfiguration(
             string instrumentationKey,
+            bool isConnectionString,
             TimeSpan? heartbeatInterval)
         {
             // Note: TelemetryConfiguration.Active is being deprecated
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -30,9 +30,9 @@ public static class ApplicationInsights
         /// <paramref name="instrumentationKey"/>, taking into account the <c>ApplicationInsights.config</c> file if present.
         /// </summary>
         /// <param name="instrumentationKey">The instrumentation key to use.</param>
-        public static ApplicationInsightsConfiguration Initialize(string instrumentationKey)
+        public static ApplicationInsightsConfiguration Initialize(string instrumentationKey, bool isConnectionString = false)
```
zhhyu, line 33: >instrumentationKey [](http://example.com/codeflow?start=73&length=18)

Will it be better if we rename the argument such as "instrumentationConnectionSecret"? Because now it's possible to pass a connection string rather than the instrumentation key as before.

Also, it's also better to update the comment otherwise it's misleading for callers.

src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -59,6 +61,16 @@ public static ApplicationInsightsConfiguration Initialize(string instrumentation
             // to take into account the ApplicationInsights.config file (if detected).
             var telemetryConfiguration = TelemetryConfiguration.CreateDefault();
 
+            // If instrumentationKey was passed as a connection string, we'll strip out the "InstrumentationKey=" portion here
+            if (isConnectionString)
+            {
+                var equalPos = instrumentationKey?.IndexOf('=') ?? -1;
```
zhhyu, line 67: I am not very sure, but is that possible that '=' exists in an instrumentation key?

src/NuGet.Services.Logging/ApplicationInsights.cs
dannyjdev: I think we can just add a regex or prefix search to know if the parameter is an Instrumentation Key or a Connection String, this will avoid having `isConnectionString` as parameter and adding it to config files as well.

src/NuGet.Services.Logging/ApplicationInsights.cs
drewgillies: We could certainly do this. I think I like this better than the regex idea for this reason. I was avoiding the specific wording on this, but give the specific format of AI connection strings it seems reasonable to bake it in. I'll proceed with this idea and remove the optional flag.

src/NuGet.Services.Logging/ApplicationInsights.cs
drewgillies: See below comment--thanks!

src/NuGet.Services.Logging/ApplicationInsights.cs
```diff
@@ -59,6 +61,16 @@ public static ApplicationInsightsConfiguration Initialize(string instrumentation
             // to take into account the ApplicationInsights.config file (if detected).
             var telemetryConfiguration = TelemetryConfiguration.CreateDefault();
 
+            // If instrumentationKey was passed as a connection string, we'll strip out the "InstrumentationKey=" portion here
```
shishirx34, line 64: If this is an expected format of the `ConnectionString` for an instrumentation key, I think you can make it easy by simply checking for the presence of this substring (`InstrumentationKey=`) and extract/replace the key as required. That way you do not have to actually break the functions or modify the callers to pass in the extra parameter. All you do is replace the config values where applicable and everything should still work. You'll have to update this dependency but that will only be it.

I did this for the usage of managed identities vs the connection string when migrating our service bus usage from SAS keys to managed identities. https://github.com/NuGet/ServerCommon/blob/main/src/NuGet.Services.ServiceBus/SubscriptionClientWrapper.cs#L32 

I've done this with the

src/NuGet.Services.Logging/ApplicationInsights.cs
shishirx34: Agree with @dannyjdev - see my comment below.

PR #405, opened 2022-11-17
Support Branch for PrivateBuildTools pull

ryuyu:
This should seamlessly enable using a new build time variable PRIVATE_BUILD_TOOL_BRANCH to pull private build tools by branch instead of commit.


zhhyu (approved):

mariaghiondea:
@ryuyu is this PR still needed?

ryuyu:
There was an explicit decision to not do this in case it broke any builds that weren't explicitly part of the update. Build tool updates should be a conscious decision by build I believe was the reasoning. I will close the PR.

PR #412, opened 2023-09-06
Move from WindowsAzure.ServiceBus to Azure.Messaging.ServiceBus

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/5011

Summary of changes:
1. Remove WindowsAzure.ServiceBus dependency, add Azure.Messaging.ServiceBus dependency
2. Introduce `IReceivedBrokeredMessage`. This mirrors a similar split in the send and receive message models in the new SDK.
   - This is a better data model IMO since some things are immutable on received messages and the models can express that.
3. Add several shim-behaviors, to ensure full back-compat
   - Generate a random message ID on the client side (new SDK doesn't so this, partitioned topics require this)
   - Add support for `sb://namespace/` style connection strings (old SDK supports it, new SDK does not)
   - Serialize message content using XML data contract. I dislike this but we must do this to handle messages sent by old SDK

Testing strategy:
1. Full deployment to DEV
    - Gallery, internal jobs, OSS jobs
    - Test send-receive compat (sender on new SDK, receiver on old SDK and vice versa)
    - Push steady stream of validations and reboot VMs (non-graceful message handling)
2. Full deployment in INT
3. PROD strategy:
    - Deploy single region of SB listeners and observe telemetry
    - Wait 48 hours
    - Deploy 2nd region of SB listeners
    - Wait 24 hours
    - Deploy gallery


mariaghiondea (dismissed):
Thank you for the work on this one!
Anything that we can do to prevent regressions, especially anything that would come up at runtime?

joelverhagen:
> Thank you for the work on this one! Anything that we can do to prevent regressions, especially anything that would come up at runtime?

I've updates the PR description with details. In short: full deployment of all SB-interacting component in DEV/INT prior to PROD (as expected) and a careful deployment in PROD where we move listeners slowly and services (i.e. senders = gallery) later.

src/NuGet.Services.ServiceBus/OnMessageOptionsWrapper.cs
```diff
@@ -24,11 +24,11 @@ public sealed class OnMessageOptionsWrapper : IOnMessageOptions
         /// </remarks>
         public int MaxConcurrentCalls { get; set; } = 1;
 
-        public OnMessageOptions GetOptions()
+        public ServiceBusProcessorOptions GetOptions()
```
agr, line 27: Can we change it to `internal`?

src/NuGet.Services.ServiceBus/SubscriptionClientWrapper.cs
```diff
@@ -2,88 +2,126 @@
...
             };
         }
 
-        public Task CloseAsync()
+        public async Task CloseAsync()
         {
-            return _client.CloseAsync();
+            if (_processor != null)
+            {
+                await _processor.CloseAsync();
```
agr, line 121: Do we need to remove event handlers added in `StartProcessingAsync` here?

tests/NuGet.Services.Messaging.Tests/ServiceBusMessageSerializerTests.cs
```diff
@@ -110,7 +110,11 @@ public void RejectsInvalidType()
             {
                 // Arrange
                 var brokeredMessage = GetBrokeredMessage();
-                brokeredMessage.Object.Properties[SchemaName] = "bad";
+                brokeredMessage.Setup(x => x.Properties).Returns(new Dictionary<string, object>
```
agr, line 113: This begs for a helper function.

agr (dismissed):
Feel free to ignore my comments, nothing important there.

src/NuGet.Services.ServiceBus/OnMessageOptionsWrapper.cs
joelverhagen: Yes, I think we can.

tests/NuGet.Services.Messaging.Tests/ServiceBusMessageSerializerTests.cs
joelverhagen: I gave this a try, but I found the variants too many (missing key, different value, etc) to make it not worth it IMO.

mariaghiondea (approved):

PR #414, opened 2023-09-08
Transition from Microsoft.Azure.KeyVault to Azure.Security.KeyVault

erdembayar:
Fixes: https://github.com/NuGet/Engineering/issues/4704

Transition from `Microsoft.Azure.KeyVault` to `Azure.Security.KeyVault` package dependency. Former one is deprecated.
I manually tested this change in [Gallery dev environment](https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-pipeline-progress&releaseId=1478870)

### DIFF
--- a/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
@@ -12,6 +12,7 @@ namespace NuGet.Services.Configuration
     {
         private string _vaultName;
         private bool _useManagedIdentity;
+        private string _tenantId;
         private string _clientId;
         private string _certificateThumbprint;
         private string _storeName;
@@ -34,6 +35,7 @@ namespace NuGet.Services.Configuration
                 _useManagedIdentity = bool.Parse(useManagedIdentity);
             }
 
+            _tenantId = config[Constants.KeyVaultTenanIdKey];
             _clientId = config[Constants.KeyVaultClientIdKey];
             _certificateThumbprint = config[Constants.KeyVaultCertificateThumbprintKey];
             if (_useManagedIdentity && IsCertificateConfigurationProvided())
@@ -83,7 +85,8 @@ namespace NuGet.Services.Configuration
                     _validateCertificate);
 
                 keyVaultConfiguration = new KeyVaultConfiguration(
-                    _vaultName,                
+                    _vaultName,
+                    _tenantId,
                     _clientId,
                     certificate,
                     _sendX5c);
@@ -100,7 +103,8 @@ namespace NuGet.Services.Configuration
         private bool IsCertificateConfigurationProvided()
         {
             return !string.IsNullOrEmpty(_clientId)
-                || !string.IsNullOrEmpty(_certificateThumbprint);
+                || !string.IsNullOrEmpty(_certificateThumbprint)
+                || !string.IsNullOrEmpty(_tenantId);
         }
     }
 }
--- a/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
+++ b/src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
@@ -70,7 +70,7 @@ namespace NuGet.Services.Configuration
 
             if (_useManagedIdentity)
             {
-                keyVaultConfiguration = new KeyVaultConfiguration(_vaultName);
+                keyVaultConfiguration = new KeyVaultConfiguration(_vaultName, _clientId);
             }
             else
             {
--- a/src/NuGet.Services.Configuration/Constants.cs
+++ b/src/NuGet.Services.Configuration/Constants.cs
@@ -7,6 +7,7 @@ namespace NuGet.Services.Configuration
     {
         public static string KeyVaultVaultNameKey = "KeyVault_VaultName";
         public static string KeyVaultUseManagedIdentity = "KeyVault_UseManagedIdentity";
+        public static string KeyVaultTenanIdKey = "KeyVault_TenantId";
         public static string KeyVaultClientIdKey = "KeyVault_ClientId";
         public static string KeyVaultCertificateThumbprintKey = "KeyVault_CertificateThumbprint";
         public static string KeyVaultValidateCertificateKey = "KeyVault_ValidateCertificate";
--- a/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
+++ b/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
@@ -10,6 +10,7 @@ namespace NuGet.Services.KeyVault
     {
         public string VaultName { get; }
         public bool UseManagedIdentity { get; }
+        public string TenantId { get; }
         public string ClientId { get; }
         public X509Certificate2 Certificate { get; }
         public bool SendX5c { get; }
@@ -32,10 +33,11 @@ namespace NuGet.Services.KeyVault
         /// The constructor for keyvault configuration when using the certificate
         /// </summary>
         /// <param name="vaultName">The name of the keyvault</param>
+        /// <param name="tenantId">AAD tenant ID where respective client app is registered.</param>
         /// <param name="clientId">Keyvault client id</param>
         /// <param name="certificate">Certificate required to access the keyvault</param>
         /// <param name="sendX5c">SendX5c property</param>
-        public KeyVaultConfiguration(string vaultName, string clientId, X509Certificate2 certificate, bool sendX5c = false)
+        public KeyVaultConfiguration(string vaultName, string tenantId, string clientId, X509Certificate2 certificate, bool sendX5c = false)
         {
             if (string.IsNullOrWhiteSpace(vaultName))
             {
@@ -48,6 +50,7 @@ namespace NuGet.Services.KeyVault
             }
 
             UseManagedIdentity = false;
+            TenantId = tenantId;
             VaultName = vaultName;
             ClientId = clientId;
             Certificate = certificate ?? throw new ArgumentNullException(nameof(certificate));
--- a/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
+++ b/src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
@@ -18,7 +18,7 @@ namespace NuGet.Services.KeyVault
         /// <summary>
         /// The constructor for keyvault configuration when using managed identities
         /// </summary>
-        public KeyVaultConfiguration(string vaultName)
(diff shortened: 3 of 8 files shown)


src/NuGet.Services.KeyVault/KeyVaultConfiguration.cs
```diff
@@ -10,14 +10,15 @@ public class KeyVaultConfiguration
...
+        public string TenantId { get; }
         public string ClientId { get; }
         public X509Certificate2 Certificate { get; }
         public bool SendX5c { get; }
 
         /// <summary>
         /// The constructor for keyvault configuration when using managed identities
         /// </summary>
-        public KeyVaultConfiguration(string vaultName)
+        public KeyVaultConfiguration(string vaultName, string clientId = null)
```
joelverhagen, line 21: nit: adding optional parameters is not runtime back-compatible, just source back-compatible. A more proper way is to add a ctor overload.

src/NuGet.Services.KeyVault/KeyVaultConfigurationExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.KeyVault
+{
+    public static class KeyVaultConfigurationExtensions
+    {
+        public static Uri GetKeyVaultUri(this KeyVaultConfiguration self)
```
joelverhagen, line 14: This appears to be used in one place. Perhaps we can just make it a private helper method?

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -43,7 +39,8 @@ public async Task<string> GetSecretAsync(string secretName)
 
         public async Task<string> GetSecretAsync(string secretName, ILogger logger)
         {
-            var secret = await _keyVaultClient.Value.GetSecretAsync(_vault, secretName);
+            var response = await _keyVaultClient.Value.GetSecretAsync(secretName);
```
joelverhagen, line 42: The idiomatic way to do this in Azure SDK is with by using the implicit converter.
```suggestion
            KeyVaultSecret response = await _keyVaultClient.Value.GetSecretAsync(secretName);
```
And then eliminating this line: `var secret = response.Value;`

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -54,35 +51,31 @@ public async Task<ISecret> GetSecretObjectAsync(string secretName)
 
         public async Task<ISecret> GetSecretObjectAsync(string secretName, ILogger logger)
         {
-            var secret = await _keyVaultClient.Value.GetSecretAsync(_vault, secretName);
-            return new KeyVaultSecret(secretName, secret.Value, secret.Attributes.Expires);
+            var response = await _keyVaultClient.Value.GetSecretAsync(secretName);
```
joelverhagen, line 54: Same comment as above. Don't use `var` for Azure SDK API response unless you want to interact with the response message (most cases you don't)

src/NuGet.Services.KeyVault/NuGet.Services.KeyVault.csproj
```diff
@@ -6,9 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.KeyVault">
-      <Version>3.0.5</Version>
-    </PackageReference>
+    <PackageReference Include="Azure.Identity" Version="1.8.0" />
+    <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.4.0" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication">
```
joelverhagen, line 11: Is this dependency still needed?

joelverhagen (changes requested):
Minor comments, looks good overall

src/NuGet.Services.KeyVault/KeyVaultConfigurationExtensions.cs
erdembayar: Addressed.

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
```diff
@@ -100,7 +103,8 @@ public ISecretInjector CreateSecretInjector(ISecretReader secretReader)
         private bool IsCertificateConfigurationProvided()
         {
             return !string.IsNullOrEmpty(_clientId)
```
joelverhagen, line 105: Can't `_clientId` be set when `_useManagedIdentity` is true? I'm wondering why the call site of this function isn't throwing the exception right now.

joelverhagen (dismissed):
:shipit: please get @agr's approval too since he's the original author

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
erdembayar: @agr 

Do you know about this logic? If we set both `_clientId` and `_useManagedIdentity` then it should throw. Somehow it doesn't when I'm testing. Does it mean we never set them together?

src/NuGet.Services.Configuration/Constants.cs
```diff
@@ -7,6 +7,7 @@ public static class Constants
     {
         public static string KeyVaultVaultNameKey = "KeyVault_VaultName";
         public static string KeyVaultUseManagedIdentity = "KeyVault_UseManagedIdentity";
+        public static string KeyVaultTenanIdKey = "KeyVault_TenantId";
```
agr, line 10: Type in the variable name.

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
agr: Assuming it is actually used, stepping through it with a debugger should get you the answer :)

agr: Is "KeyVault.ClientId" configuration is actually specified in the environment you test in?

src/NuGet.Services.KeyVault/KeyVaultReader.cs
```diff
@@ -43,8 +40,8 @@ public async Task<string> GetSecretAsync(string secretName)
 
         public async Task<string> GetSecretAsync(string secretName, ILogger logger)
         {
-            var secret = await _keyVaultClient.Value.GetSecretAsync(_vault, secretName);
-            return secret.Value;
+            AzureSecurityKeyVaultSecret response = await _keyVaultClient.Value.GetSecretAsync(secretName);
```
agr, line 43: Why variable name was changed?

erdembayar: Here is the context: https://github.com/NuGet/ServerCommon/pull/414#discussion_r1320058717

agr (dismissed):

src/NuGet.Services.KeyVault/KeyVaultReader.cs
agr: The variable should be renamed to `secret`. In the method below, too.

erdembayar: Addressed

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
agr: OK, secret reader initialization is weird in Gallery:

In one [place](https://github.com/NuGet/NuGetGallery/blob/c835b2712e255869808a00e1f5a0f3b38bc972b6/tests/NuGetGallery.FunctionalTests.Core/GalleryConfiguration.cs#L48) it uses `ConfigurationRootSecretReaderFactory`, which maybe reading wrong configuration variables.

In another it very explicitly [constructs](https://github.com/NuGet/NuGetGallery/blob/c835b2712e255869808a00e1f5a0f3b38bc972b6/src/NuGetGallery.Services/Configuration/SecretReader/SecretReaderFactory.cs#L57C30-L57C30) `KeyVaultConfiguration` and then creates `SecretReader` bypassing the code above.

agr: Ah, scratch that, the first use is not in Gallery. But that also means, we need to fix the second usage (or figure out how does it work).

agr: But that also means that the usage of configuration object in functional tests had to be checked, too.

erdembayar: I don't see `AZURE_CLIENT_ID` in env var list of dev, but following 2 vars are set.
image

erdembayar: Is code path ever exercised? I found somewhat related issue: https://github.com/NuGet/NuGetGallery/issues/4681

agr: Gallery constructs `KeyVaultReader` directly in `SecretReaderFactory` and that usage has been updated in https://github.com/NuGet/NuGetGallery/pull/9654. `ConfigurationRootSecretReaderFactory` is not used in Gallery.

It is, however, used in functional tests and that usage needs to be checked.

erdembayar: I checked the usage of `ConfigurationRootSecretReaderFactory` using the dev env config file for functional tests. It was passed using the `uninjectedConfiguration` object, as you can see in the below screenshot. It was initiated with `KeyVault.UseManagedIdentity = True` and `KeyVault.VaultName`, but `KeyVault.ClientId` was not set. That's why function tests didn't throw. Considering this code path is only for unit tests, I don't believe it's a blocker for this PR.
image

@joelverhagen @agr 
Could you please approve this PR again?

joelverhagen (approved):

src/NuGet.Services.Configuration/ConfigurationRootSecretReaderFactory.cs
joelverhagen: Okay I understand. I think it would be better to not check client ID in `IsCertificateConfigurationProvided` now so we're more tolerant of future cases.

erdembayar: Sure, removed.

joelverhagen (approved):

PR #421, opened 2024-02-15
Add error code for ScanOperationStates table

erdembayar:
Addresses: https://github.com/NuGet/Engineering/issues/5250
Adding 2 new columns to persist extra details about scan operation


drewgillies (dismissed):

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -66,6 +66,16 @@ public class ScanOperationState
         /// </summary>
         public string OperationId { get; set; }
 
+        /// <summary>
+        /// Operation errors code.
+        /// </summary>
+        public int ErrorCode { get; set; }
```
joelverhagen, line 72: does this actually need to be an int? Would a string be more flexible and get the job done?

erdembayar: Sure, addressed it.

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -66,6 +66,16 @@ public class ScanOperationState
         /// </summary>
         public string OperationId { get; set; }
 
+        /// <summary>
+        /// Operation errors code.
```
agr, line 70:
```suggestion
        /// Operation error code.
```

src/NuGet.Services.Validation/Entities/ScanOperationState.cs
```diff
@@ -66,6 +66,16 @@ public class ScanOperationState
...
+        /// </summary>
+        public string ErrorCode { get; set; }
+
+        /// <summary>
+        /// Any applicable operation details in text/json format.
```
agr, line 75: "text/json" here means "plain text or json" as far as I understand, but can be interpreted as MIME type string. I'd spell it out more explicitly to not create any confusion about potential content.

agr (approved):
Please, fix comments. Otherwise, looks good.

agr (approved):

PR #423, opened 2024-02-24
`GetRawBody` method

agr:
One of the monitoring jobs requires access to raw message body data to properly deserialize the message. Calling `BrokeredMessage.GetBody<Stream>` worked with the previous version of the Service Bus library. Migrating to new version broke that (as there is no direct equivalent and our reimplementation disregarded that case). Adding explicit method to support that scenario.

Another potential solution here is to update `GetBody<TStream>` implementation to check if `TStream` is `System.IO.Stream` and do something else, but I'm not sure what else it might break. Creating new method is safer.

### DIFF
--- a/src/NuGet.Services.Contracts/ServiceBus/IReceivedBrokeredMessage.cs
+++ b/src/NuGet.Services.Contracts/ServiceBus/IReceivedBrokeredMessage.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Threading.Tasks;
 
 namespace NuGet.Services.ServiceBus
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.Contracts/ServiceBus/IReceivedBrokeredMessage.cs
```diff
@@ -20,5 +21,6 @@ public interface IReceivedBrokeredMessage
         Task AbandonAsync();
         string GetBody();
         TStream GetBody<TStream>();
+        Stream GetRawBody();
```
sbanni, line 24: This is a breaking change on a public interface, do we need to be concerned about that here?

mariaghiondea (approved):

sbanni (approved):

PR #430, opened 2024-06-17
[Storage]Update ServerCommon library to v12 storage SDKs

ryuyu:
Updating the server common libraries to stop using the deprecated storage libraries and upgrade to the new v12 Azure storage SDK.

Note that this PR is currently a DRAFT for feedback purposes.


src/NuGet.Services.Storage/IStorage.cs
```diff
@@ -18,7 +19,7 @@ public interface IStorage
         Task<string> LoadString(Uri resourceUri, CancellationToken cancellationToken);
```
erdembayar, line 19: Save, Load, Delete, LoadString methods needs to be async because they use async calls inside them and we pass cancellation token.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -158,63 +167,56 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
-            CloudBlockBlob sourceBlob = GetBlockBlobReference(sourceName);
-            CloudBlockBlob destinationBlob = azureDestinationStorage.GetBlockBlobReference(destinationName);
-
-            var context = new SingleTransferContext();
+            var destinationOptions = new BlobStorageResourceOptions();
 
             if (destinationProperties?.Count > 0)
             {
-                context.SetAttributesCallbackAsync = new SetAttributesCallbackAsync((destination) =>
+                foreach (var property in destinationProperties)
```
erdembayar, line 174: Explicit type please, not clear type we're iterating through

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -158,63 +167,56 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
-                    return Task.CompletedTask;
-                });
+                        default:
+                            throw new NotImplementedException($"Storage property '{property.Value}' is not supported.");
+                    }
+                }
             }
 
-            context.ShouldOverwriteCallbackAsync = new ShouldOverwriteCallbackAsync((source, destination) => Task.FromResult(true));
+            var sourceStorageResource = _storageResourceProvider.FromBlob(sourceUri.ToString());
```
erdembayar, line 192: Not sure what type of value returned here, explicit type please.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -158,63 +167,56 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+            var destinationStorageResource = _storageResourceProvider.FromBlob(destinationUri.ToString(), new AppendBlobStorageResourceOptions());
```
erdembayar, line 193: Here too

src/NuGet.Services.Storage/AzureStorageFactory.cs
```diff
@@ -2,14 +2,17 @@
...
 using Microsoft.Extensions.Logging;
-using Microsoft.WindowsAzure.Storage;
 
 namespace NuGet.Services.Storage
 {
     public class AzureStorageFactory : StorageFactory
     {
-        CloudStorageAccount _account;
+        BlobServiceClient _account;
+        BlobsStorageResourceProvider _storageResourceProvider;
```
erdembayar, line 15: What is this `BlobsStorageResourceProvider`? What is the actual use of it?

src/NuGet.Services.Storage/IStorage.cs
ryuyu: The async keyword isn't allowed on the interface here apparently. I believe they are marked async at implementation point like in src/NuGet.Services.Storage/Storage.cs

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: We are iterating over a dictionary. I don't mind explicitly putting KeyValuePair<string,string> here if you think that's more clear, but it seems overly verbose to me.

src/NuGet.Services.Storage/AzureStorageFactory.cs
ryuyu: It is used in the copy scenario as the TransferManager needs storageResources, which need a storageResourceProvider to retrieve. This is a direct replacement of the transfer/copy implementation that was previously used as this is the implementation in the new v12 storage sdk.

src/NuGet.Services.Status.Table/ChildEntity.cs
```diff
@@ -1,27 +1,36 @@
...
 {
     /// <summary>
     /// Base implementation of <see cref="IChildEntity{T}"/>.
     /// </summary>
-    public class ChildEntity<TParent> : TableEntity, IChildEntity<TParent>
-        where TParent : TableEntity
+    public class ChildEntity<TParent> : ITableEntity, IChildEntity<TParent>
+        where TParent : ITableEntity
     {
+        private ITableEntity _tableEntity;
```
agr, line 20: What's the role of `_tableEntity`?

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -4,48 +4,52 @@
...
                 await queue.CreateIfNotExistsAsync();
                 return queue;
             });
         }
 
         public async Task AddAsync(string contents, CancellationToken token)
         {
-            var azureMessage = new AzureStorageQueueMessage(contents, dequeueCount: 0);
-            await (await _queueTask.Value).AddMessageAsync(azureMessage.Message, token);
+            //var azureMessage = new AzureStorageQueueMessage(contents, dequeueCount: 0);
```
agr, line 39: remove comment?

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -62,14 +66,21 @@ public async Task RemoveAsync(StorageQueueMessage message, CancellationToken tok
                 throw new ArgumentException("This message was not returned from this queue!", nameof(message));
             }
 
-            await (await _queueTask.Value).DeleteMessageAsync((message as AzureStorageQueueMessage).Message, token);
+            var queueClient = await _queueTask.Value;
+            if (await queueClient.ExistsAsync())
+            {
+                AzureStorageQueueMessage queueMessage = message as AzureStorageQueueMessage;
```
agr, line 72: `queueMessage` can end up being `null` here, then the next line would fail with NPE.

jimmylewis: we know it's not null from L64.  But using a pattern match there would avoid the redundant type check here:

```
            if (!(message is AzureStorageQueueMessage queueMessage))
            {
                throw new ArgumentException("This message was not returned from this queue!", nameof(message));
            }
```

Now it doesn't need to cast to AzureStorageQueueMessage a second time.

src/NuGet.Services.Status.Table/ChildEntity.cs
erdembayar: If we can get rid of this then that would be great.

src/NuGet.Services.Storage/AzureStorageFactory.cs
erdembayar: Does it have to be created with certain context or just create new instance with `new BlobsStorageResourceProvider()` passing is good enough?

src/NuGet.Services.Storage/CloudBlobStorageExtensions.cs
```diff
@@ -4,34 +4,53 @@
...
-                var segment = await directory.ListBlobsSegmentedAsync(
-                    useFlatBlobListing: true,
-                    blobListingDetails: getMetadata ? BlobListingDetails.Metadata : BlobListingDetails.None,
-                    maxResults: null,
-                    currentToken:  continuationToken,
-                    options: null,
-                    operationContext: null,
-                    cancellationToken: cancellationToken);
+            var items = new List<BlobItem>();
+            //var segment = await directory.GetBlobsAsync(
```
erdembayar, line 20: Remove commented code?

src/NuGet.Services.Storage/AzureStorage.cs
erdembayar: If you already have context then it's verbose but if you don't then you need to open VS to see what is the actual type here, so prefer explicit type.

src/NuGet.Services.Status.Table/ChildEntity.cs
```diff
@@ -45,5 +58,6 @@ public bool IsLinked
             get { return !string.IsNullOrEmpty(ParentRowKey); }
             set { }
         }
+
```
erdembayar, line 61: nit: remove empty line.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -49,14 +59,19 @@ public class AzureStorage : Storage
...
+            var transferManagerOptions = new TransferManagerOptions();
+            transferManagerOptions.CheckpointerOptions = checkpointerOptions;
+            _transferManager = new TransferManager(transferManagerOptions);
+            _storageResourceProvider = blobsStorageResourceProvider;
 
-            if (initializeContainer && _directory.Container.CreateIfNotExists())
+            if (initializeContainer)
             {
-                _directory.Container.SetPermissions(new BlobContainerPermissions { PublicAccess = BlobContainerPublicAccessType.Blob });
+                _directory.CreateIfNotExists(PublicAccessType.Blob);
```
erdembayar, line 70: Should the line 68 be `if (initializeContainer) && _directory.CreateIfNotExists(PublicAccessType.Blob != null`)
Here this is the return of CreateIfNotExists
    // Returns:
    //     If the container does not already exist, a Azure.Response`1 describing the newly
    //     created container. If the container already exists, null.
    //

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -309,23 +313,24 @@ protected override async Task OnDelete(Uri resourceUri, CancellationToken cancel
...
 
             return blob;
         }
 
-        private void ApplyBlobRequestOptions(CloudBlockBlob blob)
+        private Uri ResolvePathedUri(string filename)
         {
-            blob.ServiceClient.DefaultRequestOptions = _directory.ServiceClient.DefaultRequestOptions;
+            return ResolveUri(Path.Combine(_path, filename));
         }
```
erdembayar, line 334: Could add some comment or documentation on how exactly this `ResolvePathedUri` works?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -83,10 +98,10 @@ static Uri GetDirectoryUri(CloudBlobDirectory directory)
         //Blob exists
         public override bool Exists(string fileName)
         {
-            var packageRegistrationUri = ResolveUri(fileName);
+            var packageRegistrationUri = ResolvePathedUri(fileName);
```
erdembayar, line 101: Not blocking, feel free to resolve it, if you include explicit type then it would be great for this line and following ones in this file, just reading the code not sure what type was returned.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +132,27 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
             return false;
         }
 
-        public override async Task<IEnumerable<StorageListItem>> List(bool getMetadata, CancellationToken cancellationToken)
+        public override IEnumerable<StorageListItem> List(bool getMetadata)
         {
-            var files = await _directory.ListBlobsAsync(getMetadata, cancellationToken);
-
-            return files.Select(GetStorageListItem).AsEnumerable();
+            foreach (BlobItem blob in _directory.GetBlobs(prefix: _path))
```
erdembayar, line 137: Should we use async `GetBlobsAsync`?

erdembayar: Is new code actually logically equivalent to old code?
`CoPilot` suggests below code as replacement for previous code.

```
using System;
using System.Collections.Generic;
using System.Threading;
using System.Threading.Tasks;
using Azure;
using Azure.Storage.Blobs;
using Azure.Storage.Blobs.Models;

public static class BlobStorageExtensions
{
    public static async Task<IEnumerable<BlobItem>> ListBlobsAsync(
        this BlobContainerClient containerClient,
        string directoryPath,
        bool getMetadata,
        CancellationToken cancellationToken)
    {
        var items = new List<BlobItem>();
        var options = new BlobTraits();
        if (getMetadata)
        {
            options |= BlobTraits.Metadata;
        }

        var blobHierarchyItems = containerClient.GetBlobsByHierarchyAsync(
            traits: options,
            states: BlobStates.None,
            prefix: directoryPath,
            delimiter: "/",
            cancellationToken: cancellationToken);

        await foreach (var blobHierarchyItem in blobHierarchyItems)
        {
            if (blobHierarchyItem.IsBlob)
            {
                items.Add(blobHierarchyItem.Blob);
            }
        }

        return items;
    }
}

// Usage
public async Task ListBlobsInDirectoryAsync()
{
    string connectionString = "your_connection_string";
    string containerName = "your_container_name";
    string directoryPath = "your_directory_path";
    bool getMetadata = true;
    CancellationToken cancellationToken = CancellationToken.None;

    var containerClient = new BlobContainerClient(connectionString, containerName);
    var files = await containerClient.ListBlobsAsync(directoryPath, getMetadata, cancellationToken);

    // Process files as needed
    foreach (var file in files)
    {
        Console.WriteLine(file.Name);
    }
}

```

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +132,27 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
         {
-            var blob = GetBlockBlobReference(GetName(resourceUri));
-
-            foreach (var kvp in metadata)
-            {
-                blob.Metadata[kvp.Key] = kvp.Value;
-            }
-
-            await blob.SetMetadataAsync();
+            var blob = GetBlobReference(GetName(resourceUri));
```
erdembayar, line 145: I had to open VS to see when this variable actually is? I would appreciate explicit type.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +132,27 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
+            await blob.SetMetadataAsync(metadata);
         }
 
-        private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
+        private StorageListItem GetStorageListItem(BlobClient listBlobItem)
         {
-            var cloudBlockBlob = (listBlobItem as CloudBlob);
-            var lastModified = cloudBlockBlob?.Properties.LastModified?.UtcDateTime;
+            var blobPropertiesResponse = listBlobItem.GetProperties();
```
erdembayar, line 151: Could we use async  method here `GetPropertiesAsync` and bubble up async call for perf benefit?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +132,27 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
+            var blobProperties = blobPropertiesResponse.Value;
```
erdembayar, line 152: Previous logic gracefully handle possible null object value case. Just accessing `.Value` here might throw an exception.
```
    // Summary:
    //     Gets the value returned by the service. Accessing this property will throw if
    //     Azure.NullableResponse`1.HasValue is false.
```

src/NuGet.Services.Storage/AzureStorageFactory.cs
ryuyu: If you don't plan to use the copy scenario, an empty resource provider should be fine to use.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +132,27 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
+            var lastModified = blobProperties.LastModified;
 
-            return new StorageListItem(listBlobItem.Uri, lastModified, cloudBlockBlob?.Metadata);
+            return new StorageListItem(listBlobItem.Uri, lastModified, blobProperties.Metadata);
```
erdembayar, line 155: Same here null conditional operator is needed.

src/NuGet.Services.Storage/AzureStorageQueueMessage.cs
```diff
@@ -1,24 +1,30 @@
...
             Message = message;
+            MessageId = message.MessageId;
+            PopReceipt = message.PopReceipt;
         }
 
         internal AzureStorageQueueMessage(string contents, int dequeueCount)
             : base(contents, dequeueCount)
         {
-            Message = new CloudQueueMessage(contents);
+            Message = null;
```
erdembayar, line 27: Why `null`?

src/NuGet.Services.Storage/CloudBlobStorageExtensions.cs
```diff
@@ -4,34 +4,53 @@
...
+            //    useFlatBlobListing: true,
+            //    blobListingDetails: getMetadata ? BlobListingDetails.Metadata : BlobListingDetails.None,
+            //    maxResults: null,
+            //    currentToken:  continuationToken,
+            //    options: null,
+            //    operationContext: null,
+            //    cancellationToken: cancellationToken);
+            var segment = directory.GetBlobsAsync(
```
erdembayar, line 28: CoPilot give me this suggestion which is bit simpler:

```
using System;
using System.Collections.Generic;
using System.Threading;
using System.Threading.Tasks;
using Azure;
using Azure.Storage.Blobs;
using Azure.Storage.Blobs.Models;

public static class BlobStorageExtensions
{
    public static async Task<IEnumerable<BlobItem>> ListBlobsAsync(
        this BlobContainerClient containerClient,
        string directoryPath,
        bool getMetadata,
        CancellationToken cancellationToken)
    {
        var items = new List<BlobItem>();
        var traits = getMetadata ? BlobTraits.Metadata : BlobTraits.None;

        await foreach (var blobItem in containerClient.GetBlobsAsync(traits: traits, prefix: directoryPath, cancellationToken: cancellationToken))
        {
            items.Add(blobItem);
        }

        return items;
    }
}

// Usage
public async Task ListBlobsInDirectoryAsync()
{
    string connectionString = "your_connection_string";
    string containerName = "your_container_name";
    string directoryPath = "your_directory_path/";
    bool getMetadata = true;
    CancellationToken cancellationToken = CancellationToken.None;

    var containerClient = new BlobContainerClient(connectionString, containerName);
    var files = await containerClient.ListBlobsAsync(directoryPath, getMetadata, cancellationToken);

    // Process files as needed
    foreach (var file in files)
    {
        Console.WriteLine(file.Name);
    }
}

```

erdembayar:
LGTM, left some comments. Sorry for late review.

src/NuGet.Services.Status.Table/ChildEntity.cs
ryuyu: I think this may be left over from the original attempt to wrap all SDK classes to prevent interface leakage. I'll look into seeing how much of this i can revert without breaking other things.

src/NuGet.Services.Storage/AzureStorageQueue.cs
ryuyu: done

src/NuGet.Services.Storage/AzureStorageQueue.cs
ryuyu: It looks like the "is not" pattern requires c# 9.0 and is not available in c#7.3
Are there any potential caveats to "upgrading" language version here?

ryuyu: Is
if (!(message is AzureStorageQueueMessage queueMessage))

functionally identical?

src/NuGet.Services.Storage/CloudBlobStorageExtensions.cs
ryuyu: Done

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: I think this is functionally ok.
If the container exists, we don't need to do anything anyway, and the null return is ignored.
If the container didn't exist, we create it.

I think this is the behaviour we want. I will update the logging lines after to note if the container was actually created, or if the operation no-oped.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: I think we can leverage the blobtraits.Metadata part in options to make the metadata get boolean actually do something , but I think we would need to upgrade language to get access to the await foreach. I remember trying to use this before and failing.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: I added some async methods versions.

src/NuGet.Services.Storage/AzureStorageQueue.cs
ryuyu: Oh it was pretty easy to at least have IDE pick up new language version. I will see if it still builds correctly soon.

src/NuGet.Services.Storage/CloudBlobStorageExtensions.cs
ryuyu: I listed all of the Traits and Properties here somewhat intentionally so it was clear to the next person touching this file what our various options to pass through are (if we ever choose to extend it to be able to pass through other parameters).

This was also done this way because await foreach was a language feature that wasn't available in c#7.0.

Since I pulled up the language version, I'll adapt some of these changes.

src/NuGet.Services.Storage/AzureStorageQueueMessage.cs
ryuyu: Digging in to this a little bit. I'm not sure we need this at all (both the constructor and the Message internal variable. I can't tell what it is used for, and locally, this contructor is completely unused. The Message Internal variable also appears to be unused, and appears to be a wrapper enclosing the original QueueMessage.

src/NuGet.Services.Status.Table/ChildEntity.cs
ryuyu: Digging througha b it more, it looks liek this doesn't serve any purpose right now. I'll get rid of it.

src/NuGet.Services.Status.Table/ChildEntity.cs
```diff
@@ -1,25 +1,28 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.WindowsAzure.Storage.Table;
+using Azure;
+
+// Copyright (c) .NET Foundation. All rights reserved.
```
erdembayar, line 6: Did you copy Copyright notice twice by accident?

src/NuGet.Services.Status.Table/ChildEntity.cs
```diff
@@ -44,6 +51,6 @@ public bool IsLinked
         {
             get { return !string.IsNullOrEmpty(ParentRowKey); }
             set { }
-        }
+        }
```
erdembayar, line 54: Unnecessary, empty space here.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +139,60 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
+                blobList.Add(await GetStorageListItemAsync(_directory.GetBlobClient(blob.Blob.Name)));
             }
 
-            await blob.SetMetadataAsync();
+            return blobList;
+        }
+
+        public override async Task SetMetadataAsync(Uri resourceUri, IDictionary<string, string> metadata)
+        {
+            BlobClient blob = GetBlobReference(GetName(resourceUri));
```
erdembayar, line 176: Isn't it would be [BlockBlobClient](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.specialized.blockblobclient?view=azure-dotnet) because we would mostly work with large blob storage?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -158,63 +209,56 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
             }
 
-            string sourceName = GetName(sourceUri);
-            string destinationName = azureDestinationStorage.GetName(destinationUri);
-
-            CloudBlockBlob sourceBlob = GetBlockBlobReference(sourceName);
-            CloudBlockBlob destinationBlob = azureDestinationStorage.GetBlockBlobReference(destinationName);
-
-            var context = new SingleTransferContext();
+            var destinationOptions = new BlobStorageResourceOptions();
```
erdembayar, line 212: This var is created, but not used, see my PR https://github.com/NuGet/NuGet.Jobs/pull/1210 >> src/Catalog/Persistence/AzureStorage.cs>> around Line 271.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -226,33 +270,31 @@ protected override async Task OnSave(Uri resourceUri, StorageContent content, bo
 
                     destinationStream.Seek(0, SeekOrigin.Begin);
 
-                    await blob.UploadFromStreamAsync(
+                    await blob.SetHttpHeadersAsync(headers);
```
erdembayar, line 273: Should we upload first and then set the SetHttpHeadersAsync? Not sure if it matters. What happens if we set SetHttpHeadersAsync and then the next upload fails, or vice versa?

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -309,23 +353,25 @@ protected override async Task OnDelete(Uri resourceUri, CancellationToken cancel
...
 
-            await blob.DeleteAsync(cancellationToken);
+            await blob.DeleteAsync(cancellationToken: cancellationToken);
         }
 
-        private CloudBlockBlob GetBlockBlobReference(string blobName)
+        private BlobClient GetBlobReference(string blobName)
         {
-            var blob = _directory.GetBlockBlobReference(blobName);
+            var blob = _directory.GetBlobClient(blobName);
```
erdembayar, line 363: Should it be `BlockBlobClient`?
@drewgillies  @agr 
What is your opinion on BlobClient vs BlockBlobClient? I used BlockBlobClient in my https://github.com/NuGet/NuGet.Jobs/pull/1210/files

tests/NuGet.Services.Storage.Tests/AzureStorageFacts.cs
```diff
@@ -6,125 +6,128 @@
...
-            public async Task AllowsOverwriteWhenDesired()
-            {
-                // Arrange
-                await CreateBlobAsync(_dirPath, BlobName);
-
-                var inputContent = new StringStorageContent("new content");
-
-                // Act
-                await _target.Save(_resourceUri, inputContent, overwrite: true, cancellationToken: CancellationToken.None);
+        //private const string NoEmulator = "The Azure Storage emulator is not running on the CI.";
```
erdembayar, line 20: Are we going to fix tests later? Then please create workitem.

src/NuGet.Services.Storage/AzureStorageQueue.cs
```diff
@@ -57,19 +60,25 @@ public async Task<StorageQueueMessage> GetNextAsync(CancellationToken token)
 
         public async Task RemoveAsync(StorageQueueMessage message, CancellationToken token)
         {
-            if (!(message is AzureStorageQueueMessage))
+            if (message is not AzureStorageQueueMessage queueMessage)
             {
                 throw new ArgumentException("This message was not returned from this queue!", nameof(message));
```
erdembayar, line 65: Could we make this hard coded string const?

src/NuGet.Services.Status.Table/ChildEntity.cs
ryuyu: Oops. Probably a bit of a faulty copy. Fixed.

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -117,31 +139,60 @@ public override async Task<bool> ExistsAsync(string fileName, CancellationToken
...
         {
-            var files = await _directory.ListBlobsAsync(getMetadata, cancellationToken);
+            var blobTraits = new BlobTraits();
+            if (getMetadata)
+            {
+                blobTraits |= BlobTraits.Metadata;
+            }
 
-            return files.Select(GetStorageListItem).AsEnumerable();
+            foreach (BlobHierarchyItem blob in _directory.GetBlobsByHierarchy(traits: blobTraits, prefix: _path))
```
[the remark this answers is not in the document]
ryuyu: There is a second ListAsync method i added below that uses GetBlobsByHeirarchyAsync.
I didn't change the sync method as it was already in use and I don't want to break existing consumption.
it is up the consumer to switch which method is being used.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: We had a quick discussion with @agr offline and we decided to switch to BlockBlobClient here as we don't currently have scenarios that need the other specialized BlobClient types. We will revisit if these scenarios become necessary.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: You are correct; that was indeed unused. I switched it to an appendResourceOptions type and used it when creating the destination resource.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: Oh that's a great question. I think I naively assumed this to be setting headers on the upload request, but it looks like it's for setting consumption headers on the blob.
I looked around a bit and it looks like we want to reverse the order of these operations. Good catch.

src/NuGet.Services.Storage/AzureStorage.cs
ryuyu: Discussed with @erdembayar and @agr offline and we are switching to BlockBlobClient as the performance gains for using BlockBlobClient will be more useful than the genericism of BlobClient (since we don't have any scenarios currently that would benefit from using any of the other types anyway) Thanks for bringing this up!

tests/NuGet.Services.Storage.Tests/AzureStorageFacts.cs
ryuyu: I will probably delete this file. Making a note to remove it.

src/NuGet.Services.Storage/AzureStorage.cs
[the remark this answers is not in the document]
drewgillies: It's worth pointing out that we'd probably like to circle back to this work and do a sync -> async pass where appropriate. But not at this time.

erdembayar (approved):

PR #432, opened 2024-07-11
Use latest C# version

erdembayar:
Partially address: https://github.com/NuGet/Engineering/issues/5525
We could move to latest C# version like 12 which [is same as client team](https://github.com/NuGet/NuGet.Client/blob/5f1401c2de4e7d75f8c58bbb45d9df828f2de2d1/build/common.project.props#L98).
image

It's [blocking NuGet.Jobs release](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9869300&view=logs&j=8515a029-22a6-5510-9e17-f5665af94c87&t=6006c852-547b-5c40-1e5c-b5a098dddd13&l=589)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,6 +20,10 @@
     <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <LangVersion>12.0</LangVersion>
+  </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers">
--- a/src/NuGet.Services.Build/NuGet.Services.Build.csproj
+++ b/src/NuGet.Services.Build/NuGet.Services.Build.csproj
@@ -7,7 +7,7 @@
     <!--
     We force C# 5 because the FindDuplicateFiles task is compiled using CodeTaskFactory, which only supports this C# version.
     -->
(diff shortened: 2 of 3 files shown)


src/NuGet.Services.Build/NuGet.Services.Build.csproj
```diff
@@ -3,11 +3,6 @@
   <PropertyGroup>
     <TargetFramework>net472</TargetFramework>
     <Description>Shared component to assist NuGet build scripts.</Description>
-
```
erdembayar, line 6: Seems like not problem anymore, both package release and CI build succeeds.

tests/NuGet.Services.Contracts.Tests/ContractsFacts.cs
```diff
@@ -21,6 +21,11 @@ public void ShouldOnlyHaveInterfacesAndEnums()
             Assert.NotEmpty(types);
             foreach (var type in types)
             {
+                if (type.FullName.StartsWith("Microsoft.") || type.FullName.StartsWith("System."))
```
erdembayar, line 24: Latest C# adds some analysis Microsoft and System packages by default .

Directory.Build.props
```diff
@@ -20,6 +20,10 @@
     <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <LangVersion>12.0</LangVersion>
```
erdembayar, line 25: IIRC using the `latest` version might break things as there are some syntaxes that couldn't be used in .NET Framework, so version `12` is a safe bet.

PR #435, opened 2024-07-23
Add how documentation for how to create local db

erdembayar:
Added steps for how to create a local database which is used for [NuGet.Jobs](https://github.com/NuGet/NuGet.Jobs/).


drewgillies (approved):
<3

PR #437, opened 2024-07-26
Discover project list for tests and assembly info

joelverhagen:
It's silly we list all of our projects directly for assembly info generation and test running. I added `Get-SolutionProjects` to enumerate them from the .sln.

Ideally we would not generate our own assembly info and use MSBuild... but this is a bigger change.
Ideally we would use `dotnet test` on the .sln directly... but this changes test output and forgets us to support `dotnet test` on all projects at once.

So I chose this simple approach for now.

Summary of changes:
- Discover project list for tests and assembly info from solution file
- Move to latest LangVersion
- Remove last of TeamCity specific code
- Move to newer nuget.exe in common.ps1
- Delete unused common.ps1 method `Install-DotnetCLI`
- Use ADO CI log groups -- it's pretty cool


Directory.Build.props
```diff
@@ -11,7 +11,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <IncludeSymbols Condition="'$(IncludeSymbols)' == ''">true</IncludeSymbols>
     <SymbolPackageFormat Condition="'$(SymbolPackageFormat)' == ''">snupkg</SymbolPackageFormat>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>latest</LangVersion>
```
joelverhagen, line 14: @jimmylewis - hype hype hype!!!

dannyjdev (approved):

agr (approved):

PR #439, opened 2024-07-29
Align package versions, remove unused method, quality "common" build steps

joelverhagen:
This is further clean up to ease a repository merge. 

Summary of changes:
- Switch publish test results to `VSTest` since `dotnet test` is used.
- Align package versions with NuGetGallery and NuGet.Jobs
- Simplify package source mapping to match NuGetGallery and NuGet.Jobs
- Remove build.ps1 `-CleanCache`. It's not used anywhere and is weak due to only clearing `temp` and `http-cache`
- Remove useless `app.config` files

### DIFF
--- a/.pipelines/ServerCommon-CI.yml
+++ b/.pipelines/ServerCommon-CI.yml
@@ -84,3 +84,4 @@ extends:
           inputs:
             testRunner: VSTest
             testResultsFiles: $(ServerCommonPath)\Results.*.xml
+            failTaskOnFailedTests: true
--- a/.pipelines/ServerCommon-CI.yml
+++ b/.pipelines/ServerCommon-CI.yml
@@ -77,10 +77,10 @@ extends:
             scriptName: $(ServerCommonPath)\test.ps1
             arguments: -Configuration $(BuildConfiguration) -BuildNumber $(Build.BuildId)
             workingFolder: $(ServerCommonPath)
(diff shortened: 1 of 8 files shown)


.pipelines/ServerCommon-CI.yml
```diff
@@ -77,10 +77,11 @@ extends:
...
             arguments: -Configuration $(BuildConfiguration) -BuildNumber $(Build.BuildId)
             workingFolder: $(ServerCommonPath)
-        - task: PublishTestResults@1
+        - task: PublishTestResults@2
           name: PublishTestResults_3
           displayName: Publish Test Results Results.*.xml
           condition: succeededOrFailed()
           inputs:
-            testRunner: XUnit
+            testRunner: VSTest
```
erdembayar, line 85: Is there somewhere recommendation for using VSTest over Xunit if `dotnet test` is used?

joelverhagen: In the task docs it uses `VSTest` when `dotnet test` was used:
https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-test-results-v2

The way I discovered this was noticing the task was not uploading test results after I moved to `dotnet test`.

erdembayar (approved):

jimmylewis (approved):

PR #440, opened 2024-07-29
Add global MSI client id for KV

dannyjdev:
# Changes
* Added "ManagedIdentityClientId" configuration, so that we can add only one MSI on config files for all services in a job, KV/Storage/etc.
* KeyVault configuration now uses "KeyVault_ClientId" if specified or fallbacks to "ManagedIdentityClientId".

### DIFF
--- a/src/NuGet.Services.Configuration/Constants.cs
+++ b/src/NuGet.Services.Configuration/Constants.cs
@@ -7,6 +7,7 @@ namespace NuGet.Services.Configuration
     {
         public static string KeyVaultVaultNameKey = "KeyVault_VaultName";
         public static string KeyVaultUseManagedIdentity = "KeyVault_UseManagedIdentity";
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.Configuration/Constants.cs
```diff
@@ -7,6 +7,7 @@ public static class Constants
     {
         public static string KeyVaultVaultNameKey = "KeyVault_VaultName";
         public static string KeyVaultUseManagedIdentity = "KeyVault_UseManagedIdentity";
+        public static string ManagedIdentityClientIdKey = "ManagedIdentityClientId";
```
joelverhagen, line 10: Should this be prefixed with `KeyVault_`?

joelverhagen: Spoke over IM. LGTM.

joelverhagen (approved):

PR #441, opened 2024-08-01
Change Azure Storage copy to use server side copy

joelverhagen:
The only place that uses this method is here:
https://github.com/NuGet/NuGetGallery/blob/99b2b4961bc4d62e9bab60993390b08260be639a/src/Stats.PostProcessReports/DetailedReportPostProcessor.cs#L208

This uses server-side copy already. So the extra dependency on `DataMovement` is not needed.

This PR essentially copies the copy implementation from the original copy of this copied project, still existing here:
https://github.com/NuGet/NuGetGallery/blob/99b2b4961bc4d62e9bab60993390b08260be639a/src/Catalog/Persistence/AzureStorage.cs#L221-L274

Also add a helper method to replicate this:
https://github.com/NuGet/NuGetGallery/blob/2e4a1498efac7feb99ce49a71b0bf13863448fad/src/NuGetGallery.Core/Services/CloudBlobClientWrapper.cs#L45-L46

### DIFF
--- a/src/NuGet.Services.Storage/AzureStorage.cs
+++ b/src/NuGet.Services.Storage/AzureStorage.cs
@@ -10,8 +10,6 @@ using System.Threading.Tasks;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Azure.Storage.Blobs.Specialized;
-using Azure.Storage.DataMovement;
-using Azure.Storage.DataMovement.Blobs;
 using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.Storage
@@ -20,25 +18,18 @@ namespace NuGet.Services.Storage
     {
         private readonly ILogger<AzureStorage> _logger;
         private readonly BlobContainerClient _directory;
-        private readonly TransferManager _transferManager;
-        private readonly BlobsStorageResourceProvider _storageResourceProvider;
-        private readonly bool _useServerSideCopy;
         private readonly string _path;
 
         public AzureStorage(
             BlobServiceClient account,
-            BlobsStorageResourceProvider blobsStorageResourceProvider,
             string containerName,
             string path,
             Uri baseAddress,
-            bool useServerSideCopy,
             bool initializeContainer,
             ILogger<AzureStorage> logger)
             : this(
                   account.GetBlobContainerClient(containerName),
-                  blobsStorageResourceProvider,
                   baseAddress,
-                  useServerSideCopy,
                   initializeContainer,
                   logger)
         {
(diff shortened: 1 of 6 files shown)


src/NuGet.Services.Storage/AzureStorageFactory.cs
```diff
@@ -3,39 +3,40 @@
...
         BlobServiceClient _account;
-        BlobsStorageResourceProvider _storageResourceProvider;
         string _containerName;
         string _path;
         private Uri _differentBaseAddress = null;
         private readonly ILogger<AzureStorage> _azureStorageLogger;
-        private readonly bool _useServerSideCopy;
         private readonly bool _initializeContainer;
 
+        public static string PrepareConnectionString(string connectionString)
```
erdembayar, line 19: How this is used? I don't see any call for it.

joelverhagen: It will be used in NuGetGallery. Example:
https://github.com/NuGet/NuGetGallery/commit/ef925f2fc65dc1361d13671707883b79cae12b85

zhhyu (approved):

src/NuGet.Services.Storage/AzureStorage.cs
```diff
@@ -207,35 +190,45 @@ private StorageListItem GetStorageListItem(BlockBlobClient listBlobItem)
...
-            }
-
-            StorageResource sourceStorageResource = _storageResourceProvider.FromBlob(sourceUri.ToString());
-            StorageResource destinationStorageResource = _storageResourceProvider.FromBlob(destinationUri.ToString(), destinationOptions);
 
-            var transferOptions = new DataTransferOptions();
-
-            DataTransfer transfer = await _transferManager.StartTransferAsync(sourceStorageResource, destinationStorageResource, transferOptions: transferOptions, cancellationToken: cancellationToken);
-            await transfer.WaitForCompletionAsync();
+                await destinationBlockBlob.SetHttpHeadersAsync(headers, cancellationToken: cancellationToken);
```
dannyjdev, line 230: I think we should do something similar like the following example: [Manage blob properties and metadata with .NET](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-properties-metadata#set-and-retrieve-properties), since just using an empty BlobHttpHeaders will clear all previous headers and only set the ones we explicitly provide.

joelverhagen: Awesome catch @dannyjdev! I will make an update to both ServerCommon and NuGetGallery `dev` branch with a fix!

joelverhagen: PR here: https://github.com/NuGet/ServerCommon/pull/443

PR #444, opened 2024-08-02
Force AzureStorageFactory to specify blob public access

joelverhagen:
Before moving to the new Azure SDK, `AzureStorage` was tolerant of existing containers having a different public access level:
https://github.com/NuGet/ServerCommon/blob/f6781f41bb6ec921b8fe81cc05882fe0af7897f6/src/NuGet.Services.Storage/AzureStorage.cs#L53-L55

Because of this, the actual public blob access level on some containers (such as the one used for the Gallery.CredentialExpiration job) have drifted from Blob public access (i.e. becoming no public access). Because of the previous implementation, this was allowsed. `CreateIfNotExists` would return false because the container already existed and updating the public access level would not happen. With the current implementation based on the new SDK, the code tries to update the public access level every time. This fails when the storage account has public access disabled, causing related jobs to fail during startup.
https://github.com/NuGet/ServerCommon/blob/2ccce56c06eebf0759863ca2d68beaf18ebb3916/src/NuGet.Services.Storage/AzureStorage.cs#L51

The fix is to allow the caller to specify what public access level they really need. In the case of the Gallery.CredentialExpiration job, this will be set to `enablePublicAccess: false`.

This is a breaking change in the constructor but I believe it is better to be explicit about whether a container should be public or not.

List of jobs that use this:
- Gallery.CredentialExpiration, `enablePublicAccess: false`
- GitHubVulnerabilities2V3, `enablePublicAccess: true`
- GitHubVulnerabilities2Db, `enablePublicAccess: true`
- Stats.PostProcessReports
  - source storage, `enablePublicAccess: false`
  - work storage, `enablePublicAccess: false`
  - destination storage, `enablePublicAccess: true`
- Validation.PackageSigning.ProcessSignature, `enablePublicAccess: false`
- Validation.PackageSigning.ValidateCertificate, `enablePublicAccess: false`

### DIFF
--- a/src/NuGet.Services.Storage/AzureStorageFactory.cs
+++ b/src/NuGet.Services.Storage/AzureStorageFactory.cs
@@ -11,6 +11,7 @@ namespace NuGet.Services.Storage
     {
         BlobServiceClient _account;
         string _containerName;
+        private readonly bool _enablePublicAccess;
         string _path;
         private Uri _differentBaseAddress = null;
         private readonly ILogger<AzureStorage> _azureStorageLogger;
@@ -27,14 +28,15 @@ namespace NuGet.Services.Storage
         public AzureStorageFactory(
             BlobServiceClient account,
             string containerName,
+            bool enablePublicAccess,
             ILogger<AzureStorage> azureStorageLogger,
             string path = null,
             Uri baseAddress = null,
-            bool initializeContainer = true
-            )
+            bool initializeContainer = true)
         {
             _account = account;
             _containerName = containerName;
+            _enablePublicAccess = enablePublicAccess;
             _path = null;
             _azureStorageLogger = azureStorageLogger;
             _initializeContainer = initializeContainer;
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.Storage/AzureStorageFactory.cs
```diff
@@ -27,14 +28,15 @@ public static string PrepareConnectionString(string connectionString)
         public AzureStorageFactory(
             BlobServiceClient account,
             string containerName,
+            bool enablePublicAccess,
```
ryuyu, line 31: This looks like it would be an API breaking change for those that are already consuming. Do you thnik it would be ok for us to put this closer to the end and give it a default value that falls back to the currently existing behaviour to minimize potential breaks for others?

ryuyu: Oops I see you made a comment about public/private access being explicit and you thought about this. I'm ok with your reasoning. Please ignore this comment.

ryuyu (approved):

src/NuGet.Services.Storage/AzureStorageFactory.cs
joelverhagen: This API is already "broken" (as in breaking change) due to Azure SDK change. I think this breaking change adds little additional work upon the current state.

joelverhagen: IMO it's a better shape now. Prior to my PR it was possible for private blob containers to become public all of a sudden since it was hard coded to public access = blob.

